PR #15, opened 2020-10-28
Bringing the repository in line with Github idioms

dominic-mulligan-arm:
- Removed the SECURITY_DISCLOSURE.markdown file.  This was replaced by another Markdown file added to the repository (containing the same text) through the Github web-interface.  There seems to be no obvious way to avoid this.
- Updated the README.markdown and CONTRIBUTING.markdown files to reflect the fact that the project-level code-of-conduct was added through the Github web-interface.  Again, there seems to be no obvious way to tell Github that these documents already exist in the repository.


geky:
I'm not sure what is going on here. Are these files duplicated outside of the git repo?

The Markdown looks valid otherwise, though it looks like this deletes `SECURITY_DISCLOSURE.markdown`, do we not need that file?

README.markdown
```diff
@@ -1438,8 +1438,11 @@ Veracruz codebase.
 
 See the `CONTRIBUTING.markdown` file for information on how to contribute to
 Veracruz, where to find help, how to get in touch with the rest of the
-Veracruz development team, how to contribute code, and details of our
-project-level code-of-conduct.
+Veracruz development team, and how to contribute code.
+
+The Veracruz project has adopted a project-level code of conduct, the full
+text of which is available in the CODE_OF_CONDUCT.markdown file in the
```
geky, line 1444: nonblocking: 

You can use a relative link here. Both GitHub and GitLab respect these:
```
[CODE_OF_CONDUCT.markdown](CODE_OF_CONDUCT.markdown)
```

dominic-mulligan-arm:
> I'm not sure what is going on here. Are these files duplicated outside of the git repo?
> 
> The Markdown looks valid otherwise, though it looks like this deletes `SECURITY_DISCLOSURE.markdown`, do we not need that file?

Sorry, I wasn't clear what's happening: there's a measure of a repository's compliance to Github's recommended community standards under the "Community" tab in "Insights".  Most of this stuff we had already done, like have a security disclosure and code of conduct policy.  The problem is convincing Github that we have done those things already: there doesn't seem to be any way of telling Github that an existing file in the repository contains a code-of-conduct, and similar for the security policy.  So, the security policy is now in a file called SECURITY.markdown which contains the same text as the old SECURITY_DISCLOSURE.markdown file, and which has already been added directly to the main branch of the repository through the Github web-interface.  Similarly, Github seems to want a distinct CODE_OF_CONDUCT file (again, added to the repository already through the web-interface).  README and CONTRIBUTING have been updated to reflect these changes.

geky:
Ah I see, thanks for expanding. Looks good to me 👍

geky (approved):

dominic-mulligan-arm:
+1 reached, merging

Issue #155, opened 2021-06-03
Improve debugging of WASM programs


gbryant-arm:
**Requested feature**
Get more debugging information from WASM programs

**Motivation**
As of now, on the wasi branch (https://github.com/dominic-mulligan-arm/veracruz/tree/wasi-host-abi), the execution engine doesn't provide any information as to why a program fails, besides a generic error. Additionally, programs can't print anything on the host.
This makes it harder to debug WASM programs. `wasmtime` helps find WASM-specific bugs, but doesn't allow to find bugs that are related to the underlying TEE platform or to violations of Veracruz policies.

**Suggestions**
Propagating writes from the virtual `stdout`/`stderr` (inside the execution engine) to `stdout`/`stderr` on the host would greatly help.
Dumping the backtrace would be helpful as well.
Admittedly such features could introduce security issues and should be opt-in.
Some of these features have been implemented and tested on that branch: https://github.com/gbryant-arm/veracruz/tree/wasi-host-abi-debugging

PR #153, opened 2021-05-26
Add image processing example

gbryant-arm:
Adding image processing to the examples.
~~It must be built with the nightly toolchain, as specified in the toolchain file. This is due to `proper`, a dependency of `image`, using an unstable feature.~~ It doesn't have to be built with the nightly toolchain anymore :man_shrugging: 

The data generator fetches images from the web, and has the following benefits:
* no need to hardcode images in the git index
* copyright compliant
* access to realistic pictures, as opposed to computer generated ones
* code can be reused to fetch any kind of test data

Execution:
* Go to `sdk/freestanding-execution-engine`
* Run the execution engine with `test.png` mapped to the vfs:
  `cargo run -- -d test.png -p image-processing.wasm -x interp`

PR #157, opened 2021-06-10
Improve debugging of WASM programs

gbryant-arm:
Improve debugging of WASM programs. Based on #155.

Done:
* ~~Redirect writes to stdout and stderr to the host's stdout and stderr~~
* Short-term solution: Change the freestanding execution engine to dump the content of stdout/stderr after computation
* Fixed bug in execution engine
* Longer-term solution: Map this feature to Veracruz capabilities, so that only allowed principals may use it

TODO:
* Even further:
  * Dump backtrace (cf. https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/trap.rs)
  * Interface Veracruz with debuggers?

### DIFF
--- a/execution-engine/src/fs.rs
+++ b/execution-engine/src/fs.rs
@@ -21,7 +21,7 @@ use std::{
     path::{Path, PathBuf},
     vec::Vec,
 };
-use veracruz_utils::policy::principal::{Principal, RightsTable, StandardChannel};
+use veracruz_utils::policy::principal::{Principal, RightsTable, StandardStream};
 use wasi_types::{
     Advice, ClockId, DirCookie, DirEnt, ErrNo, Event, Fd, FdFlags, FdStat, FileDelta, FileSize,
     FileStat, FileType, Inode, LookupFlags, OpenFlags, PreopenType, Prestat, RiFlags, Rights,
@@ -105,7 +105,7 @@ impl FileSystem {
     /// The root directory name. It will be pre-opened for any wasm program.
     pub const ROOT_DIRECTORY: &'static str = "/";
     /// The root directory inode. It will be pre-opened for any wasm program.
-    /// File descriptors 0 to 2 are reserved for the standard channels.
+    /// File descriptors 0 to 2 are reserved for the standard streams.
     pub const ROOT_DIRECTORY_INODE: Inode = Inode(3);
     /// The root directory file descriptor. It will be pre-opened for any wasm program.
     pub const ROOT_DIRECTORY_FD: Fd = Fd(3);
@@ -120,7 +120,7 @@ impl FileSystem {
     #[inline]
     pub fn new(
         rights_table: RightsTable,
-        std_channels_table: &Vec<StandardChannel>,
+        std_streams_table: &Vec<StandardStream>,
     ) -> Self {
         let mut rst = Self {
             fd_table: HashMap::new(),
@@ -129,7 +129,7 @@ impl FileSystem {
             rights_table,
             prestat_table: HashMap::new(),
         };
-        rst.install_prestat(&vec![""], std_channels_table);
+        rst.install_prestat(&vec![""], std_streams_table);
         rst
     }
 
@@ -138,20 +138,20 @@ impl FileSystem {
     ////////////////////////////////////////////////////////////////////////
 
     #[inline]
-    /// Install standard channels (`stdin`, `stdout`, `stderr`)
-    fn install_standard_channels(
+    /// Install standard streams (`stdin`, `stdout`, `stderr`)
+    fn install_standard_streams(
         &mut self,
-        std_channels_table: &Vec<StandardChannel>,
+        std_streams_table: &Vec<StandardStream>,
     ) {
-        for std_channel in std_channels_table {
-            // Map each standard channel to an fd and inode.
+        for std_stream in std_streams_table {
+            // Map each standard stream to an fd and inode.
             // Rights are assumed to be already configured by the execution engine in the rights table
             // at that point.
             // Base rights are ignored and replaced with the default rights
-            let (path, fd_number, inode_number) = match std_channel {
-                StandardChannel::Stdin(file_rights) => (file_rights.file_name(), 0, 0),
-                StandardChannel::Stdout(file_rights) => (file_rights.file_name(), 1, 1),
-                StandardChannel::Stderr(file_rights) => (file_rights.file_name(), 2, 2),
+            let (path, fd_number, inode_number) = match std_stream {
+                StandardStream::Stdin(file_rights) => (file_rights.file_name(), 0, 0),
+                StandardStream::Stdout(file_rights) => (file_rights.file_name(), 1, 1),
+                StandardStream::Stderr(file_rights) => (file_rights.file_name(), 2, 2),
             };
             self.install_file(&path, Inode(inode_number), "".as_bytes());
             self.install_fd(
@@ -168,10 +168,10 @@ impl FileSystem {
     fn install_prestat<T: AsRef<Path> + Sized>(
         &mut self,
         dir_paths: &[T],
-        std_channels_table: &Vec<StandardChannel>,
+        std_streams_table: &Vec<StandardStream>,
     ) {
-        // Pre open the standard channels.
-        self.install_standard_channels(std_channels_table);
+        // Pre open the standard streams.
+        self.install_standard_streams(std_streams_table);
 
         // Install ROOT_DIRECTORY_FD is the first FD prestat will open.
         self.install_dir(Path::new(Self::ROOT_DIRECTORY), Self::ROOT_DIRECTORY_INODE);
--- a/execution-engine/src/fs.rs
+++ b/execution-engine/src/fs.rs
@@ -644,17 +644,6 @@ impl FileSystem {
     /// A rust-style base implementation for `fd_write`. It directly calls `fd_pwrite` with the
     /// current `offset` of Fd `fd` and then calls `fd_seek`.
     pub(crate) fn fd_write(&mut self, fd: Fd, buf: &[u8]) -> FileSystemResult<Size> {
-        // Redirect writes to fd(0) and fd(1) to the host's stdout and stderr respectively
-        if fd.0 == 0 || fd.0 == 1 {
-            let s = std::str::from_utf8(&buf).expect("Found invalid UTF-8");
-            print!("{}", s);
-            return Ok(buf.len() as u32)
-        } else if fd.0 == 2 {
-            let s = std::str::from_utf8(&buf).expect("Found invalid UTF-8");
-            eprint!("{}", s);
-            return Ok(buf.len() as u32)
-        }
-
         self.check_right(&fd, Rights::FD_WRITE)?;
         let offset = self.fd_table.get(&fd).ok_or(ErrNo::BadF)?.offset;
 
--- a/execution-engine/src/fs.rs
+++ b/execution-engine/src/fs.rs
@@ -644,6 +644,17 @@ impl FileSystem {
     /// A rust-style base implementation for `fd_write`. It directly calls `fd_pwrite` with the
     /// current `offset` of Fd `fd` and then calls `fd_seek`.
     pub(crate) fn fd_write(&mut self, fd: Fd, buf: &[u8]) -> FileSystemResult<Size> {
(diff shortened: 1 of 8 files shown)


execution-engine/src/fs.rs
```diff
@@ -644,6 +644,17 @@ impl FileSystem {
     /// A rust-style base implementation for `fd_write`. It directly calls `fd_pwrite` with the
     /// current `offset` of Fd `fd` and then calls `fd_seek`.
     pub(crate) fn fd_write(&mut self, fd: Fd, buf: &[u8]) -> FileSystemResult<Size> {
+        // Redirect writes to fd(0) and fd(1) to the host's stdout and stderr respectively
+        if fd.0 == 0 || fd.0 == 1 {
+            let s = std::str::from_utf8(&buf).expect("Found invalid UTF-8");
```
dominic-mulligan-arm, line 649: You should avoid using `expect` and other runtime panic-inducing functions in the trusted runtime's codebase.  Rather, these should be propagated up as explicit errors.

execution-engine/src/fs.rs
```diff
@@ -644,6 +644,17 @@ impl FileSystem {
     /// A rust-style base implementation for `fd_write`. It directly calls `fd_pwrite` with the
     /// current `offset` of Fd `fd` and then calls `fd_seek`.
     pub(crate) fn fd_write(&mut self, fd: Fd, buf: &[u8]) -> FileSystemResult<Size> {
+        // Redirect writes to fd(0) and fd(1) to the host's stdout and stderr respectively
+        if fd.0 == 0 || fd.0 == 1 {
```
dominic-mulligan-arm, line 648: I think we should introduce explicit constants for the `stdout`, `stdin`, and `stderr` handles, rather than comparing against bare numeric values.  Can you do that?

execution-engine/src/fs.rs
```diff
@@ -644,6 +644,17 @@ impl FileSystem {
...
+            let s = std::str::from_utf8(&buf).expect("Found invalid UTF-8");
+            print!("{}", s);
```
dominic-mulligan-arm, line 650: Somewhere, there is a debug printing function that will only print to `stdout` on the host if the policy `debug=true` flag is set.  That should be used here, rather than a `print!()`.

gbryant-arm: Are you referring to the profile option (https://doc.rust-lang.org/cargo/reference/profiles.html), or a Veracruz policy?

ShaleXIONG: There is a debug flag in the Veracruz runtime and a related `debug` function but has been exposed to the execution engine. Also, note that some TEE might simply do not support to call `print!` or the `print!` does pass information cross the TEE boundary.

execution-engine/src/fs.rs
ShaleXIONG: You might want to move these changes to `fd_pwrite` function. Another solution is to read the content in `Fd(1)` and `Fd(2)`, namely `stdout` and `stderr`, and dump them to the host, meaning cross the TEE boundary. You can define a function like `read_stdout` and `read_stderr` to return the content of these two files and let other components, e.g. veracruz runtime, decide if they want to call these functions and get the debug result.

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 372d8e901dfe24c8a1454363dd23cd25c2506218
* Result: SUCCEEDED
* [Build Logs](https://t4r5bdzqe9.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=75212f87-27ac-49e3-8b97-f64a217b210c%2Fbuild.log) (available for 30 days)

dominic-mulligan-arm:
@gbryant-arm, I think we should first step back and understand what the PR trying to accomplish: why can't Veracruz programs be debugged by compiling them to `x86_64` rather than `wasm32` and running them natively, or stepping through them with a standard debugger?  With the recent WASI changes the behaviour of the two should be identical.

I'm a bit wary of having random `println!` calls appear in the middle of the trusted runtime's codebase and wonder whether this is really a good solution to whatever problem you have...

gbryant-arm:
I believe debugging programs natively or outside Veracruz is acceptable until two many levels of non-determinism make them behave too differently from their execution inside Veracruz:
* A program, despite being successfully compiled to WASM, can exhibit non-determinism at runtime
* At the execution engine level, VC policies can affect a program's behaviour
* The isolation technology introduces another level of non-determinism, where some features (e.g. I/O) might work differently

To give a concrete example, I faced several issues when working on #153. The original Rust program executed successfully, but its WASM equivalent failed when run in wasmtime (thread pool related error, no backtrace support), and despite fixing these bugs, failed in Veracruz due to policies (nothing fancy, just the ones by default). And I didn't even try to execute it in a TEE.

To address your concern of using `println!` in the runtime:
We can implement safeguards at compile time (e.g. [debug assertions](https://doc.rust-lang.org/cargo/reference/profiles.html#debug-assertions) to make sure it doesn't end up in the production release but only in the debug release) and at runtime (Veracruz debug flag, Rust's log level for example).

Slightly related topic: To what extend does VC guarantee determinism horizontally (behaviour on different isolation technologies, or in a different WASM runtime)?
That is, is it enough to define clear APIs/ABIs shared by e.g. SGX, TrustZone and Nitro, to make sure that the program behaves "similarly" across those backends?
Also, is the VC execution engine truly compatible with e.g. WASI SDK? They share the same API, but not necessarily the same implementation, right?

dominic-mulligan-arm:
OK, thanks for clarifying.  Though, it's important to note that your experience debugging WASM programs wasn't typical, as you were trying to develop against the new WASI ABI whilst `main` (and `freestanding-execution-engine` with it) was still using the old ABI.  As a result, the tools usually used to debug were not available, and you had to use external tools like `wasmtime` that are potentially configured/setup in a different way to how we use them.  Now, coming to a few points in turn:

> A program, despite being successfully compiled to WASM, can exhibit non-determinism at runtime

It would be good to know what non-determinism you were seeing in WASM execution (modulo "obvious" things like random number generation) as I think this should either be investigated or documented.  My starting assumption is that execution of pure single-threaded WASM, as used in Veracruz, is deterministic, as should execution of the functions on our VFS.  If this isn't true, we've either got a bug, or we've got something that needs documenting, as that can affect reproducibility of computations.

> The isolation technology introduces another level of non-determinism, where some features (e.g. I/O) might work differently

The only I/O available to the WASM binary is via our virtual filesystem.  That should be deterministic.

> We can implement safeguards at compile time (e.g. debug assertions to make sure it doesn't end up in the production release but only in the debug release) and at runtime (Veracruz debug flag, Rust's log level for example).

At the moment, as far as I know, there is no use of `println` or logging within the trusted codebase (and if there is, it should be removed).  So adding this is a fairly substantial change to the project.  Moreover, it feels pretty hacky, too: as noted by @ShaleXIONG not all of our supported TEEs easily support `println` on the untrusted host.

Rather, I think the better way of doing this is to *not* modify the trusted codebase at all, but to modify `freestanding-execution-engine` to dump the content of the `stdout` and `stderr` files once a computation is complete from the runtime's VFS.  Then, `println` and friends can still be used for execution tracing without any risk of us introducing a privacy risk...

geky:
Could stdout and stderr be made files available to the policy? Then a client with correct permissions could download the logs without  bypassing Veracruz's trust model.

Later, with the streaming work, stdout/stderr might even be watchable in real-time.

Being able to print is incredibly valuable when your code in a black box doesn't work.

dominic-mulligan-arm:
@geky yeah, I think that could work: the policy could associate a list of files to the principals with `ResultReceiver` role, with `stdout` and `stderr` potentially included.

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 0eb51e0d1f97d1857eef005edfba87825e28d055
* Result: FAILED
* [Build Logs](https://t4r5bdzqe9.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=abdea078-3414-40a3-a9c9-a774e1185642%2Fbuild.log) (available for 30 days)

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 10baa908f0a356b0c1c870784e0a6fb11eb14918
* Result: FAILED
* [Build Logs](https://t4r5bdzqe9.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=15bd0029-3605-42fa-9a86-b8837cd52dcc%2Fbuild.log) (available for 30 days)

execution-engine/src/fs.rs
```diff
@@ -117,32 +118,60 @@ impl FileSystem {
...
-        rst.install_prestat(&vec![""]);
+        rst.install_prestat(&vec![""], std_streams_table);
         rst
     }
 
     ////////////////////////////////////////////////////////////////////////
     // Internal auxiliary methods
     ////////////////////////////////////////////////////////////////////////
 
+    #[inline]
```
dominic-mulligan-arm, line 140: This is too big to be marked `inline`.

sdk/freestanding-execution-engine/src/main.rs
```diff
@@ -254,5 +272,29 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+        .map_err(|e| format!("Failed to lock vfs, error: {:?}", e))?
+        .read_file_by_filename(
+            &Principal::InternalSuperUser,
+            "stderr",
+        )?;
+    let stderr_dump = std::str::from_utf8(&buf)
+        .map_err(|e| format!("Failed to convert byte stream to UTF-8 string: {:?}", e))?;
+    eprint!("{}", stderr_dump);
+
+    info!("WASM program {} loaded into VFS.", prog_file_name);
```
dominic-mulligan-arm, line 298: Is this log message in the right place?  It seems like this should be earlier, rather than right before the end of the program execution.

sdk/freestanding-execution-engine/src/main.rs
```diff
@@ -254,5 +272,29 @@ fn main() -> Result<(), Box<dyn Error>> {
     let return_code = execute(&cmdline.execution_strategy, vfs.clone(), &prog_file_name)?;
     info!("return code: {:?}", return_code);
     info!("time: {} micro seconds", main_time.elapsed().as_micros());
+
+    // Dump contents of stdout
+    let buf = vfs.lock()
```
dominic-mulligan-arm, line 277: Can we have a command-line option to enable this dumping of `stdout` and `stderr`?

veracruz-utils/src/policy/principal.rs
```diff
@@ -257,3 +257,17 @@ impl Identity<String> {
...
+////////////////////////////////////////////////////////////////////////////////
+// Standard streams.
+////////////////////////////////////////////////////////////////////////////////
+
+/// Standard streams (`stdin`, `stdout`, `stderr`) are mapped to the VFS just like any other file, though
+/// associated with specific file descriptors and inodes.
+/// Their path (filename) and access rights are configured in the JSON policy file.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+pub enum StandardStream {
+    Stdin(FileRights),
```
dominic-mulligan-arm, line 270: Can you add `rustdoc` comments for the constructors of any public enum?

sdk/freestanding-execution-engine/src/main.rs
```diff
@@ -72,12 +81,18 @@ struct CommandLineOptions {
     binary: String,
     /// The execution strategy to use when performing the computation.
     execution_strategy: ExecutionStrategy,
+    dump_stdout: bool,
```
dominic-mulligan-arm, line 84: These also need `rustdoc` comments!

dominic-mulligan-arm (commented):
Other than those missing comments on the `dump_stdout` and `dump_stderr` fields, this looks ready for merging.

dominic-mulligan-arm:
@gbryant-arm, can you fix the commenting issue above, rebase, and then I'll approve and we can get these changes in?

gbryant-arm:
Comments addressed.

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
LGTM.  Will merge when build completes.

dominic-mulligan-arm:
+1 reached, merging.

PR #146, opened 2021-05-13
Implement Certificate Authority Attestation 

dreemkiller:
Change the attestation process to match the proposal in #122.

This removes the old attestation mechanism.

PR #75, opened 2021-02-04
Add command-line interfaces

geky:
This is WIP, but pushing up to get feedback early.

I've added 3 new crates: tabasco-cli, sinaloa-cli, and durango-cli. These are thin layers over their respective crates that provide a command line interface, making it a bit easier to experiment with Veracruz as a whole.

TODO:

- [x] Add GETTING_STARTED_CLI.md
- [x] Add tests
- [x] Pass CI
- [x] Cleanup TODOs (either via fixes or issues)
- [x] Rebase onto master
- [x] Update docs (quite a bit changed)
- [x] ~Make CLIs consistent (vc-server/vc-fee mostly)~ postponed
- [x] Update based on reviewer's feedback
- [x] Remove merges?
- [ ] More feedback?
 
Should resolve https://github.com/veracruz-project/veracruz/issues/63

PR #163, opened 2021-06-23
Moved the test database generation

dreemkiller:
Moved the test database generation from `build.rs` files to the top level `Makefile`

The `build.rs` files were not being triggered consistently, resulting in sporadic test failures that were hard to diagnose.

In general, we've decided that mixing build steps between `build.rs` files and `Makefiles` makes things difficult, and `build.rs` cannot handle all of the cases we need, so this is part of a larger effort to minimize the use of `build.rs`


geky:
A problem I've noticed is that there are several `populate-test-database.sh` scattered in the repo of various levels of up-to-date (`proxy-attestation-server/populate-test-database.sh` was very out of date when I checked as a part of https://github.com/veracruz-project/veracruz/pull/75).

Have we considered consolidating these into a single populate-test-database.sh in, say, the test-collateral directory? That takes an argument specifying the path to a database to populate? This would be useful for tests, examples, and maybe even users.

Orrrrrr, as a part of a larger PR, could a default database be generated in proxy-attestation-server?

dominic-mulligan-arm:
> A problem I've noticed is that there are several `populate-test-database.sh` scattered in the repo of various levels of up-to-date (`proxy-attestation-server/populate-test-database.sh` was very out of date when I checked as a part of #75).
> 
> Have we considered consolidating these into a single populate-test-database.sh in, say, the test-collateral directory? That takes an argument specifying the path to a database to populate? This would be useful for tests, examples, and maybe even users.
> 
> Orrrrrr, as a part of a larger PR, could a default database be generated in proxy-attestation-server?

Yeah, I've noticed this too: there seems to be three copies of `populate_test_database.sh` in various subdirectories within the repository, and they all contain different material.  I got confused by this when working on the end-to-end-demo last week, as some of those scripts just don't work.  I'd be in favour of consolidating all of these into a single script, and putting the generated database in `test-collateral` too.

ShaleXIONG (dismissed):

dreemkiller:
I had forgotten about the one in `proxy-attestation-server`. That one is definitely obsolete. The other two (in `veracruz-test` and `veracruz-server-test` should be identical, but are not. (PR #146 fixes this).

It's a good idea to put one copy in `test-collateral`. I will do that.

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
LGTM, though I see there's conflicting files.  If you fix them, I'll approve again straight away.

dominic-mulligan-arm (approved):

ShaleXIONG (approved):

dominic-mulligan-arm (dismissed):

Makefile
```diff
@@ -61,7 +64,21 @@ trustzone: sdk trustzone-env
 	$(MAKE) -C trustzone-root-enclave trustzone OPTEE_DIR=$(OPTEE_DIR) OPTEE_OS_DIR=$(OPTEE_OS_DIR)
 	cd veracruz-client && RUSTFLAGS=$(SGX_RUST_FLAG) cargo build --lib --features tz
 
-sgx-veracruz-server-test:
+# Using wildcard in the dependencies because if they are there, and newer, it
+# should be rebuilt, but if they aren't there, they don't need to be built 
+# (they are optional)
+veracruz-test/proxy-attestation-server.db: #$(wildcard sgx-root-enclave/css.bin) $(wildcard nitro-root-enclave/PCR0)
```
geky, line 70: Are the dependencies supposed to be commented out here?

dreemkiller: They are not. Will fix now

geky (approved):
Looks good to me assuming tests pass 👍

test-collateral/populate-test-database.sh
```diff
@@ -7,4 +7,5 @@ echo "INSERT INTO firmware_versions VALUES(2, 'psa', '0.3.0', 'deadbeefdeadbeefd
 pcr0=`cat ../nitro-root-enclave/PCR0`
 echo "INSERT INTO firmware_versions VALUES(3, 'nitro', '0.1.0', '${pcr0}');" >> tmp.sql
 sqlite3 proxy-attestation-server.db < tmp.sql
```
nspin, line 9: Reading this script, one might assume it is run in the `test-collateral/` directory. Running it from `veracruz-test/` and `veracruz-server-test/` is a straightforward step from how it was before, and certainly works, but an alternative approach would be for this script to take the output path as an argument, and be invoked by the Makefile like

```sh
cd veracruz-test && \
		bash ../test-collateral/populate-test-database.sh $@
```

dreemkiller: That would be an alternative approach. As it works as-is, I'm going to keep it this way for now, unless you see a problem with the way it is.

dreemkiller:
All tests pass

nspin (approved):

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
+1+1 = +1 reached, merging.

test-collateral/populate-test-database.sh
geky: I've changed populate-test-database.sh in https://github.com/veracruz-project/veracruz/pull/75 to accept a path and run-from-anywhere.

Also Diesel seems to have issues running outside of a rust crate, so added a cd into proxy-attestation-server. Not the most robust, but works in our Docker image.

PR #164, opened 2021-06-23
Relocating all vendored crates in 'third-party' and remove obsolete version pinning

mathias-arm:
Re-starting this PR, which adds all "vendored" crates to `third-party` as git submodules.

This version builds on #329, with additional un-pinning.


nspin:
What is the rationale behind this change?

mathias-arm:
It's an experiment / proposal.

I have had several discussions with @dreemkiller  and @ericvh, related to some crates that are referenced as git repositories, and in particular the large ones (rust with llvm, libc, ...), some get cloned multiple times during build. Eric suggested we might have them included in the Docker images, but this would make the images quite large and create some versioning challenges.
One way that I think we could improve on the current situation, is if we populated the `third_party` directory with the crates we use with reference to git repositories as "submodules". One question is if we should just do that with crates we "vendored" or all of the crates where we reference a git repository.
- The downside is that the first clone is going to be slower, but in many situation most of those clone operations would still happen later. There are also new opportunities to use CI-level caching to speed-up cloning.
- The upsides are IMHO:
  - Duplicated clones (cargo, xargo) should not happen. Also they would not re-occur if you started a new Docker image (you'd still need to download crates-io package and rebuild the sysroots).
  - It becomes easier to work on updates or modifications to those crates (currently you must chase all the places the crate is used and point towards your modified version).
  - Those dependencies are not versioned (it just picks the latest commit in the branch), so if you wanted to re-build a specific commit in `veracruz` it would not be easy. With submodules each change in those dependencies would require a commit in `veracruz`, it is a little bit of overhead but I think this is compensated by:
    - The increased reproducibility: if you go to an old version of the repository the submodules would be rolled back to the right version).
    - Easier maintenance: currently an API-breaking update to one of those "vendored" crates means you have to approve two or more PRs at the same time (and you can't run CI until you actually merged the crates changes).
  - While making the changes I noticed that in some instances we are not consistently using our version of the crates which might lead to some surprises in the future.

nspin:
> some get cloned multiple times during build

Why is this? Is this because we depend on multiple versions of them?

> Also they would not re-occur if you started a new Docker image.

Eric's recent PR splitting the Dockerfiles gathered various cache directories in `/cache`, which can be mounted from the host into the container. Does that contribution still miss some desired persistence?

> Those dependencies are not versioned (it just picks the latest commit in the branch).

In the past, I've suggested that we pin git `rev`s in our `Cargo.toml` files.

> Easier maintenance: currently an API-breaking update to one of those "vendored" crates means you have to approve two or more PRs at the same time (and you can't run CI until you actually merged the crates changes).

Based on my experience, I believe that, currently, a commit to the main Veracruz repository which depends on an unmerged submodule commit cannot be run in CI. See, for example, https://github.com/veracruz-project/veracruz/pull/149#issuecomment-863250260.

mathias-arm:
> > some get cloned multiple times during build
> 
> Why is this? Is this because we depend on multiple versions of them?

No. I think `xargo` is the main offender, it seems to be cloning separately for each sysroot.`rust-optee-trustzone-sdk` has the large `rust.git` as a submodule.

> > Also they would not re-occur if you started a new Docker image.
> 
> Eric's recent PR splitting the Dockerfiles gathered various cache directories in `/cache`, which can be mounted from the host into the container. Does that contribution still miss some desired persistence?

It does improve the fact that you can share across containers.

> > Those dependencies are not versioned (it just picks the latest commit in the branch).
> 
> In the past, I've suggested that we pin git `rev`s in our `Cargo.toml` files.

It would be more flexible as you can point at a commit in another branch. But it would also mean having to modify all the use of that crate in the (for many of them that means dozens of places).

> > Easier maintenance: currently an API-breaking update to one of those "vendored" crates means you have to approve two or more PRs at the same time (and you can't run CI until you actually merged the crates changes).
> 
> Based on my experience, I believe that, currently, a commit to the main Veracruz repository which depends on an unmerged submodule commit cannot be run in CI. See, for example, [#149 (comment)](https://github.com/veracruz-project/veracruz/pull/149#issuecomment-863250260).

Yes. But with submodules you can merge the submodule commit and not break CI and everyone's build until the main commit has been merged and pull by everyone.

dreemkiller:
> In the past, I've suggested that we pin git revs in our Cargo.toml files.

This would solve a lot (but not all?) of the problems this PR solves with a lot less submodule pain. Submodules are notoriously tricky to use correctly, and extremely easy to screw up (as I have done many times).

mathias-arm:
One example of the issue we can see right now with veracruz-project/veracruz#167 and veracruz-project/wasmtime#2 (can't CI test the first without merging the second, merging the second will break CI and builds).

Pinning git revs could solve that issue in theory, but in practice I think it would turn out to be a nightmare for veracruz as whole. Many of the vendored crates have been modified to use to the `veracruz` branch of other vendored crates (e.g. `ring`, `rustls`, `webpki` point to `rust-optee-trustzone-sdk`, `bytes` points to `serde`). Now if you pinned revs and wanted to make an update to one of those crates, you would need to update the rev in all the other repositories that rely on it, then update all the repositories that rely on those repositories, and so on. That would be painful if you needed to commit those changes, but we use pull requests and merge commits (which change the commit IDs). Minor point: if it turns out we have a cyclic dependency graph we won't be able to use this strategy in some places.

What pinned revs do not give you is reducing the reduce the number extra git clones and control of where/when those git operations happen.

I am not saying git submodules are the easiest thing in git, but apparently my experience has been less painful than yours.

nspin:
I've been using the `[patch]` manifest section for development, and that should work for CI as well while a PR is under review.

mathias-arm:
@nspin: in my tests it seemed that [patch] has a limited scope. It only applies to the direct dependencies, not the full dependency tree.

nspin:
Were the indirect dependencies that weren't being properly replaced being resolved to the exact version and source in antecedent spot of the `[patch]` manifest section?

mathias-arm:
As an example, if you look at `proxy-attestation-server` we patch `serde_json` and `log` to use our vendored versions, 1.0.51 and 0.4.8 respectively, but if you look at the output of `cargo tree` the latest official crate versions 1.0.64 and 0.4.14 get used a lot.

ShaleXIONG (dismissed):

mathias-arm:
@ShaleXIONG: I had to rebase after #345 got merged.

ShaleXIONG (approved):

ShaleXIONG:
@mathias-arm Done.

geky:
Looks good to me, I think submodules will be quite a bit easier to maintain going forward

geky (approved):

PR #171, opened 2021-07-09
Fix image generator

gbryant-arm:
The image generator generates images for the image processing example.
It downloads random images from the Internet, which may have been the cause of some CI issues (`Failed to load image: failed to fill whole buffer`).
This PR aims at replacing the downloader with a hardcoded image to simplify the codebase and make CI less error prone.


sdk/data-generators/image-processing-generator/src/main.rs
```diff
@@ -29,82 +28,82 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+
+    // Resize image. Aspect ratio is not preserved
+    let img = img.resize_exact(image_width, image_height, FilterType::Gaussian);
+
+    // Save image as PNG.
+    // This is required by the example, as `image` uses threads to load JPG images,
+    // which are not supported in WebAssembly
     let _result = img
-        .save_with_format(file_path, ImageFormat::Png)
+        .save_with_format(format!("{}", output_filename), ImageFormat::Png)
```
dominic-mulligan-arm, line 106: Why do you need `format!("{}", output_filename)` here?

sdk/data-generators/image-processing-generator/Makefile
```diff
@@ -10,6 +10,9 @@
 # and copyright information.
 
 
+# "Puerto de Veracruz" by Eduardo Pavon is licensed under CC BY-SA 2.0
```
dominic-mulligan-arm, line 13: I think the `LICENSE.markdown` file needs updating with the text of the license, and a notice that we are using that image.

sdk/data-generators/image-processing-generator/src/main.rs
```diff
@@ -29,82 +28,82 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-        .build()
-        .expect("Failed to configure the downloader");
-    let file_path = format!("{}.dat", file_prefix);
-    let dl =
-        Download::new(format!("https://picsum.photos/{}/{}", image_width, image_height).as_str())
-            .file_name(Path::new(&file_path));
-    let _result = downloader
-        .download(&[dl])
-        .map_err(|e| format!("Failed to download image: {}", e));
+    let output_filename = format!("{}.dat", file_prefix);
```
dominic-mulligan-arm, line 89: Why `.dat` instead of `.png`?

sdk/data-generators/image-processing-generator/src/main.rs
gbryant-arm: Right. I guess I initially meant to concatenate the filename with an extension

sdk/data-generators/image-processing-generator/Makefile
gbryant-arm: Should I append CC-BY-SA 2.0 to `LICENSE.markdown`?

sdk/data-generators/image-processing-generator/src/main.rs
gbryant-arm: That's the pattern followed by every other generator, that being said they generate numbers, not images

sdk/data-generators/image-processing-generator/Makefile
dominic-mulligan-arm: I think so, yes.  Probably add each license in a separate section, with the section name being the SPDX identifier of the license, or whatever.  Also, add a short preamble indicating which files are licensed under the `CC-BY-SA 2.0` license (and pointing out that the Veracruz codebase itself is licensed under the `MIT` license).

sdk/data-generators/image-processing-generator/src/main.rs
dominic-mulligan-arm: Yeah, I think in this case it makes more sense to use the established `png` file extension, as all other examples are creating *sui generis* data formats from `serde` serialisations of data structures.

sdk/data-generators/image-processing-generator/Makefile
gbryant-arm: Done. I followed this example: https://github.com/github/choosealicense.com#license
The README specifies what is covered by each license. `LICENSE.markdown` is left untouched.

ShaleXIONG (approved):
Is the extra license file for the image?

dominic-mulligan-arm:
@ShaleXIONG, yes, that's right.  Though note that if we are having two license files all the headers in the source tree need updating to point to the MIT license file, now.

dominic-mulligan-arm (dismissed):

dominic-mulligan-arm:
This looks good to merge when CI reports back.  I'll look into that now.

dominic-mulligan-arm (approved):

Issue #41, opened 2020-11-10
Add support for MCU clients


dreemkiller:
**Requested feature**
Add support for MCU clients with minimal resources running on a Cortex-M class device

**Motivation**
Currently, the client library is written without constrained devices in mind, so it is not optimized for size and performance.
We see use-cases where low end Cortex-M devices running an embedded RTOS would like to offload computation without having to trust the owner of the device being offloaded to.

**Additional context**
We currently use RustLS, which is not optimized for low end devices. Should probably use MbedTLS as the TLS library.

dominic-mulligan-arm:
Is this a duplicate of #9?  Actually, this has more information in it, so if you agree that there's a duplication then I will close #9.

Issue #176, opened 2021-07-20
We should test CLIs, MCUs, other clients in CI


geky:
**Requested feature**
We should test CLIs, MCUs, other clients in CI

**Motivation**
At the time of writing, it's not very easy to test anything more complicated than a standalone rust executable, this is because we rely on the TrustZone tests, which in turn rely on QEMU to execute code in an Arm environment where we know TrustZone exists. However this is a bit limited as we don't build in the same QEMU environment, requiring a copy from build environment to runtime environment.

The closest solution would be to use the Linux-process based target (https://github.com/veracruz-project/veracruz/pull/116), though there are other options such as IceCap (https://github.com/veracruz-project/veracruz/pull/149) or supporting SGX in CI.

Once we have support for native execution of an isolate target, this issue is to track the need to add the CLIs, MCUs and any other clients to CI afterwards.

PR #179, opened 2021-07-22
Add MCU client

geky:
Note this is dependent on https://github.com/veracruz-project/veracruz/pull/75

Adds a MCU client and demo based on [Zephyr OS](https://github.com/zephyrproject-rtos/zephyr). In order to test on Zephry OS, this uses a Docker image with the relevant build system, which makes things a bit complicated for both the build system and for networking.

To reproduce the demo you can follow the MCU_CLIENT_INSTRUCTIONS.markdown in the new veracruz-mcu-client directory.

Note that testing-wise, this is in the same state as https://github.com/veracruz-project/veracruz/pull/75. We can't easily add this into the CI until a CI friendly platform (Linux or IceCap) is merged. See https://github.com/veracruz-project/veracruz/issues/176 for more info.

Should resolve https://github.com/veracruz-project/veracruz/issues/41

TODO:

- [x] Move demo code into its own folder (~veracruz-mcu-client/demo~ veracruz-mcu-client/example unless anyone else has a better idea)
- [x] Add recv_results/send_program
- [x] Add minimal tests covering send_data/recv_results/send_program
- [x] Check Zephyr for cryptographically secure rng
- [x] Add RAM/ROM measurements
- [x] License pass
- [x] More documentation (per popular demand)
- [x] Remove run_markdown.sh (per popular demand)


veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 19521ac868097a242a9cddfda2aee19db7858039
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=d1596d2d-25a4-4893-908e-585fcf23fa47%2Fbuild.log) (available for 30 days)

veracruz-mcu-client/transport_protocol.proto
```diff
@@ -0,0 +1,259 @@
+//! Protocol buffers for Veracruz transport protocol messages
```
dreemkiller, line 1: This looks to be very similar to the proto file in the transport-protocol crate. Why is it ?mostly? duplicated here?

veracruz-mcu-client/vc.c
```diff
@@ -0,0 +1,577 @@
...
+        const char *str) {
+    if (level <= 1) {
+        printf("%s", str);
+    }
+}
+
+static int vc_rawrng(void *p,
+        uint8_t *buf, size_t len) {
+    // TODO use cryptographically secure rng?
+    sys_rand_get(buf, len);
```
dreemkiller, line 47: Where is `sys_rand_get` defined? Is it a cryptographically secure RNG?

veracruz-mcu-client/vc.c
```diff
@@ -0,0 +1,577 @@
...
+    return 0;
+}
+
+static ssize_t vc_rawsend(void *p,
+        const uint8_t *buf, size_t len) {
+    vc_t *vc = p;
+
+    if (vc->recv_len != 0) {
+        printf("data left!?!?!?!? %d\n", vc->recv_len);
```
dreemkiller, line 56: Should we handle this error some other way?

veracruz-mcu-client/vc.c
```diff
@@ -0,0 +1,577 @@
...
+
+    if (vc->recv_len == 0) {
+        // no data available? since we communicate over POSTs,
+        // we'll never have data available
+        printf("recv timeout\n");
+        return MBEDTLS_ERR_SSL_TIMEOUT;
+    }
+
+    size_t diff = (len < vc->recv_len) ? len : vc->recv_len;
+    memcpy(buf, vc->recv_pos, diff);
```
dreemkiller, line 176: What happens when `vc->recv_len > len`? Do we just return the truncated data?

veracruz-mcu-client/vc.c
```diff
@@ -0,0 +1,577 @@
...
+        const pb_field_iter_t *field,
+        void *const *arg) {
+    if (!pb_encode_tag_for_field(stream, field))
+        return false;
+
+    struct bytes *b = *arg;
+    return pb_encode_string(stream, b->buf, b->len);
+}
+
+int vc_send_data(vc_t *vc,
```
dreemkiller, line 477: Could we get some documentation on these functions? Does this function just handle sending Veracruz `data`, or is it also used to send Veracruz `program`? How would a client retrieve a computation result?

sdk/data-generators/shamir-secret-sharing/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
```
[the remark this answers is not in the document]
geky: Ah, that is because this is based and dependent on #75, it will go away after a rebase

test-collateral/ca-cert.conf
```diff
@@ -0,0 +1,18 @@
+[req]
+default_bits = 2048
```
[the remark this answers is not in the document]
geky: Ah ok, can update, this was just copied from the client certs. Is this because ECDSA is more secure or is there a security feature we depend on?

veracruz-mcu-client/Makefile
```diff
@@ -0,0 +1,117 @@
...
+		--name $(DOCKER_CONTAINER) \
+		$(DOCKER_IMAGE) bash -c "./tap-setup.sh ; bash")
+
+# Zephyr west commands
+.PHONY: update
+update:
+	west update
+
+# Generate clap data for demo
+clap.h clap.c: claps.wav claps_to_header.py
```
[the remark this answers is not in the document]
geky: I can move it to `veracruz-mcu-client/demo`, unless you have a better location in mind

veracruz-mcu-client/Makefile
```diff
@@ -0,0 +1,117 @@
...
+		--delay $(DELAY) \
+		--header=clap.h \
+		--source=clap.c)
+
+# Generate transport_protocol.pb.h/c
+transport_protocol.pb.h transport_protocol.pb.c: \
+		transport_protocol.proto transport_protocol.options
+	./nanopb/generator/nanopb_generator.py $< -f $(word 2,$^)
+
+#.DEFAULT_GOAL :=
```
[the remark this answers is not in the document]
geky: Not permanently, but I still want to add rules to get code size/ram size, which may cannibalize these comment, (actually I should create a TODO list)

veracruz-mcu-client/Makefile
```diff
@@ -0,0 +1,117 @@
...
+#.PHONY: build
+#build: policy.h policy.c
+#build: transport_protocol.pb.h transport_protocol.pb.c
+#build:
+#	west build -p auto -b qemu_cortex_m3
+
+.DEFAULT_GOAL :=
+.PHONY: build
```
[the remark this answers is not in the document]
geky: Hmm, not easily, it calls out to west (Zephyr's build tool), which does its own dependency management. If we wanted to track the dependencies correctly we'd need to track all of the source files in make, which would be fragile

veracruz-mcu-client/Makefile
```diff
@@ -0,0 +1,117 @@
...
+build:
+	west build -b native_posix
+
+.PHONY: clean
+clean:
+	rm -rf clap.h clap.c
+	rm -rf transport_protocol.pb.h transport_protocol.pb.c
+	rm -rf build
+
+.PHONY: rom_report
```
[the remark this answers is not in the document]
geky: Same answer as above, fortunately west is doing it's own dependency management, so we shouldn't see anything get rebuilt unless it's necessary.

There is an overhead for the number of different build systems we end up going through (though this path hasn't involved cargo... yet... 🙂), but that's not something I can control easily

veracruz-mcu-client/Makefile
```diff
@@ -0,0 +1,117 @@
...
+#
+#.PHONY: debug
+#debug: build
+#	$(QEMU) -gdb tcp::1234 -S $(QEMU_FLAGS) & $(GDB) $(ELF_PATH) -ex 'target remote localhost:1234'
+#	
+#.PHONY: debug-after
+#debug-after: build
+#	$(QEMU) -gdb tcp::1234 $(QEMU_FLAGS) & sleep 2 ; $(GDB) $(ELF_PATH) -ex 'target remote localhost:1234'
+
+run:
```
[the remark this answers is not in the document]
geky: Good point, will update

veracruz-mcu-client/base64.c
```diff
@@ -0,0 +1,156 @@
...
+ * The Veracruz Development Team.
+ *
+ * ## Licensing and copyright notice
+ *
+ * See the `LICENSE.markdown` file in the Veracruz root directory for
+ * information on licensing and copyright.
+ *
+ */
+
+#include "base64.h"
```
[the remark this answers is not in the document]
geky: I just followed the Wikipedia description of the algorithm, I looked around for a base64 implementation but I couldn't find one that was very standalone. Most of the C base64 libraries on Github advertise their usage of SIMD for fast encode/decode speed (there seems to be an ongoing competition), which isn't a good fit here. I'm sure I may have missed one, but base64 isn't all that hard to reimplement.

veracruz-mcu-client/claps_to_header.py
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
geky: Can move to veracruz-mcu-client/demo (see above).

The claps.wav binary is a valid concern, it's come up before (#171, we now have an image in the repo), and will probably continue to come up as we expand the number of demos we have.

Should we create a `veracruz-data` repo to capture these? It could be included as a submodule to still allow offline compilation without bloating the size of Veracruz's core git history.

veracruz-mcu-client/main.c
```diff
@@ -0,0 +1,153 @@
...
+
+#include <kernel.h>
+
+#include "xxd.h"
+#include "vc.h"
+#include "clap.h"
+#include "policy.h"
+
+// display audio samples in terminal
+void dump_samples(const int16_t *samples, size_t len,
```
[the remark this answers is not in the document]
geky: main.c/claps.c/claps.h are the demo, the rest is the MCU client, I can split these into veracruz-mcu-client/demo (see above)

veracruz-mcu-client/vc.c
geky: Will look into this, `sys_rand_get` is NOT cryptographically secure, but I'm not sure if Zephyr provides a cryptographically secure rng vs leaving it as an exercise for the user

veracruz-mcu-client/policy_to_header.py
```diff
@@ -0,0 +1,225 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
geky: So I've looked around quite a bit for previous projects, at many times it would have been great to have a C-friendly template engine to lean on, but every time I try one I quickly run into issues and the templating engine becomes more problems than it is worth.

There may be more issues, it's been a while since I've actually used a template engine, but some problems off the top of my head:
- Most templating engines aren't intended for C code (except maybe m4), the majority are intended for HTML (jinja, handlebars, etc), the resulting syntax can get difficult to read
- Template engines don't have as complete set of libraries as python does (though some like jinja get around this by just including python)
- State management can be a real problem. Template engines are great at unrolling lists and simple text processing, but when you need to take some data and build a structure based on that data, or convert to/from base64, or output data based on complex conditions at multiple points in multiple files, it can get messy real quick. Often the solution is to just prepare the data in python first beforehand.

The main thing python gives is flexibility, you don't have the risk of hitting some limitation halfway through the project and needing to rewrite your template to something else, since you can always use python to richly manipulate the code you are wanting to output.

But it could also just be that I was burned by template engines one too many times and I should give them another try. I've just gotten use to defaulting to python for code generation, which has the flexibility benefit and usually means no new languages need to be added to projects

./policy_to_header.py is also probably not the best example, since it's a bit messy and quickly thrown together. A lot of the generating defines are common and could be grouped together (.der -> byte arrays mostly)

veracruz-mcu-client/transport_protocol.proto
geky: Ah yes! This is a problem with the dockers that I'd be happy to have suggestions, I'm not sure what the correct solution is.

You are correct that veracruz-mcu-client/transport_protocol.proto is identical to transport_protocol/protos/transport_protocol.proto

I initially wanted to symlink these together, but veracruz-mcu-client currently builds in its own docker. So at build time the symlink points no where, because the destination doesn't exist in veracruz-mcu-client's docker.

I could add a step in the docker that copies transport_protocol.proto when you start the docker, but this option would mean that 1. transport_protocol.proto would not be updated unless you restarted your docker, 2. restarting docker would always trigger a rebuild due to the new timestamp (maybe there's a mv option to override this), 3. the source of transport_protocol.proto would be a bit hidden, maybe causing maintenance issues in the future.

Another option would be to change the veracruz-mcu-client docker to start from the Veracruz root directory, actually now that I'm typing this I don't see a reason not to, so that may be what to do

veracruz-mcu-client/vc.c
geky: Ah yes, this is lingering from a debug session. Should probably be an assert.

veracruz-mcu-client/vc.c
geky: Currently yes, can change this to an error

veracruz-mcu-client/vc.c
geky: These are documented in vc.h, though I can add more. The get_results/send_program functions are currently missing, but easy to add

veracruz-mcu-client/base64.c
```diff
@@ -0,0 +1,156 @@
...
+            out[e_len-j-4+3] = BASE64_ENCODE[v & 0x3f];
+        } else {
+            out[e_len-j-4+3] = '=';
+        }
+    }
+
+    return e_len;
+}
+
+//static size_t strnlen(const char *s, size_t s_len) {
```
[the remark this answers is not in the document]
geky: This one is a bit interesting. strnlen is one of those pesky functions that isn't in libc, but is practically available on most systems. "Most systems" of course excluding MCU stdlibs. This created an annoyance where depending on which target you are compiling Zephyr for, strlen may or may not be available. If it isn't available, this implementation is needed, if it is this creates a conflict.

I should just rename this to base64_strnlen.

proxy-attestation-server/src/server.rs
```diff
@@ -195,14 +195,22 @@ async fn nitro_router(nitro_request: web::Path<String>, input_data: String) -> P
     Err(ProxyAttestationServerError::UnimplementedRequestError)
 }
 
-pub fn server(url: String, ca_cert_path: &str, debug: bool) -> Result<Server, String> {
+pub fn server<P1, P2>(url: String, ca_cert_path: P1, ca_key_path: P2, debug: bool) -> Result<Server, String>
+where
+    P1: AsRef<path::Path>,
+    P2: AsRef<path::Path>
```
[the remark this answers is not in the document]
geky: This is necessary, at type inference time P1 and P2 are not the same. One could `&str`, another could be `std::PathBuf`. If you use the same type variable for both paths, the compiler would actually enforce that both ca_cert_path and ca_key_path have identical types, which could get very confusing.

It's not pretty but it is idiomatic Rust. One thing I've learned is that readability is really not that high of a priority in the language... It also really hates 80-column limits...

veracruz-client/src/veracruz_client.rs
```diff
@@ -163,11 +163,33 @@ impl VeracruzClient {
...
+    /// related hash.
+    /// Attest the enclave.
+    pub fn with_policy_and_hash<P1, P2>(
+        client_cert_filename: P1,
+        client_key_filename: P2,
+        policy: Policy,
+        policy_hash: String,
+    ) -> Result<VeracruzClient, VeracruzClientError>
+    where
+        P1: AsRef<path::Path>,
```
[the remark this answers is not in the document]
geky: See above

run_markdown.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
geky: @dominic-mulligan-arm we had an offline discussion that ended in mutually agreed disagreement so would value a third opinion here.

IMO this is the best way to test example documents. The alternatives:
- Change MCU_CLIENT_INSTRUCTIONS.markdown to MCU_CLIENT_INSTRUCTIONS.sh with documentation in the comments may be more intimidating to new users and less readable
- Not testing MCU_CLIENT_INSTRUCTIONS.markdown may lead to an out of date example that doesn't compile

Using a script to convert the .markdown to a .sh does mean another script to maintain, but it is really just a regex that doesn't do anything very fancy.

Makefile
```diff
@@ -57,18 +58,88 @@ sgx: sdk sgx-env
...
+		OPENSSL_LIB_DIR=$(OPENSSL_LIB_DIR) \
+		C_INCLUDE_PATH=$(TRUSTZONE_C_INCLUDE_PATH) \
+		cargo build --target aarch64-unknown-linux-gnu --features tz --features cli
+	cd veracruz-client && RUSTFLAGS=$(SGX_RUST_FLAG) cargo build --features tz --features cli
+	# build CLIs in the SDK/test-collateral
+	$(MAKE) -C sdk/freestanding-execution-engine
+	$(MAKE) -C sdk/wasm-checker
+	$(MAKE) -C test-collateral/generate-policy
+
+%-cli-install: %-cli
```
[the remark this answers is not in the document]
geky: That's an interesting question, I haven't seen a `make install` not be phony before.

One issue that comes up is you may want to run `make sgx-cli-install` because you or something else messed with the existing installed binaries. Such a case would think there's no work to do because the dest is more recent, which is inaccurate. This could easily happen if you're switching between sgx/trustzone/nitro in the same machine.

It's worth noting `%-cli-install` doesn't actually do any compilation itself. It depends on `%-cli` which should not need to be rebuilt if the binary files are up to date.

dominic-mulligan-arm:
@geky, now #75 has been merged, can you rebase this to make it easier to review?

geky:
Should be up to date now 👍

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 9570660c21c8a7876c1d6db40823a7a24f7ee0c3
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=5233b834-1dce-4177-9290-4521284e527c%2Fbuild.log) (available for 30 days)

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: fb3183023383682e90e5b5d44d5c7bd40d5bf218
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=b290a017-8f77-4594-84d6-5f3ddb6ae7c6%2Fbuild.log) (available for 30 days)

sdk/rust-examples/audio-event-triangulation/src/main.rs
```diff
@@ -0,0 +1,138 @@
...
+
+struct AudioEvent {
+    #[allow(dead_code)]
+    timestamp: u32,
+    location: (i32, i32),
+    samples: Vec<i16>,
+}
+
+impl AudioEvent {
+    fn power(&self) -> f64 {
```
dominic-mulligan-arm, line 31: Can we have some `rustdoc` comments on the functions in this file, explaining what they are doing?

sdk/rust-examples/audio-event-triangulation/src/main.rs
```diff
@@ -0,0 +1,138 @@
...
+        for x in self.samples.iter() {
+            p += x.abs() as f64;
+        }
+
+        p / (self.samples.len() as f64)
+    }
+}
+
+fn decode_audio_event(event: &[u8]) -> AudioEvent {
+    AudioEvent{
```
dominic-mulligan-arm, line 43: If any of the input files aren't stored in a correct format then this will abort at runtime due to the `unwrap()`, right?  Can this be made more robust, e.g. introducing an explicit error type that's serialised to the output file indicating either success or failure (with a reason), rather than just aborting?

sdk/rust-examples/audio-event-triangulation/scripts/poll_for_result.py
```diff
@@ -0,0 +1,52 @@
...
+        print()
+        print('\033[1;33maudio event detected!\033[0m')
+        print('\033[1;33mlocation:\033[0m %s' % dump_gps(location))
+        print('\033[1;33mtimestamp:\033[0m %u' % int(time.time()))
+        print()
+
+        break
+
+if __name__ == "__main__":
+    import sys
```
dominic-mulligan-arm, line 51: Float to the top?

veracruz-mcu-client/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+
```
dominic-mulligan-arm, line 1: Missing license header/authorship information.

veracruz-mcu-client/Dockerfile
```diff
@@ -0,0 +1,77 @@
...
+# Install Zephyr
+WORKDIR /zephyr-workspace
+RUN pip3 install west && \
+    west init --mr v2.4.0 && \
+    west update && \
+    pip3 install -r zephyr/scripts/requirements.txt
+
+# Get the Zephyr SDK
+#
+# TODO we don't need all the compilers in this package... can we reduce
```
dominic-mulligan-arm, line 59: Is this TODO still relevant?  If so, can you open up an issue in the issue tracker to make sure it isn't forgotten?

veracruz-mcu-client/Makefile
```diff
@@ -0,0 +1,117 @@
+# Project name/paths, used for docker image
+TARGET ?= mini-durango
```
dominic-mulligan-arm, line 2: Stray mention of Durango, here.

veracruz-mcu-client/claps_to_header.py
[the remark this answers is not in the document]
dominic-mulligan-arm: What is the license on the `claps.wav` file?  Do we need another `LICENSE` file or is it licensed under one of our existing licenses (e.g. the CC_BY_SA 2.0 license used for the photo)?

veracruz-mcu-client/claps_to_header.py
```diff
@@ -0,0 +1,202 @@
...
+        for i in range(44, len(wav), 4)]
+
+    if args.debug:
+        print('raw:')
+        dump_samples(samples)
+
+    # downsample?
+    if args.bitrate:
+        slice = bitrate // args.bitrate
+#        samples = [samples[i]
```
dominic-mulligan-arm, line 87: Stray comments?

veracruz-mcu-client/claps_to_header.py
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python3
+#
+# ## Authors
+#
+# The Veracruz Development Team.
+#
+# ## Licensing and copyright notice
+#
```
dominic-mulligan-arm, line 8: Can you add an explanation of what is going on in this file?

veracruz-mcu-client/claps_to_header.py
```diff
@@ -0,0 +1,202 @@
...
+# See the `LICENSE.markdown` file in the Veracruz root directory for
+# information on licensing and copyright.
+#
+
+import struct
+import collections as co
+import sys
+
+TIMESTAMP = 1619580492
```
dominic-mulligan-arm, line 17: Perhaps a comment to make it clear that this is demo data is warranted, here.

veracruz-mcu-client/claps_to_header.py
```diff
@@ -0,0 +1,202 @@
...
+    (30.245633363333095, -97.75153437304053),
+    (30.222367321649394, -97.84045833888852),
+    (30.371553081578483, -97.756234602064),
+    (30.293773789031352, -97.74172587073012),
+    (30.236512828524855, -97.76291564214209),
+    (30.26576811133889,  -97.7496372275359),
+    (30.457328895861682, -97.82552717096148),
+]
+
+def dump_samples(samples, width=76, height=16):
```
dominic-mulligan-arm, line 28: I think this needs some explanation with a doc-comment.

veracruz-mcu-client/claps_to_header.py
```diff
@@ -0,0 +1,202 @@
...
+            f.writeln('#include <stdint.h>')
+            f.writeln()
+            f.writeln('const int16_t _CLAP_SAMPLES[%(len)d] = {',
+                len=len(claps[-1]))
+            for i in range(0, len(claps[-1])):
+                f.writeln('    %(sample)d,', sample=claps[-1][i])
+            f.writeln('};')
+
+if __name__ == "__main__":
+    import sys
```
dominic-mulligan-arm, line 177: Float to the top?

veracruz-mcu-client/http.c
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// HTTP GET operation
+ssize_t http_get(
+        const char *host,
+        uint16_t port,
+        const char *path,
+        uint8_t *buf,
+        size_t buf_len) {
+    // setup address, TODO use DNS?
```
dominic-mulligan-arm, line 54: Is this TODO still relevant?  If so, add an issue to the tracker?

veracruz-mcu-client/http.c
```diff
@@ -0,0 +1,201 @@
...
+    // allocate packet buffer for managing http connection
+    void *pbuf = malloc(256);
+    size_t pbuf_len = 256;
+    if (!pbuf) {
+        printf("http malloc failed (-ENOMEM)\n");
+        return -ENOMEM;
+    }
+
+    // create socket and connect to server
+    // TODO IPv6? can use net_sin(addr)->sin_family here
```
dominic-mulligan-arm, line 70: Similar comment as above about TODOs.

veracruz-mcu-client/http.c
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    // done, close
+    free(pbuf);
+    close(sock);
+    return state.pos;
+}
+
+// HTTP POST operation
+// TODO deduplicate?
```
dominic-mulligan-arm, line 120: Similar comment as above regarding TODO.

veracruz-mcu-client/http.c
```diff
@@ -0,0 +1,201 @@
...
+ssize_t http_post(
+        const char *host,
+        uint16_t port,
+        const char *path,
+        // TODO headers?
```
dominic-mulligan-arm, line 125: Similar comment as above regarding TODO.

veracruz-mcu-client/http.c
```diff
@@ -0,0 +1,201 @@
...
+        const uint8_t *payload_buf,
+        size_t payload_buf_len,
+        uint8_t *resp_buf,
+        size_t resp_buf_len) {
+    // setup address, TODO use DNS?
```
dominic-mulligan-arm, line 130: Similar comment as above regarding TODO.

veracruz-mcu-client/http_test.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
+#
+# ## Authors
+#
+# The Veracruz Development Team.
+#
+# ## Licensing and copyright notice
+#
+# See the `LICENSE.markdown` file in the Veracruz root directory for
```
dominic-mulligan-arm, line 9: Can you explain what this file is for in this header?

veracruz-mcu-client/http_test.py
```diff
@@ -0,0 +1,30 @@
...
+
+data = [
+    0x0a, 0x20, 0xbf, 0x5d, 0xd2, 0xb2, 0xc6, 0x28, 0x6d, 0xee, 0xe2, 0xf2, 0x07, 0x2a, 0xfb, 0xbc,
+    0xa2, 0x65, 0x15, 0x87, 0x3d, 0xa1, 0x2e, 0x51, 0xd8, 0xdc, 0x48, 0x1b, 0x73, 0x18, 0x61, 0xe5,
+    0x0d, 0x54                                          
+]
+
+http.client.HTTPConnection.debuglevel = 1
+
+c = http.client.HTTPConnection('172.17.0.2', 3017, timeout=3)
```
dominic-mulligan-arm, line 24: What is this IP address?

veracruz-mcu-client/http_test.py
```diff
@@ -0,0 +1,30 @@
...
+c.request("POST", "/sinaloa", body=base64.b64encode(b''.join(c.to_bytes(1, 'little') for c in data)))
```
dominic-mulligan-arm, line 25: Stray mention of Sinaloa, here.

veracruz-mcu-client/http_test.py
```diff
@@ -0,0 +1,30 @@
...
+print(repr(c.getresponse()))
+
+#c = http.client.HTTPConnection('172.17.0.2', 3017, timeout=3)
```
dominic-mulligan-arm, line 28: Stray comments?

veracruz-mcu-client/main.c
```diff
@@ -0,0 +1,153 @@
+/*
+ * A Veracruz client targetting microcontroller devices
```
dominic-mulligan-arm, line 2: I think this is related to @dreemkiller's questions about about the relationship between the demo and the MCU client itself: whilst this header indicates that this file implements a client, it seems to be entirely demo related.  Can we cleanly separate the two, including moving the demo out of the directory where the client code is defined, or similar?

veracruz-mcu-client/main.c
```diff
@@ -0,0 +1,153 @@
...
+                printf(".");
+            } else {
+                printf(" ");
+            }
+        }
+        printf("\n");
+    }
+}
+
+void dump_gps(int32_t y, int32_t x) {
```
dominic-mulligan-arm, line 71: I think we need comments on these functions to understand what they are doing.

veracruz-mcu-client/policy_to_header.py
```diff
@@ -0,0 +1,225 @@
...
+#
+# ## Authors
+#
+# The Veracruz Development Team.
+#
+# ## Licensing and copyright notice
```
dominic-mulligan-arm, line 7: Can we have a comment in this header explaining the purpose of the file?

veracruz-mcu-client/policy_to_header.py
```diff
@@ -0,0 +1,225 @@
...
+#
+# See the `LICENSE.markdown` file in the Veracruz root directory for
+# information on licensing and copyright.
+#
+
+import json
+import hashlib
+import base64
+
+# TODO fully populate this?
```
dominic-mulligan-arm, line 17: Similar comment as above regarding TODO.

veracruz-mcu-client/policy_to_header.py
```diff
@@ -0,0 +1,225 @@
...
+                    len=len(key_der))
+                for i in range(0, len(key_der), 8):
+                    f.writeln('    %(der)s',
+                        der=' '.join('0x%02x,' % key_der[j]
+                            for j in range(i, min(i+8, len(key_der)))))
+                f.writeln('};')
+
+if __name__ == "__main__":
+    import sys
+    import argparse
```
dominic-mulligan-arm, line 211: Float to the top?

veracruz-mcu-client/prj.conf
```diff
@@ -0,0 +1,69 @@
+# General config
+CONFIG_NEWLIB_LIBC=y
+#CONFIG_MAIN_STACK_SIZE=16384
```
dominic-mulligan-arm, line 3: Stray comment?

veracruz-mcu-client/prj.conf
```diff
@@ -0,0 +1,69 @@
...
+# Networking config
+CONFIG_NETWORKING=y
+CONFIG_NET_IPV4=y
+CONFIG_NET_IPV6=n
+CONFIG_NET_TCP=y
+CONFIG_NET_SOCKETS=y
+CONFIG_NET_SOCKETS_POSIX_NAMES=y
+
+##CONFIG_NET_L2_ETHERNET=y
+##CONFIG_NET_QEMU_ETHERNET=y
```
dominic-mulligan-arm, line 18: Why are some of the configuration options in this file commented out?  Are they stray comments, or do we need to keep them?

veracruz-mcu-client/prj.conf
```diff
@@ -0,0 +1,69 @@
...
+
+# MbedTLS config
+CONFIG_MBEDTLS=y
+CONFIG_MBEDTLS_DEBUG=y
+CONFIG_MBEDTLS_CIPHER_CHACHA20_ENABLED=y
+CONFIG_MBEDTLS_CHACHAPOLY_AEAD_ENABLED=y
+CONFIG_MBEDTLS_MAC_SHA256_ENABLED=y
+CONFIG_MBEDTLS_MAC_POLY1305_ENABLED=y
+CONFIG_MBEDTLS_TLS_VERSION_1_2=y
+# TODO is there a specific ECP to use?
```
dominic-mulligan-arm, line 62: Similar comment to above regarding TODO.

veracruz-mcu-client/run_markdown.sh
```diff
@@ -0,0 +1,36 @@
...
+#
+# The Veracruz Development Team.
+#
+# COPYRIGHT
+#
+# See the `LICENSE.markdown` file in the Veracruz root directory for licensing
+# and copyright information.
+#
+
+# Runs the given markdown file as a bash script, with any failing commands
```
dominic-mulligan-arm, line 12: Hmm, I don't think we should be inventing our own literate programming tool for Bash :).  Can you factor out the inline Bash comments in the Markdown file and add them to an explicit file under version control?

veracruz-mcu-client/slip-setup.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+#
+# ## Authors
+#
+# The Veracruz Development Team.
+#
```
dominic-mulligan-arm, line 6: What is the purpose of this file?  Can you add a comment in this header?

veracruz-mcu-client/tap-setup.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+#
+# ## Authors
+#
+# The Veracruz Development Team.
+#
```
dominic-mulligan-arm, line 6: Similar comment as above about the purpose of this file.

veracruz-mcu-client/tap-setup.sh
```diff
@@ -0,0 +1,19 @@
...
+# ## Licensing and copyright notice
+#
+# See the `LICENSE.markdown` file in the Veracruz root directory for
+# information on licensing and copyright.
+#
+
+ip tuntap add zeth mode tap
+ip link set dev zeth up
+ip link set dev zeth address 00:00:5e:00:53:ff
+ip address add 193.0.2.2/24 dev zeth
```
dominic-mulligan-arm, line 16: What are these IP addresses?  Are they arbitrary?

veracruz-mcu-client/vc.c
```diff
@@ -0,0 +1,577 @@
...
+    };
+
+    // figure out how much of a buffer to allocate, this needs to hold our
+    // sent data + the response, response is fairly small and honestly could
+    // be smaller
+    size_t encoded_size = 0;
+    pb_get_encoded_size(&encoded_size, &Tp_RuntimeManagerRequest_msg, &send_data);
+    size_t proto_len = (32 > encoded_size) ? 32 : encoded_size;
+    // heh
+    uint8_t *proto_buf = malloc(proto_len);
```
dominic-mulligan-arm, line 507: This `malloc` can fail, right?

veracruz-mcu-client/vc.c
```diff
@@ -0,0 +1,577 @@
...
+        }
+
+        ext_seq_ptr += len;
+    }
+
+    printf("no runtime hash?\n");
+    return -EBADE;
+}
+
+int vc_connect(vc_t *vc) {
```
dominic-mulligan-arm, line 285: I second @dreemkiller's call for more comments explaining what some of the functions in this file are doing.  Just as we've adopted `rustdoc` style comments for Rust code, could we adopt `doxygen`-style comments for the C code properly documenting what is going on?

veracruz-mcu-client/vc.h
```diff
@@ -0,0 +1,82 @@
+/*
+ * Veracruz client aimed at microcontrollers, built in Zephyr OS
+ *
+ * Note that there are very few arguments to these functions, the Veracruz
+ * client is based on a static policy file that is expected preprocessed into
+ * a policy.h file (see policy_to_header.py).
+ *
+ * TODO config struct?
```
dominic-mulligan-arm, line 8: Similar comment to above about TODO.

sdk/rust-examples/audio-event-triangulation/src/main.rs
geky: Can do 👍 This was originally written before the WASI changes landed, and I think error reporting in the examples wasn't great then (or I misunderstood it)

sdk/rust-examples/audio-event-triangulation/scripts/poll_for_result.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
geky: Good catch, can do 👍

sdk/rust-examples/audio-event-triangulation/scripts/poll_for_result.py
```diff
@@ -0,0 +1,52 @@
...
+        'N' if location[0] >= 0 else 'S',
+        absx / (1024*1024),
+        (absx / (1024*1024/60)) % 60,
+        (absx / (1024*1024/60/60)) % 60,
+        (absx / (1024*1024/60/60/100)) % 100,
+        'E' if location[1] >= 0 else 'W')
+
+def main(*command):
+    print('\033[0;32mstarting audio event triangulation service\033[m')
+    print('veracruz_server: 172.17.0.2:3010')
```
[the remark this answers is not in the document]
geky: Good catch, these are the standard virtual IP addresses for Docker. There's currently an assumption that the Veracruz Docker is the first Docker you launch on the system (or somehow else configured to 172.17.0.2), I can change this to a parameter

sdk/rust-examples/audio-event-triangulation/scripts/poll_for_result.py
geky: Can do, this is just a quick python script template I often use, sys isn't needed anywhere else except the boilerplate

veracruz-mcu-client/claps_to_header.py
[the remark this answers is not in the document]
geky: It's literally a recording of myself clapping at various distances from a microphone, so we own the copyright. It should be fine to release under the existing Veracruz license.

veracruz-mcu-client/prj.conf
geky: It was just a part of debugging, the commented configurations may be useful if someone attempts to resurrect the qemu-lm4f port, but I don't think we plan to

veracruz-mcu-client/Dockerfile
dominic-mulligan-arm: I think a single cover-all issue is OK, yeah.

veracruz-mcu-client/vc.c
geky: Good catch, can propogate as an error 👍

veracruz-mcu-client/run_markdown.sh
geky: Looks like I'm outvoted here, I can commit a copy of the markdown that is just a bash script.

I'm still of the opinion this is going to be less reliable, the motivation of this is to test the example documentation, and having separate files means they will fall out of sync over time.

geky: Will also need to update the run_cli_test.sh currently in main, can create a separate PR for that

veracruz-mcu-client/tap-setup.sh
geky: This is arbitrary, we're creating our own NAT inside the Docker, which is required to connect the tap interface that passes raw eth packets to Zephyr.

Since it's behind Docker and client side (isn't shared) I don't think it's worth making it configurable, but let me know if you disagree

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: bc9011d18ee3a4be4b5362364a552a94ee60baf8
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=5b5d37bd-d9de-474c-aa9a-59e31e046070%2Fbuild.log) (available for 30 days)

sdk/rust-examples/audio-event-triangulation/src/main.rs
dominic-mulligan-arm: Yeah, I think there's some older examples that just crash if their inputs are not as they expect, which is probably not what we want to happen.  Rather, I think all examples should be total, in the sense that they always return a defined value via one of their output files (modulo bugs), even if that value is an error from an explicit success type.

geky: I was going to suggest we could probably find a tool that would statically determine if the Rust code is total, so we could check for this at compile time, but then I remembered that for code to be panic free in Rust this:

``` rust
a.pow(2) + b.pow(2) + 1
```

Would need to become this:

``` rust
a.checked_pow(2)?.checked_add(b.checked_pow(2)?)?.checked_add(1)?
```

And now I'm not so sure anymore.

dominic-mulligan-arm: Yeah, I think manual auditing of the examples in code review (and going back and rewriting some of them to make them total, in cases where they are not) is fine.

veracruz-mcu-client/vc.c
geky: Done 👍

sdk/rust-examples/audio-event-triangulation/scripts/poll_for_result.py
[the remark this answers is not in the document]
geky: Humorously these were actually only ever used for this print statement, but reading them out from the policy file was easy enough

veracruz-mcu-client/vc.c
geky: Adopted `sys_csrand_get` was an easy drop in replacement. Well except the qemu_cortex_m3 build is now failing due to a linker error even with Zephyr's TEST_RANDOM_GENERATOR enabled... Not sure what that's about so just added a #ifdef dependent on TEST_RANDOM_GENERATOR. (Zephyr outputs a number of warnings at compile time if you enable this).

veracruz-mcu-client/Makefile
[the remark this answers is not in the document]
geky: Moved to `veracruz-mcu-client/samples` which seems to be the idiomatic location for Zephyr modules

veracruz-mcu-client/base64.c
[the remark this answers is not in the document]
geky: It's worth noting that base64 is relatively uncommon in MCU software because it is that much more complicated than hex-encoding.

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 55c3a0fc3de9b408350f820a4877ba1591fa978d
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=fcd27f35-c243-4182-a12c-4a2fbe8e584b%2Fbuild.log) (available for 30 days)

veracruz-mcu-client/vc.c
geky: Fixed 👍

veracruz-mcu-client/vc.c
geky: Wait no, sorry, this should be correct.

It's providing a stream/read/recv API, so if the buffer (`len`) is smaller than the available data `vc->recv_len`, then we return the truncated data but keep the rest of the data in our internal buffer. mbedtls or whatever consumer is on top should call vc_rawrecv repeatedly until all data is consumed.

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: efab06afe43ac40eb5122dc5e33afd93b1377e2f
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=ca0b7cd8-c4f8-4f53-8ebf-9390b13abaaa%2Fbuild.log) (available for 30 days)

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 2475be2d5b6d1a212a222be15df57674562629aa
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=4685f5bd-fe95-4e2c-9849-1e641ae09d5b%2Fbuild.log) (available for 30 days)

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 764cba8bdca988dffea2512947afc0703a40c7e2
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=5b810dc4-74f9-4f99-95b8-83f0f00f09a9%2Fbuild.log) (available for 30 days)

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: caa0efb838a3cf88c439073e68956d4c670df0c2
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=efbbbfe9-da8f-4969-b0c5-4bc74d8f5427%2Fbuild.log) (available for 30 days)

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 806530e7c33e111e354907d5c06385b8c22352ff
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=c4bd673c-e8cf-498a-bf1f-f1ffaf77735f%2Fbuild.log) (available for 30 days)

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 291ebb8a7008227de13e02d13e727dd9a2d341a8
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=af3c753f-5ec5-4815-af32-30f6a327dfb6%2Fbuild.log) (available for 30 days)

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: d1e0ffb3da20797ce091e9192f06e046a96da574
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=3bc49cb2-0d8d-4cad-a60a-f50282a8432f%2Fbuild.log) (available for 30 days)

veracruz-mcu-client/Dockerfile
geky: I think I'm going to actually open up one for each. There's not _that_ many, they can then be closed individually, and labels should help with grouping the related issues together

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 4a78d65d6ea9c9a4e6aa98f1492b31fe7d6e5c59
* Result: FAILED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=392cd782-ea1f-4e9a-b572-f5a9990c0410%2Fbuild.log) (available for 30 days)

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 8c2b331b55494f6a0ea59d6f00306539b2bb4058
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=607a2f77-b320-47e5-8ec0-6b6514ceb4f9%2Fbuild.log) (available for 30 days)

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 139ab2db269a79f2f7f917a30ab1985bdc709fa6
* Result: FAILED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=9f018952-0d25-4ae6-b316-b5498642e729%2Fbuild.log) (available for 30 days)

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 1ff187e1f9fe18381e4d2200566d739e9ffbf6c2
* Result: FAILED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=e6b50cb4-78e8-458c-9f00-05c1102c91cb%2Fbuild.log) (available for 30 days)

geky:
I've created a number of issues to capture the remaining TODO tasks, I don't have permissions to label, so someone else should probably assign the appropriate labels to group them as MCU tasks:

- https://github.com/veracruz-project/veracruz/issues/184
- https://github.com/veracruz-project/veracruz/issues/185
- https://github.com/veracruz-project/veracruz/issues/186
- https://github.com/veracruz-project/veracruz/issues/187
- https://github.com/veracruz-project/veracruz/issues/188
- https://github.com/veracruz-project/veracruz/issues/189

I'm not sure why the build is failing, it seems unrelated to this PR:

```
Caused by:
  failed to fetch submodule `src/llvm-project` from https://github.com/rust-lang/llvm-project.git
```

But other than that, I believe this PR is good to go. I will await more feedback.

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 3c1bcddbbe296bca17877d099f2103dbc124e892
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=4496ee6f-3507-442e-b4d3-e0e5679686ce%2Fbuild.log) (available for 30 days)

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: e40344489c78019b35a2773a50ba1f19a60310bc
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=842c8179-04ed-4498-96cf-bdaff4a25726%2Fbuild.log) (available for 30 days)

veracruz-mcu-client/Makefile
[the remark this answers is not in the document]
dominic-mulligan-arm: Seems to have been resolved, but @dreemkiller should have the final say...

veracruz-mcu-client/tap-setup.sh
dominic-mulligan-arm: No, that's fine, I think.

dominic-mulligan-arm (commented):
This looks good to me now barring:
1. @dreemkiller is happy you've addressed all his comments.
2. All the TODOs are now registered in the issue tracker (or deleted, where no longer needed).
3. We resolve what license `clap.wav` is under.

veracruz-mcu-client/transport_protocol.proto
dreemkiller: `cp -p` should do this. However, is there a problem with just including the `veracruz-transport/` directory in the docker container?

dreemkiller:
I've got one lingering question. When/if that is resolved (the answer may be no), then it's be good to go from my side.

veracruz-mcu-client/transport_protocol.proto
geky: Sorry I forgot to update this one, I've added a step in the Makefile to copy the transport_protocol.proto file during `make docker`.

https://github.com/veracruz-project/veracruz/blob/e40344489c78019b35a2773a50ba1f19a60310bc/veracruz-mcu-client/Makefile#L42-L44

I tried adding a COPY for the file in the Dockerfile, but it didn't work as the file is outside docker's "build context"

I can see if there's a way to change the build context to include all of Veracruz and include the directory if we prefer that.

test-collateral/ca-cert.conf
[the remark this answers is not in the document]
dreemkiller: ECDSA is (mostly) preferred over RSA

dreemkiller (approved):

dominic-mulligan-arm:
@geky did you create the `clap.wav` file yourself?  In which case is the `CC_BY_NA_2.0` license used for some of the photos used in tests OK?  Or did you download it from somewhere?

geky:
Yep, I created it myself. It's a simple recording. I think we could even use the MIT license if we wanted.

Though how do we indicate which file is under which license?

sdk/rust-examples/audio-event-triangulation/src/main.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// Compute the power of the audio event, larger+closer audio
+    /// events result in higher power
+    fn power(&self) -> f64 {
+        // compute as avg per single sample, assumes common bitrate
+        let mut p: f64 = 0.0;
+        for x in self.samples.iter() {
+            p += x.abs() as f64;
+        }
+
+        p / (self.samples.len() as f64)
```
ShaleXIONG, line 52: What happens if `len()` is zero?

ShaleXIONG (approved):

sdk/rust-examples/audio-event-triangulation/src/main.rs
geky: That's a good question, I'll change this to return 0 in this case. A zero-length audio sample mapping to zero power makes sense.

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 11de208822ab62a0e78c3a550b0c913582673566
* Result: FAILED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=631b7641-5cf1-4f7e-9a08-856d73d6be0f%2Fbuild.log) (available for 30 days)

ShaleXIONG (dismissed):

dreemkiller (approved):

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 61a11dfb724e75f555411fcc3dc324815455ed9f
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=cebbab39-fb85-4e17-a812-0c3b2419270e%2Fbuild.log) (available for 30 days)

PR #194, opened 2021-08-13
WIP: Sgx build improvements

dreemkiller:
Some minor changes to the SGX `build.rs` and Makefiles to reduce redundancy in builds and make sure changes get picked up

PR #195, opened 2021-08-13
SGX: Fix build.rs issues causing certain crates to always be rebuilt

geky:
This fixes the always-recompiling issues we were seeing when running veracruz-server-test and veracruz-test, at least for the SGX development flow.

Causes:

- cargo:rerun-if-changed doesn't actually support globs, and if a file
  is missing (the literal '*.rs' in this case), cargo conservatively
  recompiles the crate.

- If cargo:rerun-if-changed targets a crate's root directory, this ends
  up including the "target" directory as a dependency.  And, since target
  changes every compile, the next compile is always out of date,
  triggering a recompile.

These would trigger recompiles for their respective crates, and then for any dependent crates. There still quite a bit of time spent _checking_ dependencies (especially the rust-examples), but it's at least much faster than the actual tests now.


veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 84b28b9e80818e3517e69cc5de76d793d3943cd4
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=ffde4a38-db42-4309-8e56-374e14f30ca5%2Fbuild.log) (available for 30 days)

dominic-mulligan-arm (approved):

dreemkiller:
This is very similar to some of the changes in #194 . 

One point of contention: 
```
If cargo:rerun-if-changed targets a crate's root directory, this ends
up including the "target" directory as a dependency. And, since target
changes every compile, the next compile is always out of date,
triggering a recompile.
```
I haven't found this to be the case.

ShaleXIONG (approved):

geky:
@dreemkiller While trying to find documentation `cargo:rerun-if-changed`'s exact behavior, I found out apparently cargo:rerun-if-changed with a directory just relies on mtime (file creation/rename/permissions).

So it doesn't always happen, but on the first compilation the creation of the target directory does trigger the next recompilation once. 

https://github.com/rust-lang/cargo/issues/2599

At the very least these `cargo:rerun-if-changed`s aren't doing what is expected. If we need build.rs to run on any source file changes it looks like we need to manually generate a list of source files in build.rs.

dreemkiller (approved):

dreemkiller:
@geky Don't for get to sign this. You can sign a previous commit per the following:
https://superuser.com/questions/397149/can-you-gpg-sign-old-commits

geky:
Should be signed now

I also went ahead and added proper globbing with the glob crate for the places we need to depend on `src/*.rs`, so this should do the right thing now hopefully

dreemkiller (approved):

Issue #81, opened 2021-02-11
Isolate platform specific changes to their own crates


dreemkiller:
**Requested feature**
Currently, all supported isolate technologies are supported by the `Sinaloa` and `Mexico City` crates, with features enabled or disabled to support the specific technology.

This has resulted in a confused build process. It also means that the work of adding support for future technologies will create merge conflicts and further complicate the building of these crates.

We should create separate crates for the non-secure side of each of the supported technologies, as well as for the secure-side. This would effectively split out the platform-specific parts of `Sinaloa` and `Mexico City` into multiple, platform-specific crates.

**Motivation**

The current use of features results in code and build processes that are hard to understand and hard to follow. Also, development of support for new platforms is likely to create difficult merge conflicts with other work.

It is desirable to have new platform support to be added (either inside this repository or outside of it) independently of the rest of the work happening on the project.

**Additional context**
This is likely to have knock-on effects elsewhere in the code base. For example,` veracruz-utils` has features to enable support for various platforms. It might be beneficial to have a `utils-common`, and a `utils-*` crate for each supported platform (or at least those that need platform-specific utilities).

nspin:
Needless to say, this would greatly simplify the task of adding support for a new isolate platform for someone who isn't already familiar with the entire source.

When porting to seL4, I found myself `grep`ing through the source for `cfg` attributes to find out what functionality I needed to implement. This improvement would define a clear interface between the platform-independent and platform-dependent code.

dominic-mulligan-arm:
Hopefully, with the "great renaming" almost being complete, this splitting can begin in the next week.

dreemkiller:
Was accidentally closed.

PR #196, opened 2021-08-13
Introduce a Rust workspace for consolidating build dependencies

geky:
This is a bit of exploratory work to try sharing build-time dependencies by a common Rust workspace.

Note that this workspace only includes the high-level (platform-agnostic?) Veracruz libraries. This specifically excludes:

- rust-examples/*, which compile to WebAssembly

- runtime-manager, root-enclaves, sdks which have special requirements

This has also only been tested for SGX, but since most of the platform-specific configuration remains in the Makefile, it should just work for other platforms.

This was a surprisingly unintrusive change, though required a few tweaks:

- Fixed dependencies that disagreed, this is probably a good thing anyways

- Consolidated build profiles. There are package-specific build-profiles, but the only customization we had was for profile.release, and I suspect the crates that didn't customize profile.release were just never compiled outside of debug mode.

- Consolidated patches. This is probably a good thing as I suspect that the patches were required for all dependent crates, and this will reduce mistakes when adding new crates in the future.

Some other thoughts:

- We've discussed breaking the workspaces out into per-platform workspaces + symlinks. This doesn't seem like a strict requirement, since you can still build each package with its own feature/profile/etc flags. But it may be nice for organizational reasons, and to avoid future risks with conflicting dependencies/patches between platforms.

  Though this may be something that gets cleaned up nicely by https://github.com/veracruz-project/veracruz/issues/81. If each platform is its own crate/workspace, _inside_ the high-level, common Veracruz crates/workspace, then we would get the separation of per-platform configuration without need to use symlinks. It would result in some build-dependency duplication, but only worst case 2x vs Nx.

- You may think you can use the documented package shorthand in the root `cargo build --package veracruz-test --features sgx`, but you can't. At least now with the --features flag. You can, however, provide --features to builds in the directory or with --manifest-path `cargo build --manifest-path veracruz-test/Cargo.toml --features sgx`, which does use the workspace as any local builds search upwards for a workspace Cargo.toml. Not sure why this limitation exists.

  https://github.com/rust-lang/cargo/issues/5015
  
  If you went the per-platform workspace direction, you could specify these feature flags as "default-features" in the workspace.

- I also just wanted to clarify that both the relative path dependencies and non-root builds do work with workspaces and do use the common target directory. Which is why this required so few changes.

- I also considered moving the rust-examples into a WebAssembly-specific workspace, but didn't as I'm not sure there's that much of an advantage there. The examples don't have that many dependencies, are each intended to be standalone (different build profiles?), and adopting a slightly more complex workspace system in the examples risks confusing users who just copy-paste an example to get started.

- The workspace does include some non-root crates: freestanding-execution-engine, wasm-checker, generate-policy. These did have some dependencies on the rest of Veracruz, so I wonder if the sdk/test-collateral directories are the wrong place for them and they should be moved into root.

- The data-generators are included in the workspace, I don't really have any thoughts on these.


veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: a3da971de377a541734cfebcaadda3c44850723b
* Result: FAILED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=9216038e-ad47-4ed6-88fb-4b1d37d797e3%2Fbuild.log) (available for 30 days)

geky:
Oh, I also didn't touch the Makefile, but in theory the multi-project rules (make clean, make update, make fmt), can be simplified now

veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: 74131ffd92c0ccea19085307f59e2c5fc607b4c0
* Result: FAILED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=6bf0500a-6cf3-45d7-9f09-3ccf2da8a819%2Fbuild.log) (available for 30 days)

dominic-mulligan-arm:
@geky can this be closed given @mathias-arm's work?

geky:
Yep, closing for now. The code/branch is still available in case it's useful

PR #197, opened 2021-08-14
Add `make quick-clean` rule to clean everything but LLVM

geky:
Consider adding a rule that cleans everything but LLVM.

This build rule cleans everything but wasm-checker, which seems to be the only library dependent on LLVM. It vastly dominates the build time but is rarely the source of compilation issues.

This could isolate the LLVM build itself and avoid compiling that, but I didn't dig deep enough into wasm-checker to know where this occurs.

Alternatively this could be inverted, making this the behavior of `make clean`, and keeping the existing clean behavior as `make slow-clean`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -302,6 +302,31 @@ clean:
 	$(MAKE) clean -C linux-root-enclave
 	rm -rf bin
 
+# clean-quick cleans everything but LLVM (in wasm-checker)
+quick-clean:
+	# remove databases since these can easily fall out of date
+	rm -f proxy-attestation-server/proxy-attestation-server.db
+	rm -f veracruz-server-test/proxy-attestation-server.db
+	rm -f veracruz-test/proxy-attestation-server.db
+	# clean code
+	cd runtime-manager-bind && cargo clean 
+	cd sgx-root-enclave-bind && cargo clean
+	cd psa-attestation && cargo clean
+	cd proxy-attestation-server && cargo clean
+	cd session-manager && cargo clean
+	cd veracruz-utils && cargo clean
+	cd veracruz-server-test && cargo clean
+	cd veracruz-test && cargo clean && rm -f proxy-attestation-server.db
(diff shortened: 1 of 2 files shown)


veracruz-project-owner:
### AWS CodeBuild CI Report

* CodeBuild project: Veracruz
* Commit ID: b5d014016ed1e59953fb69cbdd97e07b5fe1c647
* Result: SUCCEEDED
* [Build Logs](https://01evtei465.execute-api.eu-west-2.amazonaws.com/Prod/buildlogs?key=4d83994c-014b-414d-95fd-4de7888ccb42%2Fbuild.log) (available for 30 days)

dreemkiller:
I think this brings up a bigger question: Is the wasm checker providing enough value to merit keeping it?

dominic-mulligan-arm:
I think it's probably still valuable as:

1. Some toolchains (e.g. `emscripten`) insert non-conforming host-calls into the generated Wasm binaries, even when compiling specifically for WASI, to handle e.g. C++ exceptions, and similar.
2. There's a surprising degree of manual "grepping" of binaries still going on when first compiling previously-unused libraries to Wasm.  This is done to check that the generated host-calls produced are supported by the Veracruz runtime (e.g. a library is not depending on an implementation of `epoll` or similar, which we do not support).  @ShaleXIONG mentioned doing this to me, this morning, for example, with some application to what @mofanv is currently working on.  Arguably, this could be done using `wasmchecker` rather than manually converting a binary to a `wast` file and inspecting it in a text editor, which is the practice now.

ShaleXIONG (dismissed):

dominic-mulligan-arm:
@geky can you rewrite history to sign the commits and rebase to reflect `trustzone-root-enclave` being removed and `linux-root-enclave` being added?

geky:
@dominic-mulligan-arm should be updated now, and changed to `make quick-clean`

dominic-mulligan-arm (approved):

PR #203, opened 2021-08-25
General code cleanup and dead code elimination

dominic-mulligan-arm:
A pass to reduce the number of compile-time warnings produced by the project, remove deprecated functions that are no longer required due to our new-found embrace of WASI, and so on.

- [x] Remove all functionality marked as deprecated, related to the old programming model and provisioning process.
- [x] Remove all functions that are not used when compiling for any of our targets.
- [x] Remove redundant imports.
- [x] Code health boost by running `cargo clippy` on project components.

(I was originally going to perform a `cargo fmt` pass as well.  Given the size of this changeset, and the sensitivity of some of the changes, I'll defer a `cargo fmt` pass to another pull request.)


dominic-mulligan-arm:
Whilst this is ready for review, it's best to hold off reviewing this until the Rouille changes are merged and I've rebased this on top of `main`.

dominic-mulligan-arm:
Can somebody take a look at this?

dominic-mulligan-arm:
@gbryant-arm can you review this?

veracruz-client/src/veracruz_client.rs
```diff
@@ -381,105 +384,98 @@ impl VeracruzClient {
...
+        let mut output: Vec<u8> = Vec::new();
+
+        if !input.is_empty() && (!self.tls_session.is_handshaking() || self.tls_session.wants_read()) {
             let mut slice = &input[..];
-            self.tls_session.read_tls(&mut slice)?;
 
+            self.tls_session.read_tls(&mut slice)?;
             self.tls_session.process_new_packets()?;
         }
+
```
gbryant-arm, line 449: Is skipping line between blocks something that clippy recommends?

veracruz-server-test/src/main.rs
```diff
@@ -702,10 +701,12 @@ mod tests {
     ) -> Result<(), VeracruzServerError> {
         info!("### Step 0.  Initialise test configuration.");
         // initialise the pipe
+        #[allow(clippy::type_complexity)]
```
gbryant-arm, line 704: Should we have a global clippy configuration instead of using attributes? Cf. https://github.com/rust-lang/rust-clippy/blob/master/README.md#configuration

veracruz-test/src/main.rs
```diff
@@ -444,3 +444,7 @@ mod tests {
         return Ok(policy_text);
     }
 }
+
+// NB: dummy main function to satify Clippy, which raises a hard error otherwise.
```
gbryant-arm, line 448: Typo on "satify"

runtime-manager/src/runtime_manager_sgx.rs
```diff
@@ -50,17 +50,20 @@ extern "C" {
 
 #[no_mangle]
 #[cfg(feature = "sgx")]
-pub extern "C" fn init_session_manager_enc(
+pub unsafe extern "C" fn init_session_manager_enc(
```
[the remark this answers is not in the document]
dominic-mulligan-arm: Clippy seems to strongly prefer `unsafe` functions over `unsafe` blocks.  To be honest, in hindsight I think that this is a mistake, and we should have a dedicated Clippy configuration, as you suggest below, tailored to the project.  I will change this back.

sgx-root-enclave/src/lib.rs
```diff
@@ -55,7 +55,7 @@ pub extern "C" fn get_firmware_version_len(p_fwv_len: &mut usize) -> sgx_status_
 }
 
 #[no_mangle]
-pub extern "C" fn get_firmware_version(
+pub unsafe extern "C" fn get_firmware_version(
```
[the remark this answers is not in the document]
dominic-mulligan-arm: No, once a function is marked `unsafe` then any unsafe code contained within it need not have an `unsafe` block.

veracruz-client/src/veracruz_client.rs
dominic-mulligan-arm: No, that's something I prefer when reading code.

veracruz-server-test/src/main.rs
dominic-mulligan-arm: Yes, I think this is a good idea.

transport-protocol/src/custom.rs
```diff
@@ -14,8 +14,6 @@ use crate::transport_protocol;
 use core::convert::TryInto;
 use err_derive::Error;
 use protobuf::{error::ProtobufError, Message, ProtobufEnum};
-#[cfg(feature = "sgx_attestation")]
-use sgx_types;
```
ShaleXIONG, line 18: If there is a CI problem in sgx, this line might need to stay.

ShaleXIONG (approved):
`runtime-manager/runtime_manager_enclave` should not be checked in?

transport-protocol/src/custom.rs
dominic-mulligan-arm: The SGX build seems to complete, and (from memory) all of the SGX tests passed, too, though I will double check.

dominic-mulligan-arm:
> `runtime-manager/runtime_manager_enclave` should not be checked in?

Removed.

gbryant-arm:
Regarding the unsuccessful check:
```
23 | pub fn platform_getrandom(buffer: &mut [u8]) -> result::Result {
```

 `result::Result` now embeds a type. Returning `result::Result<()>` should fix it. Rebase error?

PR #210, opened 2021-09-02
Added pre-commit hook for code formatting checking

dominic-mulligan-arm:
Add pre-commit hook for code formatting checking.  Changes include:

1. Added pre-commit Git hook for code-formatting checking in new `githooks` directory.  Note that Git prevents the local `.git/hooks/` directory from being put under source control, so if we want to track these project-wide hooks then we need to introduce another directory to contain them.
2. Added a new Makefile target `setup-githooks` which installs `rustfmt`, if not already installed, and points Git to the `githooks` directory.
3. Updated `CONTRIBUTING.markdown` to document this hook, and how to set it up, locally.

### DIFF
--- a/githooks/pre-commit
+++ b/githooks/pre-commit
@@ -1,0 +1,39 @@
+#! /bin/bash
+#
+# A pre-commit Git hook.
+#
+# Currently checks the formatting of all Rust files staged for commit,
+# failing if any need to be reformatted using `rustfmt`.  Note that the
+# hook does not modify any file.
+#
+# Authors
+#
+# The Veracruz Development Team.
+#
+# Copyright and licensing
+#
+# See the `LICENSE_MIT.markdown` file in the Veracruz repository root
+# directory for copyright and licensing information.
+
+HAS_ISSUES=0
+
+echo "Checking code formatting of files staged for commit."
+
+for file in $(git diff --name-only --staged \*.rs); do
+    RUSTFMT="$(rustfmt --check --skip-children --unstable-features $file)"
+    if [ "$RUSTFMT" != "" ]; then
+        printf "[ERROR]: $file\n"
+        HAS_ISSUES=1
+    else
+        printf "[   OK]: $file\n"
+    fi
+done
+
+if [ $HAS_ISSUES -eq 0 ]; then
+    echo "Code formatting style is OK."
+    exit 0
+fi
+
+echo "There are formatting issues in all of the files marked with ERROR above (if any)."
+echo "First format your code with \`make fmt\` or call \`rustfmt\` manually before committing."
+exit 1
--- a/CONTRIBUTING.markdown
+++ b/CONTRIBUTING.markdown
@@ -53,7 +53,11 @@ into the main Veracruz codebase by issuing a pull request through Github.
 Before submitting any pull request, please ensure that:
 1. You have run `cargo fmt` on any changes.  A project-wide `cargo fmt`
    can be executed by building the `fmt` target in the main project
-   Makefile.
+   Makefile.  A Git pre-commit hook is available in the `githooks`
+   directory can be used to automatically check Rust files are formatted
+   correctly before they are committed.  To enable this hook, run the
+   `make setup-githooks` command which will automatically install
+   `rustfmt`, if needed, and set the Git hooks directory to `githooks`.
 2. The new changes are clearly commented using `rustdoc` markup, for
    Rust code, or similar for whatever language your changes are written
    in.
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@
 # See the `LICENSE_MIT.markdown` file in the Veracruz root directory for licensing
 # and copyright information.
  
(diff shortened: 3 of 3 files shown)


githooks/pre-commit
```diff
@@ -0,0 +1,44 @@
...
+        printf "[✅]: $file\n"
+    fi
+done
+
+if [ $HAS_ISSUES -eq 0 ]; then
+    echo "Code formatting style is OK."
+    exit 0
+fi
+
+echo "There are formatting issues in the files makred with a cross (❌) above."
```
gbryant-arm, line 42: Typo ("makred")

dominic-mulligan-arm: Thanks, fixed.

egrimley-arm:
The log message seems to be misformatted: the top half has rather long lines, and the bottom half seems to have been accidentally "filled". This problem is easier to see with `git log` in a standard xterm than on github.com. (But I'm not quite sure how GitHub generates the final log message if you use "Squash and merge".)

dominic-mulligan-arm:
@egrimley-arm thanks, not sure what had happened there.  I've amended the commit.  Is that better?

githooks/pre-commit
```diff
@@ -0,0 +1,44 @@
...
+FIRST_FILE=1
+
+echo "Checking code formatting of files staged for commit."
+
+for file in $(git diff --name-only --staged *.rs); do
+    RUSTFMT="$(rustfmt --check --skip-children --unstable-features $file)"
+    if [ "$RUSTFMT" != "" ]; then
+        if [ $FIRST_FILE -eq 0 ]; then
+            echo -n ", "
+        fi
```
egrimley-arm, line 28: There are a couple of trailing spaces here, in case you care.

githooks/pre-commit
```diff
@@ -0,0 +1,44 @@
...
+
+echo "Checking code formatting of files staged for commit."
+
+for file in $(git diff --name-only --staged *.rs); do
+    RUSTFMT="$(rustfmt --check --skip-children --unstable-features $file)"
+    if [ "$RUSTFMT" != "" ]; then
+        if [ $FIRST_FILE -eq 0 ]; then
+            echo -n ", "
+        fi  
+        printf "[❌]: $file\n"
```
egrimley-arm, line 29: Unfortunately ❌ isn't displayed properly with the `xterm` I'm using here, so this could be problematic. Apparently it belongs to the "Dingbats" block. I don't think you necessarily have to stick to ASCII in this day and age but it might be a good idea to go with something slightly less exotic.

githooks/pre-commit
```diff
@@ -0,0 +1,44 @@
...
+#
+# See the `LICENSE_MIT.markdown` file in the Veracruz repository root
+# directory for copyright and licensing information.
+
+HAS_ISSUES=0
+FIRST_FILE=1
+
+echo "Checking code formatting of files staged for commit."
+
+for file in $(git diff --name-only --staged *.rs); do
```
egrimley-arm, line 23: You need a `\*.rs` instead of `*.rs`, I think.

githooks/pre-commit
```diff
@@ -0,0 +1,44 @@
...
+    RUSTFMT="$(rustfmt --check --skip-children --unstable-features $file)"
+    if [ "$RUSTFMT" != "" ]; then
+        if [ $FIRST_FILE -eq 0 ]; then
+            echo -n ", "
```
egrimley-arm, line 27: That comma comes out at the start of a line and breaks the column. I don't think commas are needed if each file has its own line.

egrimley-arm (commented):
I few issues I've found.

githooks/pre-commit
egrimley-arm: It looks as though the terminal I'm using here really does only (UTF-8-encoded) ISO-8859-1.

dominic-mulligan-arm: I've removed the Unicode entirely in place of an ASCII message, instead.

githooks/pre-commit
dominic-mulligan-arm: This if-statement has been deleted, now.

githooks/pre-commit
dominic-mulligan-arm: This has been removed.

egrimley-arm:
> @egrimley-arm thanks, not sure what had happened there. I've amended the commit. Is that better?

Yes, though you have lines with 84 characters. Some people think log messages should only have 72 characters in a line!

dominic-mulligan-arm:
Ha!  Well, I don't think we have a project-wide policy on commit message line width, and I think 72 characters is overkill in 2021.  Some messages in the `git log` are quite wide...

dominic-mulligan-arm:
@egrimley-arm are you happy with this now?  If so, can you approve it?

egrimley-arm (approved):
LGTM

gbryant-arm (approved):
LGTM too

dominic-mulligan-arm:
+1+1 = +1 reached, merging.

PR #212, opened 2021-09-07
freestanding-execution-engine: Avoid broken default_value feature.

egrimley-arm:
To avoid confusion, we want -o, -e, ... without a value to give an
error, but the default_value feature of the clap crate is broken,
so use unwrap_or instead.

### DIFF
--- a/sdk/freestanding-execution-engine/src/main.rs
+++ b/sdk/freestanding-execution-engine/src/main.rs
@@ -64,10 +64,10 @@ const OUTPUT_FILE: &'static str = "output";
 
 /// The default dump status of `stdout`, if no alternative is provided on the
 /// command line.
-const DEFAULT_DUMP_STDOUT: bool = false;
+const DEFAULT_DUMP_STDOUT: &'static str = "false";
 /// The default dump status of `stderr`, if no alternative is provided on the
 /// command line.
-const DEFAULT_DUMP_STDERR: bool = false;
+const DEFAULT_DUMP_STDERR: &'static str = "false";
 
 ////////////////////////////////////////////////////////////////////////////////
 // Command line options and parsing.
@@ -91,9 +91,6 @@ struct CommandLineOptions {
 /// of them.  If required options are not present, or if any options are
 /// malformed, this will abort the program.
 fn parse_command_line() -> Result<CommandLineOptions, Box<dyn Error>> {
-
-    let default_dump_stdout = DEFAULT_DUMP_STDOUT.to_string();
-    let default_dump_stderr = DEFAULT_DUMP_STDERR.to_string();
 
     let matches = App::new(APPLICATION_NAME)
         .version(VERSION)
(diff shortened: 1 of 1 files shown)


sdk/freestanding-execution-engine/src/main.rs
```diff
@@ -124,49 +121,45 @@ fn parse_command_line() -> Result<CommandLineOptions, Box<dyn Error>> {
...
 
     info!("Parsed command line.");
 
-    let execution_strategy = if let Some(strategy) = matches.value_of("execution-strategy") {
-        if strategy == "interp" {
+    let execution_strategy = {
+        let strategy = matches.value_of("execution-strategy").unwrap_or("jit");
+        match strategy {
+            "interp" => {
             info!("Selecting interpretation as the execution strategy.");
```
gbryant-arm, line 148: Indentation is off starting from here

gbryant-arm (changes requested):
LGTM besides the issue with indentation

sdk/freestanding-execution-engine/src/main.rs
egrimley-arm: I've fixed that, I hope. But I did `git push --force`, which perhaps wasn't the best way. Or perhaps it doesn't matter.

sdk/freestanding-execution-engine/src/main.rs
```diff
@@ -186,28 +178,30 @@ fn parse_command_line() -> Result<CommandLineOptions, Box<dyn Error>> {
         Vec::new()
     };
 
-    let dump_stdout = if let Some(dump_stdout) = matches.value_of("dump-stdout") {
+    let dump_stdout = {
+        let default_dump_stdout = DEFAULT_DUMP_STDOUT.to_string();
```
dominic-mulligan-arm, line 182: I don't think this `let` binding is required.  You can just inline `DEFAULT_DUMP_STDOUT.to_string()`.

sdk/freestanding-execution-engine/src/main.rs
```diff
@@ -186,28 +178,30 @@ fn parse_command_line() -> Result<CommandLineOptions, Box<dyn Error>> {
...
             return Err(format!("Expecting a boolean, but found {}", dump_stdout)
             .into());
         }
-    } else {
-        return Err("Default 'dump-stdout' value is not loaded correctly".into());
     };
 
-    let dump_stderr = if let Some(dump_stderr) = matches.value_of("dump-stderr") {
+    let dump_stderr = {
+        let default_dump_stderr = DEFAULT_DUMP_STDERR.to_string();
```
dominic-mulligan-arm, line 195: Similar comment here about inlining of `let`.

sdk/freestanding-execution-engine/src/main.rs
```diff
@@ -124,49 +121,44 @@ fn parse_command_line() -> Result<CommandLineOptions, Box<dyn Error>> {
...
+        match strategy {
+            "interp" => {
+                info!("Selecting interpretation as the execution strategy.");
+                ExecutionStrategy::Interpretation
+            }
+            "jit" => {
+                info!("Selecting JITting as the execution strategy.");
+                ExecutionStrategy::JIT
+            }
+            _ => {
```
dominic-mulligan-arm, line 155: Isn't this case now impossible, due to `strategy` defaulting to `jit` in the `unwrap_or()`?  If so, I would explicitly mark this as impossible with `unreachable!()`.

dominic-mulligan-arm (changes requested):

sdk/freestanding-execution-engine/src/main.rs
egrimley-arm: If I inline it in the obvious way I get:
```
error[E0716]: temporary value dropped while borrowed
   --> src/main.rs:184:25
    |
184 |             .unwrap_or(&DEFAULT_DUMP_STDOUT.to_string());
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ - temporary value is freed at the end of this statement
    |                         |
    |                         creates a temporary which is freed while still in use
185 |         if let Ok(dump_stdout) = bool::from_str(dump_stdout) {
    |                                                 ----------- borrow later used here
    |
    = note: consider using a `let` binding to create a longer lived value
```

There might be a less obvious way of inlining it.

sdk/freestanding-execution-engine/src/main.rs
egrimley-arm: It can happen if the user specifies something else:
```
$ cargo run -- --program read-file.wasm -x foobar
    Finished dev [unoptimized + debuginfo] target(s) in 0.58s
     Running `target/debug/freestanding-execution-engine --program read-file.wasm -x foobar`
Error: "Expecting \'interp\' or \'jit\' as selectable execution strategies, but found foobar"
```

(There are too many quotes in the error message, really, but all the error messages are generated in the same way.)

sdk/freestanding-execution-engine/src/main.rs
dominic-mulligan-arm: I think you can just do `unwrap_or(DEFAULT_DUMP_STDOUT)`.  The constant is already of type `&str`, which is then converted to `String`, and then a slice is taken again to get another `&str`.

sdk/freestanding-execution-engine/src/main.rs
dominic-mulligan-arm: Ah OK, yeah fair enough!

sdk/freestanding-execution-engine/src/main.rs
egrimley-arm: `DEFAULT_DUMP_STDOUT` is currently a `bool`. But I could change it into a `&'static str`. Then that would work.

dominic-mulligan-arm: Ah sorry, yeah, I'm causing confusion.  Yes, I think `DEFAULT_DUMP_STDOUT` should be a `&str` as we only ever work with it in string form, anyway, and never rely on it being of type `bool`, as far as I can see.

egrimley-arm: I've pushed an update.

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
LGTM.  Thanks, @egrimley-arm!  Can you open a second PR with your name added to the `AUTHORS.markdown` file?

dominic-mulligan-arm:
@gbryant-arm do you still need formatting changes?

gbryant-arm (approved):
LGTM

dominic-mulligan-arm:
+1+1 = +1 reached, merging.

gbryant-arm:
FWIW clap is used in other parts of the codebase, e.g. in `test-collateral/generate-policy/src/main.rs`

egrimley-arm:
> FWIW clap is used in other parts of the codebase, e.g. in `test-collateral/generate-policy/src/main.rs`

I've just submitted a PR for that, which also had other problems: #216

The other uses of `default_value` are in `sdk/data-generators/` and may be less important.

PR #213, opened 2021-09-07
Add clock-related functions to platform-services, conditionally implement them in WASI ABI

gbryant-arm:
Implement clock functions (`clock_getres()` and `clock_gettime()`) for the standard platform service.
The SGX, Nitro and Trustzone platform services don't implement those functions yet and return `result::Result::Unavailable`. They will be implemented later on in a separate PR.
`result::Result::Success` has been changed to include the time in nanoseconds (u64).

TODO:
+ ~~Link clock functions to policies: Disable them by default, unless specified otherwise~~
+ ~~Rebase on top of @egrimley-arm 's work on clap (#212 and #216)~~
+ ~~Fixing rebase issues~~
+ ~~Return more appropriate error codes~~

### DIFF
--- a/execution-engine/src/fs.rs
+++ b/execution-engine/src/fs.rs
@@ -339,13 +339,14 @@ impl FileSystem {
     #[inline]
     pub(crate) fn clock_res_get(&self, clock_id: ClockId) -> FileSystemResult<Timestamp> {
         if !self.enable_clock {
-            return Err(ErrNo::NoSys);
+            return Err(ErrNo::Access);
         }
 
         let clock_id = u32::from(clock_id) as u8;
         match getclockres(clock_id) {
             result::Result::Success(resolution) => Ok(Timestamp::from_nanos(resolution)),
-            _ => Err(ErrNo::NoSys),
+            result::Result::Unavailable => Err(ErrNo::NoSys),
+            _ => Err(ErrNo::Inval),
         }
     }
 
@@ -357,13 +358,14 @@ impl FileSystem {
         _precision: Timestamp,
     ) -> FileSystemResult<Timestamp> {
         if !self.enable_clock {
-            return Err(ErrNo::NoSys);
+            return Err(ErrNo::Access);
         }
 
         let clock_id = u32::from(clock_id) as u8;
         match getclocktime(clock_id) {
             result::Result::Success(timespec) => Ok(Timestamp::from_nanos(timespec)),
-            _ => Err(ErrNo::NoSys),
+            result::Result::Unavailable => Err(ErrNo::NoSys),
+            _ => Err(ErrNo::Inval),
         }
     }
 
--- a/execution-engine/src/fs.rs
+++ b/execution-engine/src/fs.rs
@@ -142,10 +142,7 @@ impl FileSystem {
     ////////////////////////////////////////////////////////////////////////
 
     /// Install standard streams (`stdin`, `stdout`, `stderr`).
-    fn install_standard_streams(
-        &mut self,
-        std_streams_table: &Vec<StandardStream>,
-    ) {
+    fn install_standard_streams(&mut self, std_streams_table: &Vec<StandardStream>) {
         for std_stream in std_streams_table {
             // Map each standard stream to an fd and inode.
             // Rights are assumed to be already configured by the execution engine in the rights table
--- a/execution-engine/src/fs.rs
+++ b/execution-engine/src/fs.rs
@@ -920,7 +920,7 @@ impl FileSystem {
     #[inline]
     pub(crate) fn random_get(&self, buf_len: Size) -> FileSystemResult<Vec<u8>> {
         let mut buf = vec![0; buf_len as usize];
-        if let result::Result::Success(_) = getrandom(&mut buf) {
+        if getrandom(&mut buf).is_success() {
             Ok(buf)
         } else {
             Err(ErrNo::NoSys)
--- a/execution-engine/src/fs.rs
+++ b/execution-engine/src/fs.rs
@@ -96,6 +96,8 @@ pub struct FileSystem {
     rights_table: RightsTable,
     /// Preopen FD table. Mapping the FD to dir name.
     prestat_table: HashMap<Fd, PathBuf>,
+    /// Whether clock functions (`clock_getres()`, `clock_gettime()`) should be enabled.
+    enable_clock: bool,
 }
 
 impl FileSystem {
@@ -121,6 +123,7 @@ impl FileSystem {
     pub fn new(
         rights_table: RightsTable,
         std_streams_table: &Vec<StandardStream>,
+        enable_clock: bool,
     ) -> Self {
         let mut rst = Self {
             fd_table: HashMap::new(),
@@ -128,6 +131,7 @@ impl FileSystem {
             inode_table: HashMap::new(),
             rights_table,
             prestat_table: HashMap::new(),
+            enable_clock,
         };
         rst.install_prestat::<&Path>(&Vec::new(), std_streams_table);
         rst
@@ -337,6 +341,10 @@ impl FileSystem {
     /// Get clock resolution from platform services. Return error `NoSys` in case of failure.
     #[inline]
     pub(crate) fn clock_res_get(&self, clock_id: ClockId) -> FileSystemResult<Timestamp> {
+        if !self.enable_clock {
+            return Err(ErrNo::NoSys);
+        }
+
         let clock_id = u32::from(clock_id) as u8;
         match getclockres(clock_id) {
             result::Result::Success(resolution) => Ok(Timestamp::from_nanos(resolution)),
@@ -351,6 +359,10 @@ impl FileSystem {
         clock_id: ClockId,
         _precision: Timestamp,
     ) -> FileSystemResult<Timestamp> {
+        if !self.enable_clock {
+            return Err(ErrNo::NoSys);
+        }
+
         let clock_id = u32::from(clock_id) as u8;
         match getclocktime(clock_id) {
             result::Result::Success(timespec) => Ok(Timestamp::from_nanos(timespec)),
--- a/execution-engine/src/fs.rs
+++ b/execution-engine/src/fs.rs
@@ -14,7 +14,7 @@
 //! See the `LICENSE_MIT.markdown` file in the Veracruz root directory for
 //! information on licensing and copyright.
 
-use platform_services::{getrandom, result};
+use platform_services::{getclockres, getclocktime, getrandom, result};
 use std::{
     collections::HashMap,
     convert::{AsRef, TryInto},
@@ -334,20 +334,28 @@ impl FileSystem {
     // Operations on the filesystem. Rust style implementation of WASI API
     ////////////////////////////////////////////////////////////////////////////
 
-    /// The stub implementation of `clock_res_get`. Return unsupported error `NoSys`.
+    /// Get clock resolution from platform services. Return error `NoSys` in case of failure.
     #[inline]
-    pub(crate) fn clock_res_get(&self, _clock_id: ClockId) -> FileSystemResult<Timestamp> {
-        Err(ErrNo::NoSys)
+    pub(crate) fn clock_res_get(&self, clock_id: ClockId) -> FileSystemResult<Timestamp> {
+        let clock_id = u32::from(clock_id) as u8;
+        match getclockres(clock_id) {
+            result::Result::Success(resolution) => Ok(Timestamp::from_nanos(resolution)),
+            _ => Err(ErrNo::NoSys),
+        }
     }
 
-    /// The stub implementation of `clock_time_get`. Return unsupported error `NoSys`.
+    /// Get clock time from platform services. Return error `NoSys` in case of failure.
     #[inline]
     pub(crate) fn clock_time_get(
         &self,
-        _clock_id: ClockId,
+        clock_id: ClockId,
         _precision: Timestamp,
     ) -> FileSystemResult<Timestamp> {
-        Err(ErrNo::NoSys)
+        let clock_id = u32::from(clock_id) as u8;
+        match getclocktime(clock_id) {
+            result::Result::Success(timespec) => Ok(Timestamp::from_nanos(timespec)),
+            _ => Err(ErrNo::NoSys),
+        }
     }
 
     /// Allows the programmer to declare how they intend to use various parts of
@@ -912,7 +920,7 @@ impl FileSystem {
     #[inline]
     pub(crate) fn random_get(&self, buf_len: Size) -> FileSystemResult<Vec<u8>> {
         let mut buf = vec![0; buf_len as usize];
-        if let result::Result::Success = getrandom(&mut buf) {
+        if let result::Result::Success(_) = getrandom(&mut buf) {
             Ok(buf)
         } else {
             Err(ErrNo::NoSys)
--- a/platform-services/src/lib.rs
+++ b/platform-services/src/lib.rs
@@ -39,7 +39,7 @@ cfg_if! {
     } else if #[cfg(feature = "std")] {
         #[path="std_platform_services.rs"]
         mod imp;
-    
+
     } else {
         compile_error!(
             "Unrecognised feature: platforms supported are SGX, TZ, Nitro, and std.");
--- a/platform-services/src/lib.rs
+++ b/platform-services/src/lib.rs
@@ -78,6 +78,7 @@ pub fn getrandom(buffer: &mut [u8]) -> result::Result<()> {
 ///     - `result::Result::Unavailable` if the specified clock is not available
 ///       on this platform.
 ///     - `result::Result::UnknownError` if a runtime error occurred.
+#[inline]
 pub fn getclockres(clock_id: u8) -> result::Result<u64> {
     imp::platform_getclockres(clock_id)
 }
@@ -90,6 +91,7 @@ pub fn getclockres(clock_id: u8) -> result::Result<u64> {
 ///     - `result::Result::Unavailable` if the specified clock is not available
 ///       on this platform.
 ///     - `result::Result::UnknownError` if a runtime error occurred.
+#[inline]
 pub fn getclocktime(clock_id: u8) -> result::Result<u64> {
     imp::platform_getclocktime(clock_id)
 }
--- a/platform-services/src/lib.rs
+++ b/platform-services/src/lib.rs
@@ -62,10 +62,34 @@ cfg_if! {
 ///     - `result::Result::UnknownError` if a runtime error occurred during
 ///       generation of the random numbers.  In which case, the contents of
 ///       `buffer` are undefined.
-pub fn getrandom(buffer: &mut [u8]) -> result::Result {
+pub fn getrandom(buffer: &mut [u8]) -> result::Result<()> {
     if buffer.is_empty() {
-        return Result::Success;
+        return Result::Success(());
     } else {
         imp::platform_getrandom(buffer)
     }
+}
+
+/// Gets resolution from the specified clock.
+///
+/// Returns:
+///     - `result::Result::Success`, embedding the clock resolution in
+///       nanoseconds, if the operation successfully completed.
+///     - `result::Result::Unavailable` if the specified clock is not available
+///       on this platform.
+///     - `result::Result::UnknownError` if a runtime error occurred.
+pub fn getclockres(clock_id: u8) -> result::Result<u64> {
+    imp::platform_getclockres(clock_id)
+}
+
+/// Gets time from the specified clock.
+///
+/// Returns:
+///     - `result::Result::Success`, embedding the clock time in
+///       nanoseconds, if the operation successfully completed.
+///     - `result::Result::Unavailable` if the specified clock is not available
+///       on this platform.
+///     - `result::Result::UnknownError` if a runtime error occurred.
+pub fn getclocktime(clock_id: u8) -> result::Result<u64> {
+    imp::platform_getclocktime(clock_id)
 }
--- a/platform-services/src/nitro_platform_services.rs
+++ b/platform-services/src/nitro_platform_services.rs
@@ -18,7 +18,7 @@ use nsm_lib;
 
 /// Fills a buffer, `buffer`, with random bytes sampled from the thread-local
 /// random number source.  Uses the AWS Nitro RNG
-pub fn platform_getrandom(buffer: &mut [u8]) -> result::Result {
+pub fn platform_getrandom(buffer: &mut [u8]) -> result::Result<()> {
     let nsm_fd = nsm_lib::nsm_lib_init();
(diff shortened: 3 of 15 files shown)


execution-engine/src/fs.rs
```diff
@@ -334,20 +334,28 @@ impl FileSystem {
     // Operations on the filesystem. Rust style implementation of WASI API
     ////////////////////////////////////////////////////////////////////////////
 
-    /// The stub implementation of `clock_res_get`. Return unsupported error `NoSys`.
+    /// Get clock resolution from platform services. Return error `NoSys` in case of failure.
     #[inline]
-    pub(crate) fn clock_res_get(&self, _clock_id: ClockId) -> FileSystemResult<Timestamp> {
-        Err(ErrNo::NoSys)
+    pub(crate) fn clock_res_get(&self, clock_id: ClockId) -> FileSystemResult<Timestamp> {
+        let clock_id = u32::from(clock_id) as u8;
```
dominic-mulligan-arm, line 340: What should happen if `clock_id` is not in the range `[0..255)`?  Where does this restriction come from, anyway?

execution-engine/src/fs.rs
```diff
@@ -334,20 +334,28 @@ impl FileSystem {
...
+    /// Get clock time from platform services. Return error `NoSys` in case of failure.
     #[inline]
     pub(crate) fn clock_time_get(
         &self,
-        _clock_id: ClockId,
+        clock_id: ClockId,
         _precision: Timestamp,
     ) -> FileSystemResult<Timestamp> {
-        Err(ErrNo::NoSys)
+        let clock_id = u32::from(clock_id) as u8;
```
dominic-mulligan-arm, line 354: Same comment as above.

execution-engine/src/fs.rs
```diff
@@ -912,7 +920,7 @@ impl FileSystem {
     #[inline]
     pub(crate) fn random_get(&self, buf_len: Size) -> FileSystemResult<Vec<u8>> {
         let mut buf = vec![0; buf_len as usize];
-        if let result::Result::Success = getrandom(&mut buf) {
+        if let result::Result::Success(_) = getrandom(&mut buf) {
```
dominic-mulligan-arm, line 923: This is ugly: I'd define a `is_success()` method on the `result::Result` type and use `if getrandom(&mut but).is_success()` instead.  This is also more in line with idiomatic Rust patterns for similar types in the Rust standard library (e.g. `Result` and `Option`).

platform-services/src/lib.rs
```diff
@@ -62,10 +62,34 @@ cfg_if! {
...
+
+/// Gets resolution from the specified clock.
+///
+/// Returns:
+///     - `result::Result::Success`, embedding the clock resolution in
+///       nanoseconds, if the operation successfully completed.
+///     - `result::Result::Unavailable` if the specified clock is not available
+///       on this platform.
+///     - `result::Result::UnknownError` if a runtime error occurred.
+pub fn getclockres(clock_id: u8) -> result::Result<u64> {
```
dominic-mulligan-arm, line 81: You can mark this as `[inline]`.

platform-services/src/lib.rs
```diff
@@ -62,10 +62,34 @@ cfg_if! {
...
+
+/// Gets time from the specified clock.
+///
+/// Returns:
+///     - `result::Result::Success`, embedding the clock time in
+///       nanoseconds, if the operation successfully completed.
+///     - `result::Result::Unavailable` if the specified clock is not available
+///       on this platform.
+///     - `result::Result::UnknownError` if a runtime error occurred.
+pub fn getclocktime(clock_id: u8) -> result::Result<u64> {
```
dominic-mulligan-arm, line 93: Similarly, this can be marked `[inline]`.

platform-services/src/nitro_platform_services.rs
```diff
@@ -29,9 +29,19 @@ pub fn platform_getrandom(buffer: &mut [u8]) -> result::Result {
...
     };
     return match status {
-        nsm_io::ErrorCode::Success => result::Result::Success,
+        nsm_io::ErrorCode::Success => result::Result::Success(()),
         _ => result::Result::UnknownError,
     };
 }
 
+/// Returns the clock resolution in nanoseconds.
+/// TODO: implement it
```
dominic-mulligan-arm, line 38: Is there an issue in the issue tracker, so this doesn't get forgotten?

platform-services/src/nitro_platform_services.rs
```diff
@@ -29,9 +29,19 @@ pub fn platform_getrandom(buffer: &mut [u8]) -> result::Result {
...
+pub fn platform_getclockres(clock_id: u8) -> result::Result<u64> {
+    result::Result::Unavailable
+}
 
+/// Returns the clock time in nanoseconds.
+/// TODO: implement it
```
dominic-mulligan-arm, line 44: Same here.

platform-services/src/sgx_platform_services.rs
```diff
@@ -19,9 +19,21 @@ use sgx_trts::trts;
...
+pub fn platform_getrandom(buffer: &mut [u8]) -> result::Result<()> {
     if let Ok(_) = trts::rsgx_read_rand(buffer) {
-        return result::Result::Success;
+        return result::Result::Success(());
     }
     result::Result::UnknownError
 }
+
+/// Returns the clock resolution in nanoseconds.
+/// TODO: implement it
```
dominic-mulligan-arm, line 30: Is there an issue in the issue tracker?

platform-services/src/sgx_platform_services.rs
```diff
@@ -19,9 +19,21 @@ use sgx_trts::trts;
...
+pub fn platform_getclockres(clock_id: u8) -> result::Result<u64> {
+    result::Result::Unavailable
+}
+
+/// Returns the clock time in nanoseconds.
+/// TODO: implement it
```
dominic-mulligan-arm, line 36: Same here.

platform-services/src/std_platform_services.rs
```diff
@@ -15,14 +17,48 @@
...
     if let Ok(_) = getrandom::getrandom(buffer) {
-        return result::Result::Success;
+        return result::Result::Success(());
     }
     result::Result::UnknownError
 }
+
+/// Returns the clock resolution in nanoseconds.
+///
+/// This is for use with "freestanding-execution-engine".
```
dominic-mulligan-arm, line 38: I think this comment should be omitted: it may find use elsewhere and then the comment will be out of date.

platform-services/src/tz_platform_services.rs
```diff
@@ -19,7 +19,19 @@ use optee_utee::{Random};
...
 /// SDK to implement this.
-pub fn platform_getrandom(buffer: &mut [u8]) -> result::Result {
+pub fn platform_getrandom(buffer: &mut [u8]) -> result::Result<()> {
     Random::generate(buffer);
-    result::Result::Success
+    result::Result::Success(())
+}
+
+/// Returns the clock resolution in nanoseconds.
+/// TODO: implement it
```
dominic-mulligan-arm, line 28: Is there an issue in the issue tracker?

platform-services/src/tz_platform_services.rs
```diff
@@ -19,7 +19,19 @@ use optee_utee::{Random};
...
+pub fn platform_getclockres(clock_id: u8) -> result::Result<u64> {
+    result::Result::Unavailable
+}
+
+/// Returns the clock time in nanoseconds.
+/// TODO: implement it
```
dominic-mulligan-arm, line 34: Same comment as above.

dominic-mulligan-arm (changes requested):

dominic-mulligan-arm:
@gbryant-arm can you mark this as WIP and convert it to a draft PR until the policy aspect is implemented?

execution-engine/src/fs.rs
ShaleXIONG: I think `ClockId` is effectively  `u8` https://github.com/oasislabs/wasi-types/blob/3414ddf62499e793d72710a4866ec9312ce0a035/src/lib.rs#L34. It should be a safe conversion here.

dominic-mulligan-arm: Thanks, I'll mark this as resolved.

platform-services/src/std_platform_services.rs
```diff
@@ -15,14 +17,48 @@
...
+pub fn platform_getclockres(clock_id: u8) -> result::Result<u64> {
+    let clock_id = time::ClockId::from_raw(clock_id.into());
+    let timespec = match time::clock_getres(clock_id) {
+        Ok(t) => t,
+        Err(_) => return result::Result::Unavailable,
+    };
+
+    // Catch overflow
+    if timespec.tv_sec() == 0 {
```
egrimley-arm, line 47: Don't you expect `tv_sec` to be zero in the result from `clock_getres`?

egrimley-arm: Here's the source for `num_nanoseconds`: https://github.com/nix-rust/nix/blob/master/src/sys/time.rs#L173

Doesn't that already handle `tv_sec` for you?

platform-services/src/std_platform_services.rs
egrimley-arm: Note that the same comment appears in at least one other place.

execution-engine/src/fs.rs
ShaleXIONG: The crate is labelled with no_std. We could weaken this limitation.

platform-services/src/tz_platform_services.rs
gbryant-arm: Just added the issue to the tracker

platform-services/src/std_platform_services.rs
gbryant-arm: Agreed, should we remove it from `getrandom()` too?

dominic-mulligan-arm: Yeah, I think so.

dominic-mulligan-arm:
Incidentally, @gbryant-arm, you need to be signing all of your commits now using GPG.  I have put instructions for how to do this in the public Veracruz Slack channel.  Please enable this and edit your commit history on this pull request to sign all commits (e.g. do `git reset --soft $HASH_JUST_BEFORE_YOUR_FIRST_COMMIT` then `git commit`).

egrimley-arm:
> (e.g. do `git reset --soft $HASH_JUST_BEFORE_YOUR_FIRST_COMMIT` then `git commit`).

Or (probably) `git commit --amend`.

platform-services/src/std_platform_services.rs
gbryant-arm: Thanks for the spotting.
Actually I don't think there's a point in catching overflows here, as timespec.tv_nsec is an i64

execution-engine/src/fs.rs
gbryant-arm: @ShaleXIONG Interesting. What do you suggest? What features of `std` do we need here?

gbryant-arm:
> You've just introduced a use of `default_value`, which I carefully removed in other pull requests!
@egrimley-arm I noticed. As mentioned in the TODO I'll rebase once the next PR is merged

test-collateral/generate-policy/src/main.rs
```diff
@@ -347,6 +353,18 @@ binary.",
...
+            Arg::with_name("enable-clock")
+                .short("c")
+                .long("enable-clock")
+                .help(
+                    "Specifies whether the Veracruz trusted runtime should allow the WASM \
+binary to call clock functions (`clock_getres()`, `clock_gettime()`).",
+                )
+                .required(true)
+                .value_name("BOOLEAN")
+                .default_value(&default_enable_clock),
```
egrimley-arm, line 366: Don't forget to remove `required` and `default_value` here!

gbryant-arm: Thanks @egrimley-arm . git rebase didn't spot that

gbryant-arm:
Rebased and ready for review

test-collateral/Makefile
```diff
@@ -128,9 +128,10 @@ css-$(TEE).bin: $(RUNTIME_MANAGER_CSS_BIN)
 get_random_policy.json: pgen pgen css-$(TEE).bin client_rsa_cert.pem random-source.wasm
 	./pgen --certificate client_rsa_cert.pem --capability "output : $(READ_RIGHT), random-source.wasm : $(WRITE_RIGHT)" \
 		--binary random-source.wasm --capability "output : $(WRITE_RIGHT)" \
-       --veracruz-server-ip 127.0.0.1:3011 --proxy-attestation-server-ip 127.0.0.1:3010 --output-policy-file $@ --certificate-expiry $(CERTIFICATE_EXPIRY) \
+		--veracruz-server-ip 127.0.0.1:3011 --proxy-attestation-server-ip 127.0.0.1:3010 --output-policy-file $@ --certificate-expiry $(CERTIFICATE_EXPIRY) \
```
egrimley-arm, line 131: It's not the fault of this pull request, but this `Makefile` is very repetitive and hard to read.

execution-engine/src/fs.rs
```diff
@@ -334,20 +338,36 @@ impl FileSystem {
...
-        Err(ErrNo::NoSys)
+    pub(crate) fn clock_res_get(&self, clock_id: ClockId) -> FileSystemResult<Timestamp> {
+        if !self.enable_clock {
+            return Err(ErrNo::NoSys);
+        }
+
+        let clock_id = u32::from(clock_id) as u8;
+        match getclockres(clock_id) {
+            result::Result::Success(resolution) => Ok(Timestamp::from_nanos(resolution)),
+            _ => Err(ErrNo::NoSys),
```
egrimley-arm, line 351: Is `NoSys` the right error code here? You don't seem to be distinguishing between the clock being disabled and it failing for some other reason, which will almost certainly be the `clock_id` being invalid, I would guess, though I'm a bit confused about which platform(s) are involved here.

gbryant-arm: What error code do you suggest? I can't find any appropriate WASI error code for that.

egrimley-arm (changes requested):
I can see one problem, I think: in several places, different kinds of error are not being distinguished.  Perhaps they don't have to be distinguished, but there's also a comment in one place that claims that there will be two different error codes: `Unavailable` or `UnknownError`.

There are several ways the clock funtions could fail:

1. They are not implemented on a particular platform.
2. They have not been enabled.
3. The `clock_id` was invalid. (This is `EINVAL` on Linux.)
4. Some unexpected or unrecognised error code was returned from a lower level.

However, I would guess that it is best to return the same error code in cases 1 and 2, and for case 4 you could perhaps panic and die rather than return `UnknownError`: that could be a lot less painful for people to debug in future.

Anyone else have any thoughts on this?

execution-engine/src/fs.rs
egrimley-arm: Presumably this is the list of possibilities: https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#errno

egrimley-arm: On Linux there's `ENOSYS` for the thing not being implemented, I think, and `EINVAL` for the `clock_id` being unrecognised, and anything else is arguably an internal error and could perhaps cause an immediate exit.

gbryant-arm: We're converting to WASI. The clock function call is made from the WASM program, goes through the execution engine before reaching the the platform services in the trusted runtime.
There doesn't seem to be any WASI error code for generic or internal errors, that's a shame.

egrimley-arm:
The commit log message in the first commit (5b6ade221d7df0dc408ca919ef17a6a4c777d341) doesn't conform to the standard format, but I don't understand how GitHub generates the final log message when you do "Squash and merge".

execution-engine/src/fs.rs
egrimley-arm: What do you think about panicking rather than returning a generic or internal error for the application to deal with? I find it hard to imagine how the application could do anything useful if it gets a value that suggests that the runtime is buggy.

egrimley-arm: So this is the list of options, is it? https://docs.rs/wasi-types/0.1.6/wasi_types/enum.ErrNo.html

egrimley-arm: Or is it here? https://github.com/veracruz-project/wasi-types/blob/veracruz/src/lib.rs#L120

egrimley-arm: It looks like `NoSys` and `Inval` are the two errors we should expect to happen.

dominic-mulligan-arm: I agree that there should be separate error codes returned, here.

platform-services/src/result.rs
```diff
@@ -20,11 +20,20 @@
...
-    Success,
+    Success(T),
     /// The operation is unavailable on this platform.
     Unavailable,
     /// An unknown error occurred during the execution of the operation.
     UnknownError,
 }
+
+impl<T> Result<T> {
+    pub fn is_success(&self) -> bool {
```
dominic-mulligan-arm, line 33: Missing `rustdoc` comments on a public function, here.

platform-services/src/std_platform_services.rs
```diff
@@ -15,14 +17,36 @@
...
+    sys::time::TimeValLike,
+};
 
 /// Fills a buffer, `buffer`, with random bytes sampled from the random number
 /// source provided by the host operating system, as provided by `getrandom`.
-///
-/// This is for use with "freestanding-execution-engine".
-pub fn platform_getrandom(buffer: &mut [u8]) -> result::Result {
+pub fn platform_getrandom(buffer: &mut [u8]) -> result::Result<()> {
     if let Ok(_) = getrandom::getrandom(buffer) {
```
dominic-mulligan-arm, line 28: Why not use `is_success()` here, rather than a dummy pattern match?

dominic-mulligan-arm (commented):
Have you checked this with the `rustfmt` Git hook?

dominic-mulligan-arm:
> I can see one problem, I think: in several places, different kinds of error are not being distinguished. Perhaps they don't have to be distinguished, but there's also a comment in one place that claims that there will be two different error codes: `Unavailable` or `UnknownError`.
> 
> There are several ways the clock funtions could fail:
> 
> 1. They are not implemented on a particular platform.
> 2. They have not been enabled.
> 3. The `clock_id` was invalid. (This is `EINVAL` on Linux.)
> 4. Some unexpected or unrecognised error code was returned from a lower level.
> 
> However, I would guess that it is best to return the same error code in cases 1 and 2, and for case 4 you could perhaps panic and die rather than return `UnknownError`: that could be a lot less painful for people to debug in future.
> 
> Anyone else have any thoughts on this?

I agree that we need to distinguish between `EINVAL` and `ENOSYS` for different failure modes of the clock.  If we ever come across an unexpected return code, or an internal invariant failed, then we can panic.  Note that I think that's the only legitimate use of `panic!()` and similar partial functions/macros in the runtime: as a rule, panics should only ever be used to signal a serious bug in our code and we should be hardened against potentially malicious code trying to induce panics by feeding bad data to the runtime.

platform-services/src/std_platform_services.rs
gbryant-arm: This `getrandom()` function belongs to another crate, it doesn't return a `result::Result`

execution-engine/src/fs.rs
gbryant-arm: We should definitely return `inval` for unsupported clocks according to https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#-clock_res_getid-clockid---resulttimestamp-errno

gbryant-arm:
> > I can see one problem, I think: in several places, different kinds of error are not being distinguished. Perhaps they don't have to be distinguished, but there's also a comment in one place that claims that there will be two different error codes: `Unavailable` or `UnknownError`.
> > There are several ways the clock funtions could fail:
> > 
> > 1. They are not implemented on a particular platform.
> > 2. They have not been enabled.
> > 3. The `clock_id` was invalid. (This is `EINVAL` on Linux.)
> > 4. Some unexpected or unrecognised error code was returned from a lower level.
> > 
> > However, I would guess that it is best to return the same error code in cases 1 and 2, and for case 4 you could perhaps panic and die rather than return `UnknownError`: that could be a lot less painful for people to debug in future.
> > Anyone else have any thoughts on this?
> 
> I agree that we need to distinguish between `EINVAL` and `ENOSYS` for different failure modes of the clock. If we ever come across an unexpected return code, or an internal invariant failed, then we can panic. Note that I think that's the only legitimate use of `panic!()` and similar partial functions/macros in the runtime: as a rule, panics should only ever be used to signal a serious bug in our code and we should be hardened against potentially malicious code trying to induce panics by feeding bad data to the runtime.

According to the POSIX doc, `clock_gettime()` and `clock_getres()` will only return two possible errors:
* EFAULT if `tp` (timespec structure containing the time) points outside the accessible address space
  + unlikely to happen, as `tp` doesn't exist in the WASI API (the timespec is returned instead of being passed by reference); the only place where it could happen is in the nix implementation
* EINVAL if the clk_id specified is not supported on this system
  + might never happen depending on the implementation; glibc for example selects a default clock when the clock id doesn't match any existing clock
  
 We could simply return:
 * at the platform service level: `result::Result::Success` or `result::Result::UnknownError`
 * at the WASI level: `ErrNo::Access` (clock flag not set in the policy) or `ErrNo::Inval` (generic error)
 
Alternatively we could propagate `errno` (the one that the low-level function returns) back to WASI to avoid having to define hundreds of error codes at the platform service level (`platform-services/src/result.rs`)

test-collateral/Makefile
gbryant-arm: Could we factorise those long calls into a call to a function? The only variable seems to be the capability and the binary name.
Example:
```
define call_pgen
    ./pgen --certificate client_rsa_cert.pem --capability "$(1)" --binary "$(2)" <...>
endef

get_random_policy.json: pgen pgen css-$(TEE).bin client_rsa_cert.pem random-source.wasm
    $(call call_pgen, "output : $(READ_RIGHT), random-source.wasm : $(WRITE_RIGHT)", "random-source.wasm")
```

Incidentally, `--enclave-debug-mode` is not always set, is that on purpose?

egrimley-arm: I'm going to mark this as resolved. It would be nice to simplify this `Makefile`, but that would be a different issue.

dominic-mulligan-arm: @egrimley-arm, I agree that the Makefile needs simplifying, but I think that is outside of the scope of this changeset and should be handled in another PR.

egrimley-arm (approved):

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
+1+1 = +1 reached, merging.

PR #216, opened 2021-09-09
generate-policy: Bug fixes in parsing of command line.

egrimley-arm:
- The "default_value" feature is broken and must be avoided.
- It appears that "short" options must use only one letter.

### DIFF
--- a/test-collateral/generate-policy/src/main.rs
+++ b/test-collateral/generate-policy/src/main.rs
@@ -212,8 +212,6 @@ fn check_capability(capabilities: &[Vec<String>]) {
 /// of them.  If required options are not present, or if any options are
 /// malformed, this will abort the program.
 fn parse_command_line() -> Arguments {
-    let default_debug = DEFAULT_DEBUG_STATUS.to_string();
-
     let matches = App::new(APPLICATION_NAME)
         .version(VERSION)
         .author(AUTHORS)
@@ -281,8 +279,6 @@ fn parse_command_line() -> Arguments {
                 .long("output-policy-file")
                 .value_name("FILE")
                 .help("Filename of the generated policy file.")
-                .default_value(DEFAULT_OUTPUT_FILENAME)
-                .required(true),
         )
         .arg(
             Arg::with_name("certificate-expiry")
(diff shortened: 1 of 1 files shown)


test-collateral/generate-policy/src/main.rs
```diff
@@ -327,28 +321,23 @@ information to be produced by the executing WASM binary.",
...
             Arg::with_name("stdin")
-                .short("si")
                 .long("stdin")
                 .value_name("STDIN")
                 .help("The configuration of the standard input in the form 'path:rights'")
                 .required(false),
         )
         .arg(
             Arg::with_name("stdout")
-                .short("so")
```
dominic-mulligan-arm, line 343: Can we retain a short flag for each of these command line arguments?  e.g. 'e' for `stderr`, 'I' for `stdin`, and 'o' for `stdout`?

dominic-mulligan-arm (changes requested):

test-collateral/generate-policy/src/main.rs
egrimley-arm: Perhaps, though `-e` and `-o` are already in use. The existing use could be changed, but how careful would I have to be about changing the meaning of a flag? I can't be sure that all existing uses are in the repo, or that I'd find them all even if they were.

What I could do is use `-E`, `-I`, `-O` for `stderr`, `stdin`, `stdout`.

dominic-mulligan-arm: All existing uses of this are almost certainly in this repository, and any problems will be identified by CI, as `generate_policy` is used to generate the policy files needed for integration and unit tests (see e.g. the `test-collateral` Makefile, which invokes `generate_policy`, renamed to `pgen`).  I think there, though, we use the long flags, rather than short.

`E`, `I`, and `O` seem reasonable short names.

egrimley-arm: I've done that. (Changing `-e` to `-x`, and `-x` to something else, etc., might have been possible, but seems less attractive!)

dominic-mulligan-arm (approved):

dominic-mulligan-arm (approved):

gbryant-arm (approved):
LGTM

PR #218, opened 2021-09-16
Force rustfmt to use 2018 Rust edition in pre-commit hook

dominic-mulligan-arm:
There is a bug in `rustfmt` currently that ignores the edition set in `Cargo.toml` files and defaults to 2015 edition. This means it errors on any staged file containing the `async` keyword, and any other change introduced in rust 2018.
The solution is to force the Rust edition to use when calling `rustfmt`.

### DIFF
--- a/githooks/pre-commit
+++ b/githooks/pre-commit
@@ -20,7 +20,7 @@ HAS_ISSUES=0
 echo "Checking code formatting of files staged for commit."
 
 for file in $(git diff --name-only --staged \*.rs); do
-    RUSTFMT="$(rustfmt --check --skip-children --unstable-features $file)"
+    RUSTFMT="$(rustfmt --edition=2018 --check --skip-children --unstable-features $file)"
     if [ "$RUSTFMT" != "" ]; then
         printf "[ERROR]: $file\n"
         HAS_ISSUES=1


egrimley-arm:
Could you add a comment (in `githooks/pre-commit`) with a link to the upstream bug?

I tried looking for it myself, but all I could find was https://github.com/rust-lang/rustfmt/issues/4645, which was closed with someone saying the reporter was using the wrong version of `rustfmt`.

Could you also point to a file in the repo (I mean in the discussion here rather than in the code) that needs the `--edition=2018` so I can check whether my `rustfmt` is affected by the bug?

egrimley-arm:
To answer my own question, an example of something that needs `--edition=2018` is: `async fn foo() {}`

It seems that `rustfmt` is supposed to ignore the `Cargo.toml`, while `cargo fmt` is incapable of looking at a single source file, so invoking `rustfmt` with `--edition=2018` seems to be the right thing to do in our case.

ShaleXIONG (approved):

gbryant-arm (approved):

dominic-mulligan-arm:
+1+1=+1 reached, merging.

dominic-mulligan-arm:
Oops, seems there was some issue with the signing of commits.  I've resigned and pushed again.  @ShaleXIONG, @gbryant-arm can you take another look and approve, ready for merging?

gbryant-arm (approved):

dominic-mulligan-arm:
+1 reached, merging.

PR #221, opened 2021-09-17
WIP: Move some stuff from fs.rs into common.rs.

egrimley-arm:
In preparation for implementing an strace-like function,
move some stuff that does not appear to be file-system-related
from fs.rs into common.rs. This includes the configuration
parameter enable_clock, which is now part of a struct Options
that can easily be extended. The main system calls affected are:

* clock_res_get
* clock_time_get
* random_get

The behaviour of fd_filestat_set_times and path_filestat_set_times
is slightly altered: they will no longer fail if the clock is not
enabled.


gbryant-arm (changes requested):
Can you run rustfmt please? LGTM otherwise

egrimley-arm:
> Can you run rustfmt please? LGTM otherwise

Done. I think I was expecting a `git commit --amend` or a `git rebase` to rerun the checks, but it didn't. (I actually spent 10-15 minutes trying to get `githooks/pre-commit` to work in those cases, but I couldn't find a way to do it.)

gbryant-arm (approved):

PR #222, opened 2021-09-17
Implement "strace" output functions.

egrimley-arm:
Output is written to stderr so that nothing is lost if the process
exits unexpectedly. Currently not all types of argument are printed
in a useful way.

A lot of the comments below are no longer relevant!

### DIFF
--- a/execution-engine/src/wasi/strace.rs
+++ b/execution-engine/src/wasi/strace.rs
@@ -9,30 +9,99 @@
 //! See the `LICENSE_MIT.markdown` file in the Veracruz root directory for
 //! information on licensing and copyright.
 
-#![allow(unused_variables)] // while most of the functions are unimplemented
-
 use super::common::MemoryHandler;
 use crate::fs::FileSystemResult;
 use std::fmt;
 
-pub struct Strace {}
+enum TraceState {
+    Func,
+    Args,
+    Result,
+}
+
+pub struct Strace {
+    enabled: bool,
+    state: TraceState,
+}
+
+// Convert a vector of bytes into a printable ASCII string.
+// The string may contain readable text, which is useful to display,
+// but it may also contain binary data, so we cannot decode it as UTF-8.
+fn strace_string(bytes: &[u8], max: usize) -> String {
+    let mut res = String::from("\"");
+    let n = if bytes.len() > max { max } else { bytes.len() };
+    for i in 0..n {
+        if 0x20 <= bytes[i] && bytes[i] < 0x7f {
+            if bytes[i] == b'\\' || bytes[i] == b'"' {
+                res.push_str("\\");
+            }
+            res.push_str(&String::from_utf8_lossy(&bytes[i..i + 1].to_vec()))
+        } else if bytes[i] == 9 {
+            res.push_str("\\t")
+        } else if bytes[i] == 10 {
+            res.push_str("\\n")
+        } else if i + 1 < n && b'0' <= bytes[i + 1] && bytes[i + 1] <= b'9' {
+            // The following character is a digit, so use three octal digits.
+            res.push_str(&format!("\\{:03o}", bytes[i]))
+        } else {
+            res.push_str(&format!("\\{:o}", bytes[i]))
+        }
+    }
+    res.push_str("\"");
+    if bytes.len() > max {
+        res.push_str("...")
+    }
+    res
+}
 
 impl Strace {
     pub fn func(enabled: bool, name: &str) -> Self {
-        // NOT YET IMPLEMENTED
-        Strace {}
+        if enabled {
+            eprint!("{}(", name)
+        };
+        Strace {
+            enabled,
+            state: TraceState::Func,
+        }
+    }
+
+    fn arg(&mut self) -> bool {
+        if !self.enabled {
+            return true;
+        }
+        match self.state {
+            TraceState::Func => self.state = TraceState::Args,
+            TraceState::Args => {
+                eprint!(", ");
+            }
+            TraceState::Result => eprint!("\nUnexpected strace arg: "),
+        }
+        false
     }
 
     pub fn arg_buffer<T: MemoryHandler>(&mut self, mem: &mut T, adr: u32, len: u32) {
-        // NOT YET IMPLEMENTED
+        if self.arg() {
+            return;
+        }
+        let mut bytes = vec![0u8; len as usize];
+        match mem.read_buffer(adr, &mut bytes) {
+            Ok(()) => eprint!("{}", strace_string(&bytes, 32)),
+            Err(_) => eprint!("BAD_MEM_REF"),
+        }
     }
 
     pub fn arg_dec<T: fmt::Display>(&mut self, n: T) {
-        // NOT YET IMPLEMENTED
+        if self.arg() {
+            return;
+        }
+        eprint!("{}", n)
     }
 
     pub fn arg_dots(&mut self) {
-        // NOT YET IMPLEMENTED
+        if self.arg() {
+            return;
+        }
+        eprint!("...")
     }
 
     pub fn arg_dirents<T: MemoryHandler>(
@@ -42,23 +111,42 @@ impl Strace {
         _buf_len: u32,
         _result_ptr: u32,
     ) {
+        if self.arg() {
+            return;
+        }
         // NOT YET IMPLEMENTED
+        eprint!("DIRENTS")
     }
 
     pub fn arg_events<T: MemoryHandler>(&mut self, _mem: &mut T, _events: u32, _size: u32) {
+        if self.arg() {
+            return;
+        }
         // NOT YET IMPLEMENTED
+        eprint!("EVENTS")
     }
 
     pub fn arg_fdstat<T: MemoryHandler>(&mut self, _mem: &mut T, _adr: u32) {
+        if self.arg() {
+            return;
+        }
         // NOT YET IMPLEMENTED
+        eprint!("FDSTAT")
     }
 
     pub fn arg_filestat<T: MemoryHandler>(&mut self, _mem: &mut T, _adr: u32) {
+        if self.arg() {
+            return;
+        }
         // NOT YET IMPLEMENTED
+        eprint!("FILESTAT")
     }
 
     pub fn arg_hex<T: fmt::LowerHex>(&mut self, n: T) {
-        // NOT YET IMPLEMENTED
+        if self.arg() {
+            return;
+        }
+        eprint!("0x{:x}", n)
     }
 
     pub fn arg_iovec<T: MemoryHandler>(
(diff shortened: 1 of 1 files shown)


gbryant-arm (changes requested):
It's hard to tell this PR apart from #221.
I suggest we first merge #221 before you rebase this PR on top of it.

egrimley-arm:
This is currently a single commit on top of #221, so you can just look at that commit even before it's merged:

https://github.com/veracruz-project/veracruz/pull/222/commits/8c16a3efbd9d011d4272284a3d5e487e6d2e0e49

execution-engine/src/wasi/common.rs
```diff
@@ -371,6 +371,21 @@ pub struct WasiWrapper {
...
+    /// Whether strace is enabled.
+    pub enable_strace: bool,
+}
+
+macro_rules! strace {
+    ($self: expr, $name: expr, $block: block) => {{
+        let result = (|| $block)();
+        if $self.enable_strace {
+            match result {
+                Ok(()) => eprintln!("{}(...) = Success", $name),
```
gbryant-arm, line 383: How about calling `println!` here, or using the `log` crate? Should we consider system call errors as Veracruz errors?

execution-engine/src/wasi/common.rs
```diff
@@ -395,6 +410,7 @@ impl WasiWrapper {
             principal,
             exit_code: None,
             enable_clock: false,
+            enable_strace: false,
```
gbryant-arm, line 413: Consistency issue here? We should pick between `strace` and `enable_strace`

gbryant-arm (commented):
Thanks for opening this PR @egrimley-arm.
Besides the few comments I left I think having a macro here makes a lot of sense

egrimley-arm:
I've added some fancy macros so that arguments can be nicely formated.

The output looks like this:
```
path_open(3, 0x1, "input.txt", 0x0, 0xf87febe, 0xf87febe, 0x0) = Access
```

The source looks like this:

((I wanted to create a link to it but I can't make it work. So search for `strace!` in `common.rs`.)

execution-engine/src/wasi/common.rs
egrimley-arm: System call errors are fairly unusual but they are expected in some circumstances. I'm using `eprint` because Linux's `strace` writes to `stderr`: not an amazingly good argument, but I have to decide somehow.

execution-engine/src/wasi/common.rs
```diff
@@ -373,6 +395,189 @@ pub struct WasiWrapper {
...
+        return;
+    }
+    if let Ok(len) = memory_ref.read_u32(address) {
+        let max = if len < 32 { len } else { 32 } as usize;
+        // This inefficiently copies everything, but it's only used for tracing.
+        if let Ok(bufs) = (|| {
+            let iovecs = memory_ref.unpack_array::<IoVec>(base, count)?;
+            memory_ref.read_iovec_scattered_in_reverse(&iovecs)
+        })() {
+            let mut buf: Vec<u8> = Vec::new();
```
dominic-mulligan-arm, line 449: Is this doing the same job as `.iter().flatten()`?

execution-engine/src/wasi/common.rs
```diff
@@ -373,6 +395,189 @@ pub struct WasiWrapper {
...
+            for b in &bufs {
+                buf.extend(b)
+            }
+            eprint!("{}", strace_string(&buf, max))
+        } else {
+            eprint!("BAD_IOVEC")
+        }
+    } else {
+        eprint!("BAD_IOVEC_LEN") // This will probably never happen.
```
dominic-mulligan-arm, line 458: Is it should never happen then this should be a runtime trap when it does happen.

execution-engine/src/wasi/common.rs
```diff
@@ -373,6 +395,189 @@ pub struct WasiWrapper {
...
+    }
+}
+
+macro_rules! strace_arg {
```
dominic-mulligan-arm, line 462: Can we move all of this `strace`-related material into a dedicated module and document what it is doing?  Also, is there no better way of doing this than using the macro system?  The disadvantages of macros are longer compile times and impaired support from code editing/IDE tools --- the latter's especially important here as it seems every syscall-handler is being wrapper in a giant macro.

execution-engine/src/wasi/common.rs
```diff
@@ -373,6 +395,189 @@ pub struct WasiWrapper {
...
+    ($success: expr, ($mem: expr, $adr: expr, $len: expr, buffer)) => {
+        match $mem.read_buffer($adr, $len) {
+            Ok(x) => eprint!("{}", strace_string(&x, 32)),
+            Err(_) => eprint!("BAD_MEM_REF"),
+        }
+    };
+    ($success: expr, ($n: expr, dec)) => {
+        eprint!("{}", $n)
```
dominic-mulligan-arm, line 470: I'm wary of us having `eprintln` and friends directly in the runtime itself.  I think we've taken pains to avoid that until now whenever possible.  Rather than printing directly to `stderr` is there no way of printing to a dedicated `strace` stream which can be read after the program has finished executing, like `stdout` and `stderr` are, with access managed by the policy?

execution-engine/src/wasi/common.rs
```diff
@@ -373,6 +395,189 @@ pub struct WasiWrapper {
...
+    };
+    ($success: expr, ($mem: expr, $events: expr, $size: expr, events)) => {
+        eprint!("EVENTS") //xx
```
dominic-mulligan-arm, line 473: What are these comments?

execution-engine/src/wasi/common.rs
```diff
@@ -373,6 +395,189 @@ pub struct WasiWrapper {
...
+        let tracing = $self.enable_strace;
+        // Print function name first in case execution causes crash.
+        if tracing {
+            eprint!("{}(", $name);
+        }
+        // Execute function.
+        let result = (|| $block)();
+        // Print arguments last as they may include output arguments.
+        if tracing {
+            let _success = match result {
```
dominic-mulligan-arm, line 568: I think this can be `result.is_ok()`.

execution-engine/src/wasi/common.rs
```diff
@@ -436,116 +642,144 @@ impl WasiWrapper {
...
     ) -> FileSystemResult<()> {
-        let environc = self.program_arguments.len() as u32;
-        let environ_buf_size = self
-            .program_arguments
-            .iter()
-            .fold(0, |acc, arg| acc + format!("{}\0", arg).as_bytes().len());
-
-        memory_ref.write_u32(address_for_counts, environc)?;
-        memory_ref.write_u32(address_for_buffer_size, environ_buf_size as u32)
+        strace!(self, "args_sizes_get", [((), dots)], {
```
dominic-mulligan-arm, line 663: Why must everything be wrapped in this macro?

dominic-mulligan-arm (changes requested):

execution-engine/src/wasi/common.rs
```diff
@@ -294,6 +295,27 @@ pub trait MemoryHandler {
         Ok(rst)
     }
 
+    /// The default implementation for reading a u16 from `address`.
+    fn read_u16(&mut self, address: u32) -> FileSystemResult<u16> {
+        let vec = self.read_buffer(address, 2)?;
+        let bytes: [u8; 2] = vec.as_slice().try_into().unwrap();
```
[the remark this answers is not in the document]
egrimley-arm: I've removed `unwrap` from here.

execution-engine/src/wasi/common.rs
```diff
@@ -373,6 +395,189 @@ pub struct WasiWrapper {
     exit_code: Option<u32>,
     /// Whether clock functions (`clock_getres()`, `clock_gettime()`) should be enabled.
     pub(crate) enable_clock: bool,
+    /// Whether strace is enabled.
+    pub(crate) enable_strace: bool,
+}
+
+// Convert a vector of bytes into a printable string.
+// TODO: Perhaps we should use a more generous definition of "printable".
```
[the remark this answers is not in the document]
egrimley-arm: I've noted it handles only ASCII.

execution-engine/src/wasi/common.rs
```diff
@@ -373,6 +395,189 @@ pub struct WasiWrapper {
...
+            res.push_str("\\t")
+        } else if bytes[i] == 10 {
+            res.push_str("\\n")
+        } else if i + 1 < n && b'0' <= bytes[i + 1] && bytes[i + 1] <= b'9' {
+            // The following character is a digit, so use three octal digits.
+            res.push_str(&format!("\\{:03o}", bytes[i]))
+        } else {
+            res.push_str(&format!("\\{:o}", bytes[i]))
+        }
+    }
```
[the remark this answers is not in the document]
egrimley-arm: I've improved the comment:
```
// Convert a vector of bytes into a printable ASCII string.
// The string may contain readable text, which is useful to display,
// but it may also contain binary data, so we cannot decode it as UTF-8.
```

execution-engine/src/wasi/common.rs
```diff
@@ -373,6 +395,189 @@ pub struct WasiWrapper {
     exit_code: Option<u32>,
     /// Whether clock functions (`clock_getres()`, `clock_gettime()`) should be enabled.
     pub(crate) enable_clock: bool,
+    /// Whether strace is enabled.
+    pub(crate) enable_strace: bool,
+}
+
+// Convert a vector of bytes into a printable string.
+// TODO: Perhaps we should use a more generous definition of "printable".
+fn strace_string(bytes: &Vec<u8>, max: usize) -> String {
```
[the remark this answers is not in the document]
egrimley-arm: Done.

execution-engine/src/wasi/common.rs
```diff
@@ -373,6 +395,189 @@ pub struct WasiWrapper {
...
+    address: u32,
+    base: u32,
+    count: u32,
+) {
+    if !success {
+        eprint!("_");
+        return;
+    }
+    if let Ok(len) = memory_ref.read_u32(address) {
+        let max = if len < 32 { len } else { 32 } as usize;
```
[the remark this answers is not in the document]
egrimley-arm: In fact the logic here was wrong and I've removed it.

execution-engine/src/wasi/common.rs
egrimley-arm: This code has changed now for other reasons.

execution-engine/src/wasi/common.rs
egrimley-arm: I think whether it can happen or not may depend on undocumented behaviour of other parts of the runtime.

If you give a Linux system call a bad pointer then the system call returns `EFAULT`: the kernel does not panic, and the kernel does not even kill the process. If you give a Veracruz "system call" a bad pointer then probably the runtime will die before it gets to this place. But if it does get this far, then I think it should be allowed to continue. I don't think the `strace` facility should introduce additional failure modes.

execution-engine/src/wasi/common.rs
egrimley-arm: I don't see why it shouldn't be moved into a separate module. Macros certainly make error messages more confusing. I'll try rewriting part of it to see what it looks like without macros.

execution-engine/src/wasi/common.rs
egrimley-arm: But what if the runtime panics, or the program never finishes? It would be nice to get strace output when that happens.

execution-engine/src/wasi/common.rs
egrimley-arm: I write `xx` for "this should be improved at some point". I picked that up from Alan Donovan, when he was at Arm, a long time ago. It's quick to type, but I could put something more comprehensible for the uninitiated instead.

execution-engine/src/wasi/common.rs
egrimley-arm: Done.

execution-engine/src/wasi/strace.rs
```diff
@@ -0,0 +1,274 @@
...
+        &mut self,
+        _mem: &mut T,
+        _buf_ptr: u32,
+        _buf_len: u32,
+        _result_ptr: u32,
+    ) {
+        if self.arg() {
+            return;
+        }
+        eprint!("DIRENTS") //xx
```
egrimley-arm, line 117: What should I put instead of `xx`?

execution-engine/src/wasi/common.rs
dominic-mulligan-arm: If there's a desire to improve something then I think that should be recorded in the issue tracker.

execution-engine/src/wasi/common.rs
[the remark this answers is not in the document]
dominic-mulligan-arm: I see, that makes sense.

execution-engine/src/wasi/common.rs
dominic-mulligan-arm: I see.  Maybe then the `eprint!` message needs expanding to indicate that this is potentially a Veracruz runtime bug and should be investigated?

execution-engine/src/wasi/common.rs
dominic-mulligan-arm: Great, thanks.

execution-engine/src/wasi/common.rs
dominic-mulligan-arm: Hmm, this is a good point that I didn't consider.  I think, without introducing some complex scaffolding, that you're right that `eprintln!` is the best that can be done for tracing non-terminating or crashing programs.

execution-engine/src/wasi/strace.rs
dominic-mulligan-arm: I'll mark this as resolved, as this thread of conversation is a duplicate of another thread above (re: using the issue tracker).

execution-engine/src/wasi/common.rs
egrimley-arm: I've created https://github.com/veracruz-project/veracruz/issues/303, though it refers to a file that hasn't yet been merged.

execution-engine/src/wasi/common.rs
egrimley-arm: The message is part of the `strace` output so I can't make it too long. I could change it to `BAD_IOVEC_LEN_POSSIBLE_BUG`?

egrimley-arm:
I've separated the plumbing from the output and put the plumbing in https://github.com/veracruz-project/veracruz/pull/308.

egrimley-arm:
The history of comments on this PR is just too confusing. I'll make a new one:
https://github.com/veracruz-project/veracruz/pull/319

PR #224, opened 2021-09-22
Add http = "=0.2.4" to four Cargo.toml files to avoid E0658.

egrimley-arm:
Tests were failing with E0658: "if" and "while" not allowed in const fn.


dreemkiller:
Does `veracruz-server-test/Cargo.toml` not also need this change?

egrimley-arm:
> Does `veracruz-server-test/Cargo.toml` not also need this change?

Yes, probably. I tried to find the right crates by grepping an old log file, but I seem to have missed that one somehow. In my defence, my Linux box was being moved at the time so I was forced to use some ghastly Windows stuff for the grepping.

dreemkiller (approved):

PR #226, opened 2021-09-23
PAS: Removed dependency on policy.json in the Proxy Attestation Server CLI

geky:
This was a lingering limitation from when the Proxy Attestation Server required more info from the policy file.

Now, in `vc-pas`, the primary argument is what address to bind to, and no policy is needed in the Proxy Attestation Server.

It's almost possible to let users omit this argument, and default to the standard Proxy Attestation Port for all addresses, except, well, we don't have a standard port for the Proxy Attestation Server. Choosing two arbitrary ports as standard ports for Veracruz's Proxy Attestation Server and Server Server would be nice for quality-of-life things like this, but I've omitted it for now.

### DIFF
--- a/proxy-attestation-server/src/server.rs
+++ b/proxy-attestation-server/src/server.rs
@@ -18,6 +18,7 @@ use crate::attestation::sgx;
 use crate::attestation::nitro;
 
 use lazy_static::lazy_static;
+use std::net::ToSocketAddrs;
 use std::sync::atomic::{AtomicBool, Ordering};
 
 lazy_static! {
@@ -195,8 +196,9 @@ async fn nitro_router(nitro_request: web::Path<String>, input_data: String) -> P
     Err(ProxyAttestationServerError::UnimplementedRequestError)
 }
 
-pub fn server<P1, P2>(url: String, ca_cert_path: P1, ca_key_path: P2, debug: bool) -> Result<Server, String>
+pub fn server<U, P1, P2>(url: U, ca_cert_path: P1, ca_key_path: P2, debug: bool) -> Result<Server, String>
 where
+    U: ToSocketAddrs,
     P1: AsRef<path::Path>,
     P2: AsRef<path::Path>
 {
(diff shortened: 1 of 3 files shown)


proxy-attestation-server/src/server.rs
```diff
@@ -195,8 +196,9 @@ async fn nitro_router(nitro_request: web::Path<String>, input_data: String) -> P
     Err(ProxyAttestationServerError::UnimplementedRequestError)
 }
 
-pub fn server<P1, P2>(url: String, ca_cert_path: P1, ca_key_path: P2, debug: bool) -> Result<Server, String>
+pub fn server<U, P1, P2>(url: U, ca_cert_path: P1, ca_key_path: P2, debug: bool) -> Result<Server, String>
```
dreemkiller, line 199: Is there something that requires `U`, `P1`, and `P1` instead of using the types directly? Two levels of indirection to find out what types the parameters are seems a bit much if it can be avoided.

dominic-mulligan-arm (dismissed):

dominic-mulligan-arm:
LGTM, though I'll wait for @dreemkiller's approval before this can be merged.

proxy-attestation-server/src/server.rs
geky: Strictly no, these could be changed to a bunch of `&str`s. But that would move more work into the user to convert down to `&str` when HttpServer::bind/File::open are going to take U:ToSocketAddrs/P:AsRef<Path> anyways. It would also be a little bit less type safe, using `&str` for everything instead of specific traits. 

As for the indirection, that's sort of on Rust's syntax. The traits could be inlined in this case (not always), but the syntax of traits require at least one indirection without changing the behavior of the API:

``` rust
pub fn server<
    U: ToSocketAddrs,
    P1: AsRef<path::Path>,
    P2: AsRef<path::Path>
>(
    url: U,
    ca_cert_path: P1,
    ca_key_path: P2,
    debug: bool
) -> Result<Server, String> {
    blablabla
```

alexandref75:
Can't comment on the code (not a rust programmer) but the functionality LGTM.

proxy-attestation-server/src/server.rs
alexandref75: The functionality LGTM. Cannot check the code.

dominic-mulligan-arm:
@dreemkiller are you happy with @geky's explanation above?

PR #227, opened 2021-09-23
CLI: Added CLI_QUICKSTART.markdown, a quick overview of Veracruz's CLIs without needing to manually build all artifacts

geky:
Based on feedback from @alexandref75:

- Added Shamir's secret sharing demo to the test-collateral, this does a few things including:
  1. Tests that Shamir's secret sharing demo still compiles.
  2. Makes the artifacts necessary for the demo (shamir-secret-sharing-policy.json, shamir-secret-sharing.wasm, share-1,2,3.dat) available after building Veracruz. This makes it easy to test the command-line interfaces without needing to manually build all artifacts.
  3. Makes it so it's possible to write a quickstart that jumps into Veracruz without doing a few laps explaining how to build everything.

- Added a generated proxy-attestation-server.db to the test-collateral, this needs to go somewhere for the quickstart. Note that I haven't changed veracruz-test/veracruz-server-test to use this as the comment in their build.rs's indicates the proxy-attestation-server.db should be pristine, and build.rs will be invoked at different times than test-collateral's Makefile rule.
  
- Added CLI_QUICKSTART.markdown, a quick overview of Veracruz's CLIs without needing to manually build all artifacts.

- Renamed GETTING_STARTED_CLI.markdown -> CLI_INSTRUCTIONS.markdown to be consistent with other guides in the repo and hopefully be more discoverable.

- Removed run_cli_test.sh rather than updating it, this was a pending TODO item from a previous discussion.

I still think a CLI_INSTRUCTIONS.markdown, with instructions for each artifact, is valuable. Especially for new users who may not be familiar with Rust, SQL, or openssl commands. But this offers a fast-path for experienced developers.

---

Note: This is currently dependent #226 due to the change to the vc-pas command line parameters. This is easy to remove if #226 is closed.


ShaleXIONG (dismissed):

alexandref75:
I would suggest add an entry to the main README pointing to CLI_QUICKSTART.markdown so it is easier to find.

geky:
Updated with links in the readme, @dominic-mulligan-arm feel free to update the wording if you'd like, you should have push access to this branch.

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
@geky, the README looks fine to me.

dominic-mulligan-arm:
+1 reached, merging.

PR #228, opened 2021-09-27
WIP: Add command-line arguments and environment variables for Wasm.

egrimley-arm:
Just a proof of concept as probably we do not need this. To test:

cd sdk/rust-examples/argvenv/
cargo build --target=wasm32-wasi
cp target/wasm32-wasi/debug/argvenv.wasm ../../freestanding-execution-engine/
cd ../../freestanding-execution-engine
cargo run -- --program argvenv.wasm -o true \
--argv "a b" --argv "c" --env qwe=123 --env asd=456


egrimley-arm:
Since `environment_variables` and `program_arguments` were already part of the VFS, I've left them there, to make this patch minimal, but perhaps they should be part of the new `Options` structure instead.

execution-engine/src/wasi/wasmi.rs
```diff
@@ -1226,6 +1226,8 @@ impl ExecutionEngine for WASMIRuntimeState {
         file_name: &str,
         options: Options,
     ) -> Result<u32, FatalEngineError> {
+        self.vfs.environment_variables = options.environment_variables;
+        self.vfs.program_arguments = options.program_arguments;
```
gbryant-arm, line 1230: Could we move those two variables to somewhere else? Do they have to be in the VFS?

sdk/freestanding-execution-engine/src/main.rs
```diff
@@ -235,13 +253,29 @@ fn parse_command_line() -> Result<CommandLineOptions, Box<dyn Error>> {
...
+
     Ok(CommandLineOptions {
         data_sources,
         binary,
         execution_strategy,
         dump_stdout,
         dump_stderr,
         enable_clock,
+        environment_variables,
+        program_arguments,
```
gbryant-arm, line 278: How are the arguments and environment variables passed to the program? `main()` doesn't seem to be using `environment_variables` and `program_arguments`.

gbryant-arm (commented):
IMO if we're to implement this feature, it should be available in the freestanding execution engine only for now, to allow users to quickly try out programs requiring arguments. Adding it to the other execution engines would entail more complexity: more fields in the policy file, more data to provision.

egrimley-arm:
> IMO if we're to implement this feature, it should be available in the freestanding execution engine only for now

Yes, I agree with that.

execution-engine/src/wasi/wasmi.rs
egrimley-arm: I've already moved them, in a later commit. Perhaps I should rebase and squash at this point.

egrimley-arm: I've squashed and rebased. (For some reason GPG has stopped working, though.)

sdk/freestanding-execution-engine/src/main.rs
egrimley-arm: Is that clearer now with the single commit?

gbryant-arm: Oh I hadn't noticed the syscalls were already implemented, `environ_get()` and `args_get()`

egrimley-arm: Yes, this patch is small because a lot of the stuff was there already. Apart from `sdk/freestanding-execution-engine` it's 11 lines!

PR #167, opened 2021-07-06
Update execution engine to use Wasmtime 0.26

enjhnsn2:
Depends on https://github.com/veracruz-project/wasmtime/pull/2

This pull request makes some very minor updates to the Veracruz execution engine to be able to use Wasmtime 0.26.

PR #235, opened 2021-10-07
Update execution engine to use Wasmtime 0.26 (duplicate)

gbryant-arm:
Duplicate of #167 for the purpose of restarting the CI


gbryant-arm:
> It seems to make the crypto program I'm currently experimenting with run about 15% faster. Is that what you'd expect?

Great! I measured a 10% performance boost for object detection and video decoding

PR #238, opened 2021-10-15
Removed stray mention of nitro-root-enclave from Makefile

dominic-mulligan-arm:
This was causing a compile issue for the `nitro-cli` target.


dominic-mulligan-arm:
@mathias-arm can you approve this?  I noticed that the other Nitro-related PR does something similar to this, but comments out the build invocation and misses the other mention of the root enclaves.  If this gets merged I will rebase the other changes on top.

PR #242, opened 2021-10-15
Separate the lifetime of the Veracruz Server and enclave instances

geky:
The goal of this PR is to make it possible to setup and teardown policy-bound enclaves independently of the lifetime of the `vc-server` process.

So now you can launch a server without a policy (note you need to specify an address to run on):

``` bash
$ vc-server :3017 &
Veracruz Server running on 0.0.0.0:3017
```

And create a new enclave later with `vc-client`:

``` bash
$ vc-client :3017 --policy example/example-policy.json --setup
Loaded policy example/example-policy.json
  (with hash) 17115447cda3d9e8d70d7c78abf0b34f1446cc1acc00b9308f44e60085f52c01
Connecting to 0.0.0.0:3017
Setting up enclave
```

There are also some new quality-of-life features for managing enclaves:

- `--query-list` gives you a list of running enclaves and their uptime:

  ``` bash
  $ vc-client :3017 --query-list
  Connecting to 0.0.0.0:3017
  Querying list of enclaves
     id policy_hash                                                       uptime
      1 17115447cda3d9e8d70d7c78abf0b34f1446cc1acc00b9308f44e60085f52c01  0d 0h 4m 12s
  ```
  
- `--query-policy` gives you the running enclave's policy:

  ``` bash
  $ vc-client :3017 --id=1 --query-policy=-
  Connecting to 0.0.0.0:3017
  Querying policy into <stdout>
  {
      "ciphersuite": "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256",
      "debug": false,
      "enable_clock": false,
  ...
  ```
  
You can manually teardown an enclave with `--teardown`:

``` bash
$ vc-client :3017 --id=1 --teardown
Connecting to 0.0.0.0:3017
Tearing down enclave
```

Or let an enclave tear itself down (though it really asks the server to tear itself down) after all client with read permissions have signaled they are done.

This is accomplished by the SignalDone request into the runtime, which is really just a rename and slight tweaking of the previous RequestShutdown command. Renamed to hopefully make it more clear that all clients with read permission must signal done as there was a bit of confusion around this earlier.

The full flow is documented in the updated CLI_INSTRUCTIONS.markdown:
https://github.com/veracruz-project/veracruz/compare/main...geky:admin-separation?expand=1&short_path=971f80b#diff-971f80b10a659d96ff77661e29b7539fd55bdb37dff6c387a96dd5fc147973be

### Things this PR does NOT do

- Authenticate administrative clients

  Currently anyone with network access is able to setup/teardown/query enclaves on a running server.

  One way to fix this is to enable HTTPS, and check administrative clients's client certificates against an allow-list provided to the server on instantiation. These identities could be repurposed for policy roles.

  But I figured this is enough for one PR, and admin authentication can always be added later.

- Enable multiple parallel enclaves for all platforms

  I've conservatively prevented multiple parallel enclaves from most of the platforms since I wasn't sure which platforms can actually run multiple enclaves simultaneously. This check causes VeracruzServer{Platform} to error with TooManyEnclaves if more than one enclave is instantiated, returning HTTP 429 from the server.

  Except for SGX, locking SGX in this way actually caused the veracuz-server-test to fail, since it seems to run the tests in parallel on that platform. So I've allowed multiple SGX instances.
  
  If platforms do support multiple enclaves, they can remove the check I added to enable this behavior in the server.
  
---

Most of the technical details are now fleshed out, but this is open to changes

cc @alexandref75, @dreemkiller

### DIFF
--- a/veracruz-client/src/cli.rs
+++ b/veracruz-client/src/cli.rs
@@ -239,23 +239,23 @@ fn main() {
     // load policy
     let policy = match opt.policy_path.as_ref() {
         Some(policy_path) => {
-            let (policy_json, policy) = match
+            let (policy_json, policy, policy_hash) = match
                 fs::read_to_string(&policy_path)
                     .map_err(|err| PolicyError::from(err))
                     .and_then(|policy_json| {
-                        Policy::from_json(&policy_json)
-                            .map(|policy| (policy_json, policy))
+                        Policy::and_hash_from_json(&policy_json)
+                            .map(|(policy, policy_hash)| (policy_json, policy, policy_hash))
                     })
             {
-                Ok((policy_json, policy)) => (policy_json, policy),
+                Ok((policy_json, policy, policy_hash)) => (policy_json, policy, policy_hash),
                 Err(err) => {
                     eprintln!("{}", err);
                     process::exit(1);
                 }
             };
             qprintln!(opt, "Loaded policy {}", policy_path.to_string_lossy());
-            qprintln!(opt, "  (with hash) {}", policy.policy_hash().unwrap_or("???"));
-            Some((policy_json, policy))
+            qprintln!(opt, "  (with hash) {}", policy_hash);
+            Some((policy_json, policy, policy_hash))
         }
         None => None,
     };
@@ -265,17 +265,15 @@ fn main() {
     // if policy is provided and hashes mismatch, error
     //
     let policy_hash = match (opt.policy_hash.as_ref(), policy.as_ref()) {
-        (Some(hash), policy) => {
-            if let Some(policy_hash) = policy.and_then(|(_, policy)| policy.policy_hash()) {
-                if hash != policy_hash {
-                    eprintln!("Explicit hash does not match policy's hash");
-                    process::exit(1);
-                }
-            }
+        (Some(hash), Some((_, _, policy_hash))) if hash != policy_hash => {
+            eprintln!("Explicit hash does not match policy's hash");
+            process::exit(1);
+        }
+        (Some(hash), _) => {
             Some(hash.deref())
         }
-        (None, Some((_, policy))) => {
-            policy.policy_hash()
+        (None, Some((_, _, policy_hash))) => {
+            Some(policy_hash.deref())
         }
         _ => {
             None
@@ -285,7 +283,7 @@ fn main() {
     // figure out the URL to connect to
     let url = match (opt.url.as_ref(), policy.as_ref()) {
         (Some(url), _) => url,
-        (None, Some((_, policy))) => policy.veracruz_server_url(),
+        (None, Some((_, policy, _))) => policy.veracruz_server_url(),
         _ => unreachable!(),
     };
     qprintln!(opt, "Connecting to {}", url);
@@ -310,8 +308,8 @@ fn main() {
             lazy_veracruz_client.borrow_mut(),
             |ref_| {
                 ref_.get_or_insert_with(|| {
-                    let policy = match &policy {
-                        Some((_, policy)) => policy,
+                    let (policy, policy_hash) = match &policy {
+                        Some((_, policy, policy_hash)) => (policy, policy_hash),
                         None => {
                             eprintln!("Requires policy_path to interact with enclave");
                             process::exit(1);
@@ -332,7 +330,7 @@ fn main() {
                         identity,
                         key,
                         policy.clone(),
-                        policy.policy_hash().unwrap().to_string(),
+                        policy_hash.clone()
                     ) {
                         Ok(veracruz_client) => veracruz_client,
                         Err(err) => {
@@ -371,7 +369,7 @@ fn main() {
                     }
                 }
             }
-            (None, Some((policy_json, _))) => {
+            (None, Some((policy_json, _, _))) => {
                 policy_json.clone()
             }
             (None, None) => {
--- a/veracruz-server/src/server.rs
+++ b/veracruz-server/src/server.rs
@@ -82,8 +82,7 @@ async fn enclave_setup(
     enclave_handler: web::Data<Arc<Mutex<EnclaveHandler>>>,
 ) -> Result<String, VeracruzServerError> {
     // parse policy
-    let policy = Policy::from_json(&policy_json)?;
-    let policy_hash = policy.policy_hash().unwrap();
+    let (_, policy_hash) = Policy::and_hash_from_json(&policy_json)?;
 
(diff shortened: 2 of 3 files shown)


veracruz-client/src/admin_client.rs
```diff
@@ -0,0 +1,145 @@
...
+//!
+//! The Veracruz Development Team.
+//!
+//! ## Licensing and copyright notice
+//!
+//! See the `LICENSE_MIT.markdown` file in the Veracruz root directory for
+//! information on licensing and copyright.
+
+use crate::error::VeracruzClientError;
+use std::convert::identity;
```
dominic-mulligan-arm, line 13: Can these be grouped together?

veracruz-client/src/admin_client.rs
```diff
@@ -0,0 +1,145 @@
...
+
+
+/// Provides VeracruzAdminClient, a client API for administrative actions
+/// that don't require a policy
+///
+/// For interaction with the actual enclave see VeracruzClient
+///
+#[derive(Debug)]
+pub struct VeracruzAdminClient {
+    url: String,
```
dominic-mulligan-arm, line 27: Should this be a `String` or an `Url` type?

veracruz-client/src/admin_client.rs
```diff
@@ -0,0 +1,145 @@
...
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Serialize, Deserialize)]
+pub struct VeracruzEnclaveStatus {
+    pub policy_hash: String,
+    pub id: u32,
+    pub uptime: Duration,
+}
+
+impl VeracruzAdminClient {
+    /// Create a new VeracruzAdminClient to talk to the given server
+    pub fn new(url: &str) -> Self {
```
dominic-mulligan-arm, line 41: Should this parameter be `String` rather than taking a slice and allocating?

veracruz-client/src/cli.rs
```diff
@@ -11,27 +11,88 @@
...
-use veracruz_client::VeracruzClient;
 use veracruz_utils::policy::{
     error::PolicyError,
     parsers,
     policy::Policy,
 };
 
 
+/// A bit of extra parsing to allow omitting addr/port
+fn parse_bind_addr(s: &str) -> String {
```
dominic-mulligan-arm, line 38: Should this be a `String` too to avoid an extra allocation if the caller already has a `String`?

veracruz-client/src/veracruz_client.rs
```diff
@@ -186,6 +188,32 @@ impl VeracruzClient {
         policy: Policy,
         policy_hash: String,
     ) -> Result<VeracruzClient, VeracruzClientError>
+    where
+        P1: AsRef<path::Path>,
+        P2: AsRef<path::Path>
+    {
+        Self::with_url_id_policy_and_hash(
+            &policy.veracruz_server_url().to_owned(),
```
dominic-mulligan-arm, line 196: Is this not converting to a `String` and then immediately converting to a slice?  Can the `.to_owned()` be dropped?

veracruz-server/src/cli.rs
```diff
@@ -20,12 +20,70 @@ use veracruz_utils::policy::{
 };
 
 
+/// A bit of extra parsing to allow omitting addr/port
+fn parse_bind_addr(s: &str) -> String {
```
dominic-mulligan-arm, line 24: Is this function repeated?  Maybe move it out into e.g. `veracruz_utils` or similar.

veracruz-server/src/server.rs
```diff
@@ -17,46 +17,168 @@ use crate::veracruz_server_sgx::veracruz_server_sgx::VeracruzServerSGX as Veracr
...
+struct EnclaveHandler {
+    enclaves: HashMap<u32, EnclaveState>,
+    id_gen: Box<dyn Iterator<Item=u32> + Send>,
+    auto_shutdown_tx: Option<mpsc::Sender<()>>,
+}
+
+impl EnclaveHandler {
+    /// Find the best enclave handler id from the given query string, or zero
+    /// if no matching enclaves are found
+    fn best_id(&self, query: &str) -> Result<u32, VeracruzServerError> {
```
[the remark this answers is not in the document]
dominic-mulligan-arm: I think launching multiple "enclaves" is also possible for Linux.

veracruz-server/src/server.rs
```diff
@@ -17,46 +17,167 @@ use crate::veracruz_server_sgx::veracruz_server_sgx::VeracruzServerSGX as Veracr
...
+    let enclave_handler = enclave_handler.lock()?;
+    for (id, enclave) in enclave_handler.enclaves.iter() {
+        // Apparently SystemTime can error if time has moved backwards,
+        // if this happens (clock change?) we just show an uptime of zero
+        let uptime = enclave.start_time.elapsed()
+            .unwrap_or_else(|_| Duration::from_secs(0));
 
-    let enclave = enclave_handler_locked.as_mut().ok_or(VeracruzServerError::UninitializedEnclaveError)?;
+        // note we can't return a web::Json object here because Actix and Veracruz
+        // are actually using two incompatible versions of serde at the moment
```
dominic-mulligan-arm, line 144: Oof.  Is this an easy fix?

ShaleXIONG (approved):

dominic-mulligan-arm:
What is the status of this PR, @geky?

veracruz-server/src/server.rs
geky: This may be much easier to fix now with the workspaces.

veracruz-client/src/admin_client.rs
geky: See https://github.com/veracruz-project/veracruz/pull/242#discussion_r821761233

veracruz-client/src/admin_client.rs
geky: Do we have a `Url`? I don't think Rust has one. The closest would be `SocketAddr`, but this doesn't preserve the human readable representation.

geky:
Per offline discussion, we have decided to not bring this in for now. It's a big change, to both the code and how the CLIs interact, without a strong source of motivation.

It may be better to add these features individually as smaller PRs when needed.

I've moved these changes into the [`proposals/admin-separation`](https://github.com/veracruz-project/veracruz/tree/proposals/admin-separation) branch in case we want to refer to them in the future.

PR #243, opened 2021-10-19
Create the "veracruz_si" module alongside "wasi_snapshot_preview1"

egrimley-arm:
For now, the new module contains just `fd_create`, which opens a new anonymous file.

A test program in C is included.


egrimley-arm:
This PR is replaced by #267.

PR #248, opened 2021-10-26
Splitting veracruz-utils into smaller more focussed libraries

dominic-mulligan-arm:
Split `veracruz-utils` into smaller, more focussed libraries:
- `policy-utils`, containing material related to working with Veracruz policies, and their textual form,
- `io-utils`, containing common IO-related code.  At this time, it seems that only the Nitro backend makes use of this material.

What remains in `veracruz-utils`:
- Material related to common error types.  In previous discussions, there's a consensus that these error types are unsatisfactory.  However, I'll leave addressing that to future work.
- Material related to certificate signing requests (CSR).

As an additional benefit the contents of the various `Cargo.toml` files have been slimmed down as various dependencies contained within were no longer needed.


ShaleXIONG (approved):

dominic-mulligan-arm:
With some assistance from @mathias-arm I've rebased `runtime-manager-linux` on top of these changes and got everything compiling again.  Whilst I have some test failures in the `icecap` and `sgx` targets on that branch, they're unrelated to these changes.  In my opinion, this is now ready for merge.

@mathias-arm, I agree with both of your points, above.  I think the various `Cargo.toml` files are in a bit of a state and need both tidying up, and going through with a fine-toothed comb to make sure that everything being imported is actually being used.  I intend to do that, and do a full-project formatting pass, after the Linux changes are in...

io-utils/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+version = "0.3.0"
+authors = ["The Veracruz Development Team"]
+edition = "2018"
+description = "Common IO-related code used by multiple Veracruz components."
+
+[features]
+nitro = ["serde_json/std", "nix", "byteorder"]
+
+[dependencies]
+sgx_tstd = { rev = "v1.1.2", git = "https://github.com/apache/teaclave-sgx-sdk.git", optional = true }
```
mathias-arm, line 12: `io-utils` is for now exclusively nitro-specific. Why have `sgx_std` as an optional dependency?

mathias-arm (approved):

io-utils/Cargo.toml
dominic-mulligan-arm: I actually deleted that on my rebased `runtime-manager-linux` this morning.

dominic-mulligan-arm:
+1+1 = +1 reached, merging.

PR #249, opened 2021-10-26
Introduce stateful Docker configuration for IceCap.

nspin:
The original Docker configuration for hacking on IceCap uses a persistent `/nix` directory, which is bind-mounted into the container. This makes obtaining a clean Docker environment much simpler. However, bind mounts are unsuitable for this kind of use in the case of Docker on Mac. Not only are they very slow, but this kind of filesystem access results odd states that cause crashes.

This commit adds a so-called stateful configuration, where the `/nix` directory is located inside the container. This configuration is more robust on MacOS, at the cost of making obtaining a clean development environment more expensive.

This new stateful configuration is now the default.


dominic-mulligan-arm:
What is the status of this PR?  Is there a consensus that a volume-based approach would be better?

mathias-arm:
I like the volume-based approach, but I think the hacking and CI rules should be further unified or revisited. The difference between hacking and CI seems to focus on what version of IceCap is built. This seems to be useful for people who want to develop on the latest IceCap and CI systems, but this leaves out people who wants to develop and test Veracruz on the "stable" IceCap.

A more useful solution would have two orthogonal options:
- Choose latest version of IceCap considered stable for Veracruz or the latest upstream version when building the image.
- Have a make target to create a volume and make the docker commands will use it if it exists.

nspin:
Any difference in which version of IceCap is built between the hacking and CI images is incidental.

The primary difference between the two setups is how the uninteresting part of the build is cached.

For development, we use a Nix cache running nix-serve. Drawing from that cache in CI would be convenient, but would result in a lot of traffic to that server. A more cost efficient way to cache the uninteresting part of the build for CI is to pre-load the local Nix store in the Docker image itself instead. That's what the CI Docker setup you're looking at does.

nspin:
I changed this PR to use a Docker volume, in line with @ericvh's suggestion.

It is ready for review.

dominic-mulligan-arm (approved):
LGTM, though will wait for a second opinion before merging.

dominic-mulligan-arm:
@ShaleXIONG can you take a look at this?

mathias-arm:
This got essentially merged into https://github.com/veracruz-project/veracruz-docker-image/pull/29

dominic-mulligan-arm:
So is this PR completely defunct and can be closed?

mathias-arm:
IMHO, yes

PR #253, opened 2021-11-03
Normalized all Cargo/Xargo.toml files

dominic-mulligan-arm:
Folliowing a comment by @mathias-arm about the state of the project's `Cargo.toml` files, I've applied a formatting tool to our `Cargo.toml` and `Xargo.toml` files to normalise them.

### DIFF
--- a/execution-engine/Cargo.toml
+++ b/execution-engine/Cargo.toml
@@ -1,44 +1,53 @@
 [package]
+authors = ["The Veracruz Development Team"]
+description = "WASM execution engine for the trusted Veracruz runtime"
+edition = "2018"
 name = "execution-engine"
 version = "0.3.0"
-authors = ["The Veracruz Development Team"]
-edition = "2018"
-description = "WASM execution engine for the trusted Veracruz runtime"
 
 [features]
 default = []
 # NOTE: Introduce non_sgx in the dependencies to deal with sgx-related modification.
-std = ["wasmi/non_sgx", "platform-services/std", "wasmtime", "ring/non_sgx", "policy-utils/std", "wasi-types/std"]
-sgx = ["sgx_tstd", "wasmi/mesalock_sgx", "platform-services/sgx", "serde/mesalock_sgx", "typetag/mesalock_sgx", "ring/mesalock_sgx", "policy-utils/sgx", "wasi-types/sgx"]
-tz = ["platform-services/tz", "wasmi/non_sgx", "ring/non_sgx", "policy-utils/tz", "wasi-types/std"]
-nitro = ["platform-services/nitro", "policy-utils/std", "wasmtime", "wasi-types/std"]
 icecap = ["platform-services/icecap", "policy-utils/icecap", "wasmi/non_sgx"]
+nitro = ["platform-services/nitro", "policy-utils/std", "wasmtime", "wasi-types/std"]
+sgx = [
+  "sgx_tstd",
+  "wasmi/mesalock_sgx",
+  "platform-services/sgx",
+  "serde/mesalock_sgx",
+  "typetag/mesalock_sgx",
+  "ring/mesalock_sgx",
+  "policy-utils/sgx",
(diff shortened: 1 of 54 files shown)


mathias-arm:
I think the only part I like is the "features" normalization, but even then they are not sorted and some comments were removed.

I am not fond of:
- the change from double quotes to single quotes.
- the dependencies are not sorted neither in the `[dependencies]` section, nor the `[dependencies.*]` sections
- putting dependencies into separate section, I'd prefer keeping them on one line but having that line be consistently formatted.

dominic-mulligan-arm:
I've demoted this to a draft after a conversation with @mathias-arm and us looking for alternatives to `toml-fmt`, which is the tool that I used for formatting.

nspin:
The implementation [1] of `toml-fmt` doesn't seem particularly tailed for Cargo manifests.

I recently had to solve a similar problem in IceCap. After a quick survey, the best solution I could come up with was rolling my own little Python script that did the ordering and choosing of what dictionaries to inline [2].

[1] https://github.com/segeljakt/toml-fmt/blob/master/src/main.rs
[2] https://gitlab.com/arm-research/security/icecap/icecap/-/blob/main/nix/scope/crates/generate/append-to-manifest.py

mathias-arm:
There are a lot of tools that do some part of it. `taplo` is close, but is tripped up by comments in dependencies. `cargo-sort` is a nice complement (`cargo-sort` only does dependencies, `taplo` also sorts things like `package` and `features` sections).

```
$ cargo-sort <path/to/Cargo.toml>
$ taplo fmt -o align_entries=false,reorder_keys=true <path/to/Cargo.toml>
```

dominic-mulligan-arm:
How about now?  This was formatted using @mathias-arm's suggestion of a combination of `cargo-sort` and `taplo` above:

```
for F in `ls */Xargo.toml`; do cargo-sort $F && taplo fmt -o align_entries=false,reorder_keys=true $F; done;
```

If everybody is happy with this formatting then I can also promote this into a Git pre-commit check via an appropriate hook.

mathias-arm:
That's a lot better.  The only thing that looks out of place are those:

```
serde = { git = "https://github.com/veracruz-project/serde.git", features = [
  "derive",
], branch = "veracruz" }
```

mathias-arm:
@dominic-mulligan-arm, can you try with adding `column_width=120` option to `taplo` call?

dominic-mulligan-arm:
@mathias-arm, yeah, I'll try that.

dominic-mulligan-arm:
OK, done!

mathias-arm (approved):
That's good!

ShaleXIONG (approved):
Is there tool to do that? or it relies on us to maintain the normal form?

geky (approved):
Looks good to me, though I try to not have an opinion on style.

Since this is a rerunnable script I do think we should try to merge as many PRs before to save on rebasing headaches.

mathias-arm:
> Is there tool to do that? or it relies on us to maintain the normal form?

The tools used are `cargo-sort` and `taplo` (`cargo-sort <file>` and `taplo fmt -o align_entries=false,reorder_keys=true, column_width=120 <file>`).

mathias-arm:
> Since this is a rerunnable script I do think we should try to merge as many PRs before to save on rebasing headaches.

Since this is an automated process you can also run it at the end of the branch and squash your branch (not sure if you could rebuild the whole branch and keep distinct commits). This way, there should not be any rebase headaches. Same thing should be possible with `rustfmt` changes.

mathias-arm:
Adding `array_auto_collapse=false` to taplo command line prevents it from putting smaller arrays into one line.
`taplo fmt -o align_entries=false,array_auto_collapse=false,reorder_keys=true,column_width=120`

nspin (approved):

dominic-mulligan-arm:
+1+1+1+1 = +1 reached, merging.

PR #256, opened 2021-11-05
Cargo Git configuration improvements

dominic-mulligan-arm:
Configured Cargo to:
- Use the Git command line tool rather than the Rust `libgit2` library,
- Retry all network attempts 3 times.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -27,3 +27,7 @@ rustflags = [
 [target.arm-unknown-linux-gnueabihf]
 linker = "/work/rust-optee-trustzone-sdk/optee/toolchains/aarch32/bin/arm-linux-gnueabihf-gcc"
 ar = "/work/rust-optee-trustzone-sdk/optee/toolchains/aarch32/bin/arm-linux-gnueabihf-gcc-ar"
+
+[net]
+retry = 3
+git-fetch-with-cli = true


egrimley-arm:
There's an RFC, at least, for sharing stuff between `Cargo.toml` files, but I don't think it's been implemented yet, unfortunately: https://github.com/rust-lang/cargo/issues/8415

nspin:
I think this is actually a `.cargo/config` option:

https://doc.rust-lang.org/cargo/reference/config.html#net

I think we'll only need to include it once, just in the top-level `.cargo/config`:

https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure

dominic-mulligan-arm:
Ah, that makes more sense than amending every `Cargo.toml` file.  I'll push a change in a bit.

egrimley-arm:
So I tried putting the `git-fetch-with-cli` in `veracruz/.cargo/config` and running `cargo clean; cargo build` in `sdk/freestanding-execution-engine`, and it made no difference, though the `config` was parsed because I got a warning when I deliberately mistyped the thing.

When I say it made no difference I mean the script called `git` that was on my path got executed just two times, with the arguments `rev-parse HEAD`, in each case, with or without the `config` change. Also I didn't seen any signs of `git` being invoked when I tried `strace -f -o log`.

What am I doing wrong?

egrimley-arm:
What I'm doing wrong is not deleting the cache from my `~/.cargo/`!

egrimley-arm:
It now seems to work, though the `retry = 3` probably won't make it retry `git` or cause `git` to retry. But presumably it doesn't hurt, either.

egrimley-arm:
Conclusion: Just add the three lines to `veracruz/.cargo/config` instead.

dominic-mulligan-arm:
OK, done!

nspin (approved):

mathias-arm (approved):

dominic-mulligan-arm:
+1+1 = +1 reached, merging.

PR #260, opened 2021-11-10
WIP: native module

ShaleXIONG:


execution-engine/src/fs.rs
```diff
@@ -857,7 +1030,7 @@ impl FileSystem {
             .0)
     }
 
-    ////////////////////////////////////////////////////////////////////////////
+    ///////////////////////////////////////////////////////////////s/////////////
```
egrimley-arm, line 1033: Cat ran across the keyboard there?

ShaleXIONG: hahaha... sadly it is my own hand.

egrimley-arm:
Do you have an example program and command line for testing this code?

ShaleXIONG:
> Do you have an example program and command line for testing this code?

I just commit the example. It is a super simple demonstration.

execution-engine/src/fs.rs
```diff
@@ -144,11 +147,32 @@ impl InodeEntry {
     pub(self) fn len(&self) -> FileSystemResult<FileSize> {
         self.data.len()
     }
+
+    /// Return the number of bytes, if it is a file,
```
egrimley-arm, line 151: This comment is not right!

execution-engine/src/fs.rs
```diff
@@ -353,6 +431,31 @@ impl InodeTable {
         Ok(rst)
     }
 
+    /// Install all the (file_system_handler, input_paths, output_path, service_instance) tuples. Assume `path` is an absolute path to a (special) file.
+    /// NOTE: this function is intended to be called after the root filesystem (handler) is
+    /// created, since each service is assocaited a child filesystem (handler).
```
egrimley-arm, line 436: assocaited -> associated with

execution-engine/src/fs.rs
```diff
@@ -574,6 +680,14 @@ impl InodeTable {
         }
         Ok(())
     }
+
+    //fn read_file(&mut self, inode: Inode, max: usize, offset: FileSize) -> FileSystemResult<Vec<u8>> {
```
egrimley-arm, line 684: This function should presumably be really deleted.

execution-engine/src/fs.rs
```diff
@@ -647,13 +775,23 @@ impl FileSystem {
...
+        ));
+        rst.install_services(services)?;
         Ok(rst)
     }
 
     /// This is the ONLY public API to create a new FileSystem (handler).
     /// It return a FileSystem where directories are pre-opened with appropriate
     /// capabilities in related to `principal`.
     pub fn spawn(&self, principal: &Principal) -> FileSystemResult<Self> {
+        println!("spawn on: {:?}", principal);
```
egrimley-arm, line 794: Are all the `println!`s added just temporary debug code that will be removed before we merge?

egrimley-arm (commented):
Some similar changes have been made in the `.toml` files on `main` so this PR might get simpler when it's rebased.

Some nitpicking on comments.

This PR should presumably be labelled as "Draft" rather than "Review required"?

execution-engine/src/fs.rs
ShaleXIONG: It is temporary.

execution-engine/src/fs.rs
```diff
@@ -270,9 +317,10 @@ impl InodeImpl {
 
     /// Return the number of the bytes, if it is a file,
     /// or the number of inodes, if it it is a directory.
+    #[inline]
     pub(self) fn len(&self) -> FileSystemResult<FileSize> {
         let rst = match self {
-            Self::File(f) => f.len(),
+            Self::NativeModule(.., f) | Self::File(f) => f.len(),
             Self::Directory(f) => f.len(),
```
dreemkiller, line 324: Out of curiosity, why is this `match` statement even necessary? Since each block is doing the same thing (calling `f.len()`), why can't we just call `f.len()` without the match? Is it something weird with enums?

execution-engine/src/fs.rs
```diff
@@ -287,6 +335,44 @@ impl InodeImpl {
...
+    pub(crate) fn is_service(&self) -> bool {
+        match self {
+            Self::NativeModule(..) => true,
+            _ => false,
+        }
+    }
+
+    /// Try to parse the input of a service. Return true if it is a valid input.
+    #[inline]
+    pub(crate) fn service_valid_input(&self) -> FileSystemResult<bool> {
```
dreemkiller, line 349: When calling this function, we fully parse the input. Do we save the output of that parsing, so we won't have to parse it again when we consume it?

execution-engine/src/native_modules/postcard.rs
```diff
@@ -0,0 +1,83 @@
+use crate::fs::{FileSystem, FileSystemResult, Service};
```
dreemkiller, line 1: Doesn't this file need the copyright notice and authors list?

dreemkiller: Also, there's no documentation in this file, and the identifiers are unhelpful (`E1`, `T1`, etc). A naive reading of the file gives me no information

sdk/data-generators/postcard-generator/src/main.rs
```diff
@@ -0,0 +1,167 @@
...
+//! cargo run -- --file [STRING] --size [VEC_SIZE] --seed [RANDOM_SEED];
+//! ```
+
+use clap::{App, Arg};
+use rand::{rngs::StdRng, Rng, SeedableRng};
+use serde::{Deserialize, Serialize};
+use std::{error::Error, fs, string::String, vec::Vec};
+
+#[derive(Deserialize, Serialize, Clone, Debug)]
+enum E1 {
```
dreemkiller, line 23: Aren't many of the items in this file duplicates to the items in the `postcards.rs` file? Is there a reason they are repeated here?

execution-engine/src/fs.rs
dominic-mulligan-arm: The `f` comes from the `match` in each case (i.e. it's not the same `f` in each branch of the `match`).

execution-engine/src/fs.rs
ShaleXIONG: The native module might need to parse it again, depending on how the native module implements the interface `try_parse`, where the native module could buffer previous results in their own space.

I am using this function as a `try-parse` pattern. When it is successful (returning true), it means there is enough input for the native module. This is because we are triggering the native module on writing to a special file, and there is no guarantee a single write (in high-level source code) will be done in one invocation of `fd_write` (in the wasi level). Essentially, this function is called on every invocation to check if there is enough.

dreemkiller: Understood.
So the native module might save the result of a successful `try-parse`. That's the preferred way to avoid multiple successful parses.

execution-engine/src/native_modules/postcard.rs
ShaleXIONG: Will add the copyright. It is a made-up type. I will give a more descriptive name.

sdk/rust-examples/postcard-wasm/src/main.rs
```diff
@@ -0,0 +1,60 @@
+use postcard::from_bytes;
```
dominic-mulligan-arm, line 1: Missing license/copyright header and description of what the file is.

test-collateral/generate-policy/src/main.rs
```diff
@@ -86,14 +86,9 @@ const POLICY_CIPHERSUITE: &'static str = "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305
...
 const DEFAULT_OUTPUT_FILENAME: &'static str = "output.json";
-/// The default debug status of the Veracruz enclave, if no alternative is
-/// provided on the command line.
-const DEFAULT_DEBUG_STATUS: bool = false;
 /// The default execution strategy for the WASM binary, if no alternative is
 /// provided on the command line.
-const DEFAULT_EXECUTION_STRATEGY: &'static str = "Interpretation";
-/// The default clock flag, if no alternative is provided on the command line.
-const DEFAULT_ENABLE_CLOCK: bool = false;
+const DEFAULT_EXECUTION_STRATEGY: &'static str = "JIT";
```
dominic-mulligan-arm, line 91: Can you add a `rustdoc` comment explaining what this is?

workspaces/host/Makefile
```diff
@@ -20,6 +20,7 @@ all: crates/sdk/wasm-checker/bin/wasm-checker
 DATA_GENERATORS = idash2017-logistic-regression-generator \
                       intersection-set-sum-generator \
                       moving-average-convergence-divergence-generator \
+				      postcard-generator \
```
dominic-mulligan-arm, line 23: Bad indenting?

execution-engine/src/native_modules/postcard.rs
```diff
@@ -0,0 +1,83 @@
...
+use postcard::from_bytes;
```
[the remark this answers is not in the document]
ShaleXIONG: The native module is part of the runtime. For this demonstration, we deserialise the postcard encoding and serializing to JSON string. Postcard encoding requires the type in compilation time rather than dynamically detecting the type in runtime.

I will expect this demonstration to be removed and replaced by more realistic encryption and decryption.

sdk/rust-examples/postcard-wasm/src/main.rs
```diff
@@ -0,0 +1,60 @@
...
+use serde::{Deserialize, Serialize};
+use std::fs;
+use std::string::String;
+use std::vec::Vec;
+
+#[derive(Deserialize, Serialize, Clone, Debug)]
```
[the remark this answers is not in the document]
ShaleXIONG: it is only for tests for sure.
The problem is when deserialising a postcard encoding, the type must be statically decided. Many languages have features that supports functions like `deserialise(input, some_presentation_of_type)`, while I cannot see how Rust implement this.
Hence the test native module is tailored for a specific type, and the type definition is copied through the data generator, the native module and here.

execution-engine/src/fs.rs
```diff
@@ -173,9 +206,10 @@ impl InodeImpl {
 
     /// Resize a file to `size`, and fill with `fill_byte` if it grows.
     /// Otherwise, return ErrNo::IsDir if it is not a file.
+    #[inline]
     pub(self) fn resize_file(&mut self, size: FileSize, fill_byte: u8) -> FileSystemResult<()> {
         match self {
-            Self::File(file) => {
+            Self::NativeModule(.., file) | Self::File(file) => {
```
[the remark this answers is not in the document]
ShaleXIONG: I think the special file should behaviour as closely as a normal file, but we can use the capabilities to restrict if a resize is allowed, i.e. `fd_filestat_set_size` and `path_filestat_set_size`.

execution-engine/src/fs.rs
```diff
@@ -234,10 +278,11 @@ impl InodeImpl {
     }
 
     /// Truncate the file.
-    /// Return ErrNo::IsDir if it is not a file.
+    /// Return ErrNo::IsDir if it is a dir.
+    #[inline]
     pub(self) fn truncate_file(&mut self) -> FileSystemResult<()> {
         match self {
-            Self::File(b) => {
```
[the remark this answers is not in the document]
ShaleXIONG: I think the special file should behaviour as closely as a normal file, but we can use the capabilities to restrict if a trunc is allowed, i.e. `fd_filestat_set_size`.

dominic-mulligan-arm:
@ShaleXIONG can you rebase to get rid of the conflict?

dominic-mulligan-arm:
Thanks!

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
+1 reached, merging.

PR #262, opened 2021-11-15
Made tz_env.sh executable

dominic-mulligan-arm:
Note that the build instructions for TrustZone are out-of-date, as an environment variable for the C compiler must now also be set before building TrustZone otherwise an obscure compiler/linker error is encountered.  The correct, but undocumented, way to setup the build environment is to run `tz_env.sh`.  Unfortunately, this is not an executable shell script.  This changeset fixes that, and an accompanying change on the Docker image repository updates the build instructions.


nspin:
GitHub says this is an empty file.

Also, to populate the environment, wouldn't the script be sourced (with `source` or `.`) rather than executed?

dominic-mulligan-arm:
Yeah, thinking more you are right that `source`ing the script is what's needed.  I'll close this and update the `README` on the Docker repository as appropriate.

PR #265, opened 2021-11-16
Fix pgen

gbryant-arm:
Change short name for --enable-clock to avoid collisions with other arguments (certificates).
Fixes crash when running pgen aka generate-policy

### DIFF
--- a/test-collateral/generate-policy/src/main.rs
+++ b/test-collateral/generate-policy/src/main.rs
@@ -332,7 +332,7 @@ binary.",
         )
         .arg(
             Arg::with_name("enable-clock")
-                .short("c")
+                .short("n")
                 .long("enable-clock")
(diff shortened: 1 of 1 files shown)


dominic-mulligan-arm (commented):
Has the `test-collateral` Makefile, and anything else that may be calling `pgen`, been checked and updated to use the new flag, as appropriate?

gbryant-arm:
Nothing seems to call `pgen` with short arguments.
I haven't run the tests, will wait for the CI.

dominic-mulligan-arm:
This is failing due to `openssl-sys` updating earlier today and bumping the MSRV.  You will need to wait until #266 is merged.

dominic-mulligan-arm (approved):

PR #269, opened 2021-11-18
Adopt copy-less APIs in the VFS

geky:
This mostly required changes to the MemoryHandler trait in order to allow direct access to the WebAssembly's linear memory space. This turned out to be surprisingly complicated because of Rust's borrow checker and differences between Wasmi's and Wasmtime's linear memory APIs (though this does mean this code should be portable to other execution engines).

I've tried to document the code thoroughly as this borders on what's possible in Rust. And considering the amount of transmutes I needed I may be doing something terribly wrong, but this was the best solution I could come up with to provide safe access to engine memories safely. I'm open to other ideas if there are any.

---

- Added `get_slice` and `get_slice_mut`, which return associated types `Slice` and `SliceMut`. These represent slices of the underlying linear memory, but can other contexts that may be required by the execution engine such as Mutex/RefCell guards, Rcs, or even a Vec if the underlying engine doesn't actually have a linear memory.

  There's a bit more trickiness here due to lifetimes.
  
  `Slice` and `SliceMut` share the lifetime of the linear memory they take a slice of. The proper way to describe this in Rust would be like this:
  
  ``` rust
  trait MemoryHandler {
      type Slice<'a>: AsRef<[u8]>;
      fn get_slice<'a>(&'a self) -> Slice<'a>;
  }
  ```
  
  But this requires a nightly, and unfinished, feature called GATs (the `type Slice<'a>: AsRef<[u8]>` part specifically):
  
  https://github.com/rust-lang/rust/issues/44265
  
  The workaround here is to require the associated-type implementations to all be "lifetime-less", which probably requires unsafe code in the implementations.
  
  To keep this API from being just completely unsafe (we could just pass around pointers). I've added a new type called Bound for the sole purpose of binding an unrelated lifetime to a lifetime-less object. Whether or not the lifetime is valid can't be checked by the compiler of course, but this at least means any consuming code will be borrow-checked correctly:
  
  ``` rust
  trait MemoryHandler {
      type Slice: AsRef<[u8]>;
      fn get_slice<'a>(&'a self) -> Bound<'a, Slice>;
  }
  ```

- Added `unpack_iovec` and `unpack_iovec_mut` which return _sets_ of these slices. This allows iovecs to be passed all the way from the WebAssembly application to the internals of the filesystem, allowing multiple iovecs to be consumed in a single filesystem operation.

  This also handles the splitting of the linear memory's mutable borrow, which is difficult to do otherwise.

- Required MemoryHandler::Slice and MemoryHandler::SliceMut to be _movable_, even behind a borrow. This is a very _very_ unusual trait requirement but prevents the need for unnecessary boxing in the hot-path of file read/writes. Fortunately it holds true for most of the types you would use in this API, such as slices and even vecs.

- Added small-object optimization (SOO) for 1 and 2 element iovecs (these are by far the most common from personal experience, I've never actually seen a >2 iovec used before, though it's certainly possible in hand-optimized code)

---

The end result is zero allocations in the hot-path of file reads/writes, aside from the internal allocations required for the filesystem's internal data-structures.

This results in a significant performance gain, mostly due to no longer copying the read/write buffers multiple times before the data reaches the filesystem.

image

### DIFF
--- a/execution-engine/src/fs.rs
+++ b/execution-engine/src/fs.rs
@@ -220,9 +220,8 @@ impl InodeImpl {
         //             ----------------------------------
         //             | buf                            |
         //             ----------------------------------
-        let remain_length = bytes.len() - offset;
-        if buf.len() > remain_length{
-            bytes.resize(buf.len() - remain_length, 0);
+        if offset + buf.len() > bytes.len() {
+            bytes.resize(offset + buf.len(), 0);
         }
         let write_length = buf.len();
         bytes[offset..(offset + write_length)].copy_from_slice(&buf);
--- a/execution-engine/src/fs.rs
+++ b/execution-engine/src/fs.rs
@@ -21,8 +21,9 @@ use policy_utils::principal::{FileRights, Principal, RightsTable};
 use std::os::unix::ffi::{OsStrExt, OsStringExt};
 
 use std::{
+    cmp::min,
     collections::HashMap,
-    convert::{AsRef, TryFrom},
+    convert::{AsRef, TryFrom, TryInto},
     ffi::OsString,
     // TODO: wait for icecap to support direct conversion between bytes and os_str, bypassing
     // potential utf-8 encoding check
@@ -84,15 +85,15 @@ impl InodeEntry {
     /// Read maximum `max` bytes from the offset `offset`.
     /// Return ErrNo::IsDir if it is not a file.
     #[inline]
-    pub(self) fn read_file(&self, max: usize, offset: FileSize) -> FileSystemResult<Vec<u8>> {
-        self.data.read_file(max, offset)
+    pub(self) fn read_file(&self, buf: &mut [u8], offset: FileSize) -> FileSystemResult<usize> {
+        self.data.read_file(buf, offset)
     }
 
     /// Write `buf` to the file from the offset `offset`,
     /// update the file status and return the number of written bytes.
     /// Otherwise, return ErrNo::IsDir if it is not a file.
     #[inline]
-    pub(self) fn write_file(&mut self, buf: Vec<u8>, offset: FileSize) -> FileSystemResult<Size> {
+    pub(self) fn write_file(&mut self, buf: &[u8], offset: FileSize) -> FileSystemResult<usize> {
         let rst = self.data.write_file(buf, offset)?;
         self.file_stat.file_size = self.data.len()?;
         Ok(rst)
@@ -179,7 +180,7 @@ impl InodeImpl {
 
     /// Read maximum `max` bytes from the offset `offset`.
     /// Otherwise, return ErrNo::IsDir if it is not a file.
-    pub(self) fn read_file(&self, max: usize, offset: FileSize) -> FileSystemResult<Vec<u8>> {
+    pub(self) fn read_file(&self, buf: &mut [u8], offset: FileSize) -> FileSystemResult<usize> {
         let bytes = match self {
             Self::File(b) => b,
             Self::Directory(_) => return Err(ErrNo::IsDir),
@@ -196,18 +197,14 @@ impl InodeImpl {
         //             | rst                  |
         //             ------------------------
         let (_, to_read) = bytes.split_at(offset);
-        let read_length = if max < to_read.len() {
-            max
-        } else {
-            to_read.len()
-        };
-        let (rst, _) = to_read.split_at(read_length);
-        Ok(rst.to_vec())
+        let read_length = min(buf.len(), to_read.len());
+        buf[..read_length].copy_from_slice(&to_read[..read_length]);
+        Ok(read_length)
     }
 
     /// Write `buf` to the file from the offset `offset` and return the number of written bytes.
     /// Otherwise, return ErrNo::IsDir if it is not a file.
-    pub(self) fn write_file(&mut self, buf: Vec<u8>, offset: FileSize) -> FileSystemResult<Size> {
+    pub(self) fn write_file(&mut self, buf: &[u8], offset: FileSize) -> FileSystemResult<usize> {
         let bytes = match self {
             Self::File(b) => b,
             Self::Directory(_) => return Err(ErrNo::IsDir),
@@ -224,13 +221,12 @@ impl InodeImpl {
         //             | buf                            |
         //             ----------------------------------
         let remain_length = bytes.len() - offset;
-        if remain_length <= buf.len() {
-            let mut grow_vec = vec![0; buf.len() - remain_length];
-            bytes.append(&mut grow_vec);
+        if buf.len() > remain_length{
+            bytes.resize(buf.len() - remain_length, 0);
         }
-        let rst = buf.len();
-        bytes[offset..(offset + rst)].copy_from_slice(&buf);
-        Ok(rst as Size)
+        let write_length = buf.len();
+        bytes[offset..(offset + write_length)].copy_from_slice(&buf);
+        Ok(write_length)
     }
 
     /// Truncate the file.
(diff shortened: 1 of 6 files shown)


execution-engine/src/fs.rs
```diff
@@ -998,18 +993,27 @@ impl FileSystem {
...
         self.check_right(&fd, Rights::FD_READ)?;
         let inode = self.fd_table.get(&fd).ok_or(ErrNo::BadF)?.inode;
 
-        self.lock_inode_table()?
-            .get(&inode)?
-            .read_file(buffer_len, offset)
+        let f = self.lock_inode_table()?;
+        let f = f.get(&inode)?;
+
+        let mut offset = offset;
```
dominic-mulligan-arm, line 1008: You can mark the parameter `offset` as `mut` here, instead.

execution-engine/src/fs.rs
```diff
@@ -1035,29 +1039,42 @@ impl FileSystem {
...
         self.check_right(&fd, Rights::FD_WRITE)?;
         let inode = self.fd_table.get(&fd).ok_or(ErrNo::BadF)?.inode;
 
-        self.lock_inode_table()?
-            .get_mut(&inode)?
-            .write_file(buf, offset)
+        let mut f = self.lock_inode_table()?;
+        let f = f.get_mut(&inode)?;
+
+        let mut offset = offset;
```
dominic-mulligan-arm, line 1054: Same comment as above regarding `mut offset`.

execution-engine/src/fs.rs
```diff
@@ -998,18 +993,27 @@ impl FileSystem {
...
-            .get(&inode)?
-            .read_file(buffer_len, offset)
+        let f = self.lock_inode_table()?;
+        let f = f.get(&inode)?;
+
+        let mut offset = offset;
+        let mut len = 0;
+        for buf in bufs {
+            let delta = f.read_file(buf.as_mut(), offset)?;
+            offset += u64::try_from(delta).unwrap();
```
dominic-mulligan-arm, line 1012: When does this fail?

execution-engine/src/wasi/common.rs
```diff
@@ -150,12 +152,15 @@ impl TryFrom<&str> for WasiAPIName {
 ////////////////////////////////////////////////////////////////////////////////
 
 /// Unpack a sequence of `bytes` and return a `T`.
-pub trait Unpack<T> {
-    fn unpack(bytes: &[u8]) -> FileSystemResult<T>;
+pub trait Unpack: Sized {
+    const SIZE: u32;
```
dominic-mulligan-arm, line 156: What is the purpose of this `SIZE`?  Can we get `rustdoc` comments on the methods/associated constants of the trait?

execution-engine/src/wasi/common.rs
```diff
@@ -247,52 +256,328 @@ impl Unpack<Event> for Event {
...
+
+/// Number of iovecs to store internally before needing dynamic memory
+const IOVECSLICES_SOO_COUNT: usize = 2;
+
+/// A type wrapping an array of IoVecs as directly-accessible slices.
+///
+/// This indirection is needed to hold the MemoryHandler::Slice type which
+/// needs to be stored somewhere
+///
+/// This may allocated, though most iovecs are fairly small. As an extra
```
dominic-mulligan-arm, line 389: sp: allocated --> allocate

execution-engine/src/wasi/common.rs
```diff
@@ -247,52 +256,328 @@ impl Unpack<Event> for Event {
...
-    fn unpack_array<T: Unpack<T>>(&self, ptr: u32, count: u32) -> FileSystemResult<Vec<T>> {
-        let size = size_of::<T>();
-        let all_bytes = self.read_buffer(ptr, count * (size as u32))?;
-        let mut rst = Vec::new();
+/// A type wrapping a mutable array of IoVecs as directly-accessible slices.
+///
+/// This indirection is needed to hold the MemoryHandler::SliceMut type which
+/// needs to be stored somewhere
+///
+/// This may allocated, though most iovecs are fairly small. As an extra
```
dominic-mulligan-arm, line 430: sp: allocated --> allocate

dominic-mulligan-arm (commented):
This looks great, really nice @geky!  Just a few minor comments.

execution-engine/src/wasi/common.rs
```diff
@@ -247,52 +256,328 @@ impl Unpack<Event> for Event {
...
+        address: u32,
+        length: u32
+    ) -> FileSystemResult<BoundMut<'a, Self::SliceMut>>;
+
+    /// Get the size of the underlying memory
+    fn get_size(&self) -> FileSystemResult<u32>;
+
+    /// Write the `buffer` to `address`.
+    fn write_buffer(&mut self, address: u32, buffer: &[u8]) -> FileSystemResult<()> {
+        self.get_slice_mut(address, u32::try_from(buffer.len()).unwrap())?
```
ShaleXIONG, line 560: unrwap can be converted to ErrNo::inval

execution-engine/src/wasi/wasmi.rs
```diff
@@ -40,25 +40,60 @@ impl HostError for FatalEngineError {}
...
-        Ok(())
+    type Slice = &'static [u8];
+    type SliceMut = &'static mut [u8];
+
+    fn get_slice<'a>(
+        &'a self,
+        address: u32,
+        length: u32
+    ) -> FileSystemResult<Bound<'a, Self::Slice>> {
+        let address = usize::try_from(address).unwrap();
```
ShaleXIONG, line 51: unwrap can be converted to ErrNo::Inval.

execution-engine/src/wasi/common.rs
```diff
@@ -447,7 +923,7 @@ impl WasiWrapper {
     }
 
     /// The implementation of the WASI `args_sizes_get` function. It requires an extra `memory_ref` to
-    /// interact with the execution engine.
+    /// interact with the execution engine.
```
mathias-arm, line 926: There are a few occurrences of that extra whitespace

execution-engine/src/fs.rs
geky: You can but that mixes implementation details with the function signature which IMO isn't great.

Here's a really terrible/ridiculous example of a function that does not take a mutable reference:
``` rust
fn do_the_thing(mut x: &[u8]) {
    ...
}
```

Happy to move the mut to the arguments if you feel strongly but I tend to re-declare locals as mut in the function for that reason.

execution-engine/src/wasi/wasmi.rs
geky: I'll run a sed command on all of these

execution-engine/src/wasi/common.rs
geky: Oh strange, these changed a lot more during development, so I think it was a mix of rustfmt changes on main and a mistake during rebasing. Can remove

execution-engine/src/fs.rs
geky: These really can't fail, it's just safer than using `delta as u64` in case someone does something crazy like run Veracruz on a 16-bit system. Or more likely some bug introducing some out-of-range value.

Just like bound-checking, these should (in theory (emphasis on theory)) get elided in many cases by the compiler.

I can adopt `try_from_or_errno` here.

execution-engine/src/wasi/common.rs
geky: Ah yes can do. Previously the Unpack trait relied on size_of to determine the size to read from the linear memory. Having a separate associated constant means you can unpack things that have different sizes in/outside of Wasm.

execution-engine/src/fs.rs
geky: Hmm, alternatively could we add a conversion from TryFromIntError for Errno? Then `u64::try_from(x)?` would be valid here.

geky: Ah, no we can't since ErrNo is outside of the current crate. Hmm.

I went ahead and changed it to a trait so it's a bit easier to replace `<_>::try_from(x).unwrap()` calls with, but let me know if you want me to revert that. Should be updated now to use `<_>::try_from_or_errno(x)?`

execution-engine/src/fs.rs
dominic-mulligan-arm: No, I think that makes sense.  I'll mark this and the other comment as resolved.

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
@ShaleXIONG, @mathias-arm are you happy with this?

ShaleXIONG (approved):

dominic-mulligan-arm:
+1+1 = +1 reached, merging.

PR #270, opened 2021-11-18
Test if we can call psa_crypto::init.

egrimley-arm:
The "system call" test_psa_crypto is added, in the same way that
fd_create was added, but it takes no arguments and just calls
psa_crypto::init.

rust-examples/linear-regression is modified so that it calls
that function.


egrimley-arm:
It gets rejected by `wasm-checker`, of course!

PR #271, opened 2021-11-22
Added the Veracruz project technical charter

dominic-mulligan-arm:
This is apparently a requirement of the CCC.


egrimley-arm (approved):

egrimley-arm:
I'll just mention here that the PDF cannot be viewed with `evince`. That seems to be a feature of all documents that have been mangled by DocuSign. I've not tried testing the file with a PDF validator, so I don't know who's to blame here. If anyone happens to understand the problem, I'd be interested to hear any explanations, but it's not a problem for Veracruz. The file can be viewed on Linux with `mupdf`, for example.

gbryant-arm (approved):

egrimley-arm:
Oh, another transient network problem?

`fatal: unable to access 'https://github.com/rust-lang/rls.git/': Failed to connect to github.com port 443: Connection timed out`

dominic-mulligan-arm:
Yes, we seem to be suffering from a networking affliction, again: yesterday all of the `clone` operations for the `icecap` repository were failing.  We've got past that, but now we are hitting other issues.

ShaleXIONG (approved):

PR #268, opened 2021-11-17
Remove redundant imports from project components

dominic-mulligan-arm:
Removed redundant imports, left from previous refactorings and changes in the codebase, from all project component `Cargo.toml` files.

PR #274, opened 2021-11-24
In veracruz-server/Cargo.toml specify "veracruz" branch for "serde".

egrimley-arm:
Currently "veracruz" is the default branch anyway, but it seems
dangerous to not specify the branch here when it is specified
everywhere else.


dominic-mulligan-arm:
The `serde` dependency has been removed in my branch (#268) as it's redundant.

dominic-mulligan-arm:
@egrimley-arm are you happy for this to be closed?

PR #275, opened 2021-11-26
runtime-manager: Use a separate directory for the Dockerfile.

egrimley-arm:
The main motivation for this is that the rust-psa-crypto crate
makes symbolic links under target/.../build/, and these cause
nitro-cli to get confused and stop with error E49.

The single file needed by the Dockerfile is hard-linked into
dockerdir/. We trust that nitro-cli is not confused by hard links.
However, see https://xkcd.com/2531/

Also, use "--output-file $@.new", then "mv $@.new $@". This means
that if nitro-cli fails for any reason then a subsequent invocation
of "make" will not think that the target file is up-to-date.


geky:
Since `dockerdir/runtime_manager_enclave` is hard-linked at build time, should it be in the .gitignore? That might help prevent someone from accidentally checking it in.

Other than that it looks good to me, though I'm not that familiar with nitro-cli or the Nitro builds

egrimley-arm:
Yes. Done.

geky (dismissed):
Looks good to me then 👍

geky:
Looks good to me still :+1:

geky (approved):

PR #277, opened 2021-11-26
Reduce latency in the network stack

gbryant-arm:
Reduce latency in the network stack, specifically on the runtime manager and client sides.
- Prefix every protocol buffer (runtime manager & proxy attestation messages) with their length to allow the receiver to know when a protocol buffer is complete. This avoids the need to try to deserialize the protocol buffer until it succeeds, which is more and more time consuming as the incoming buffer grows (it takes about 100ms to deserialize a 30MB protocol buffer)
- Turn 'TCP no delay' on to reduce latency on the Linux backend, the only one for which the server and the runtime manager communicate over TCP. This is effective when small packets are regularly sent, which is the case here. This measure decreases latency from 40ms down to <1ms

Note that a high latency between the client and server will still have a big impact on throughput, as 16KB protobuf fragments –split by TLS – are still sent between the client and runtime manager one by one.
The obvious solution is to remove HTTP on the client/server side and replace it with a plain TCP repeater.
Other ideas:
* The client could encapsulate groups of fragments in each HTTP request to the server, which might require to configure the TLS server to handle groups of TLS records. This would have the downside of increasing memory usage on the server.
* Remove base64 (de)serializing, which could shave off a few hundreds of microseconds here and there

All these suggestions are however out of the scope of this PR.

TODO:
* Benchmark "no length prefix" vs "length prefix" to validate changes
* Avoid unnecessary conversions between arrays, vectors and mutable vectors

### DIFF
--- a/runtime-manager/src/managers/execution_engine_manager.rs
+++ b/runtime-manager/src/managers/execution_engine_manager.rs
@@ -25,7 +25,7 @@ use transport_protocol::transport_protocol::{
 
 lazy_static! {
     // TODO: wrap into a runtime manager management object.
-    static ref INCOMING_BUFFER_HASH: Mutex<HashMap<u32, Vec<u8>>> = Mutex::new(HashMap::new());
+    static ref INCOMING_BUFFER_HASH: Mutex<HashMap<u32, (u64, Vec<u8>)>> = Mutex::new(HashMap::new());
 }
 
 ////////////////////////////////////////////////////////////////////////////////
@@ -149,10 +149,15 @@ fn dispatch_on_request(client_id: u64, request: MESSAGE) -> ProvisioningResult {
     }
 }
 
-/// Tries to parse the incoming data into a `RuntimeManagerRequest`.  If this is not
-/// possible, returns `Err(reason)`.  If we still need to receive more data in
-/// order to parse a full request, returns `Ok(None)`.  Otherwise, returns
-/// `Ok(request)` for the parsed request.
+/// Append received chunks to the session's incoming buffer until the protocol
+/// buffer, serializing a request, is complete and can be deserialized.
+/// The first chunk of the series is prefixed with the protocol buffer's total
+/// length.
+/// If the incoming buffer reaches its expected length, then the protocol buffer
+/// is considered complete and parsed into a `RuntimeManagerRequest` before
+/// flushing the incoming buffer and returning `Ok(request)`.
+/// Otherwise, if we still need to receive more data in order to parse a full
+/// request, returns `Ok(None)`.
 ///
 /// TODO: harden this against potential malfeasance.  See the note, below.
 fn parse_incoming_buffer(
@@ -162,33 +167,52 @@ fn parse_incoming_buffer(
     let mut incoming_buffer_hash = INCOMING_BUFFER_HASH.lock()?;
 
     // First, make sure there is an entry in the hash for the TLS session.
-    if incoming_buffer_hash.get(&tls_session_id).is_none() {
-        incoming_buffer_hash.insert(tls_session_id, Vec::new());
+    // If not, this means we are receiving the first chunk of the protocol
+    // buffer.
+    let mut input = if incoming_buffer_hash.get(&tls_session_id).is_none() {
+        // Extract the protocol buffer's total length as u64
+        let input_data = input.split_off(8);
+        let mut expected_length: [u8; 8] = [0; 8];
+        expected_length.copy_from_slice(input.as_slice());
+        let expected_length = u64::from_be_bytes(expected_length);
+
+        // Insert the expected length in the hash table
+        incoming_buffer_hash.insert(tls_session_id, (expected_length, Vec::new()));
+
+        input_data
     }
+    else {
+        input
+    };
 
     // This should not panic, given the above.  If it does, something is wrong.
-    let incoming_buffer = incoming_buffer_hash.get_mut(&tls_session_id).ok_or(
+    let (expected_length, incoming_buffer) = incoming_buffer_hash.get_mut(&tls_session_id).ok_or(
         RuntimeManagerError::UnavailableIncomeBufferError(tls_session_id as u64),
     )?;
 
     incoming_buffer.append(&mut input);
 
-    // NB: `parse_from_bytes()` returning failure is interpreted as meaning the
-    // full protocol buffer has not yet been received. So we return `Ok(None)`
-    // with the hope that eventually we will receive all of it, and then
-    // `parse_from_bytes()` will return `Ok(parsed)`. In a well-behaving system,
-    // this is reasonable.  In a poorly-behaved system (under attack, clients
-    // just getting confused) it is not reasonable, and can eventually result in
-    // "Out of Memory" or Garbage out.
+    // We return `Ok(None)` as long as the full protocol buffer has not yet been
+    // received. Once received, `parse_from_bytes()` parses it and returns
+    // `Ok(parsed)`.
+    // In a well-behaving system, this is reasonable.  In a poorly-behaved
+    // system (under attack, clients just getting confused) it is not
+    // reasonable, and can eventually result in "Out of Memory" or Garbage out.
     //
     // TODO: It would be nice to check the error, and then determine if this might
     // be the case or if it is hopeless and we could just error out.
-    match protobuf::parse_from_bytes::<transport_protocol::RuntimeManagerRequest>(&incoming_buffer)
-    {
-        Err(_) => Ok(None),
-        Ok(parsed) => {
-            incoming_buffer_hash.remove(&tls_session_id);
-            Ok(Some(parsed))
+
+	if incoming_buffer.len() < *expected_length as usize {
+		return Ok(None)
+	}
+    else {
+        match protobuf::parse_from_bytes::<transport_protocol::RuntimeManagerRequest>(&incoming_buffer)
+        {
+            Err(_) => Ok(None),
+            Ok(parsed) => {
+                incoming_buffer_hash.remove(&tls_session_id);
+                Ok(Some(parsed))
+            }
         }
     }
 }
--- a/runtime-manager/src/managers/execution_engine_manager.rs
+++ b/runtime-manager/src/managers/execution_engine_manager.rs
@@ -174,21 +174,18 @@ fn parse_incoming_buffer(
     // First, make sure there is an entry in the hash for the TLS session.
     // If not, this means we are receiving the first chunk of the protocol
     // buffer.
-    let mut input = if incoming_buffer_hash.get(&tls_session_id).is_none() {
+    if incoming_buffer_hash.get(&tls_session_id).is_none() {
         // Extract the protocol buffer's total length as u64
-        let input_data = input.split_off(8);
-        let mut expected_length: [u8; 8] = [0; 8];
-        expected_length.copy_from_slice(input.as_slice());
-        let expected_length = u64::from_be_bytes(expected_length);
+        let remaining_data = input.split_off(8);
+        let mut expected_length_bytes: [u8; 8] = [0; 8];
+        expected_length_bytes.copy_from_slice(input.as_slice());
+        let expected_length = u64::from_be_bytes(expected_length_bytes);
 
         // Insert the expected length in the hash table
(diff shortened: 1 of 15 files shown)


runtime-manager/src/managers/execution_engine_manager.rs
```diff
@@ -172,33 +177,52 @@ fn parse_incoming_buffer(
...
     // First, make sure there is an entry in the hash for the TLS session.
-    if incoming_buffer_hash.get(&tls_session_id).is_none() {
-        incoming_buffer_hash.insert(tls_session_id, Vec::new());
+    // If not, this means we are receiving the first chunk of the protocol
+    // buffer.
+    let mut input = if incoming_buffer_hash.get(&tls_session_id).is_none() {
+        // Extract the protocol buffer's total length as u64
+        let input_data = input.split_off(8);
+        let mut expected_length: [u8; 8] = [0; 8];
+        expected_length.copy_from_slice(input.as_slice());
```
dominic-mulligan-arm, line 186: Should this be `input.as_slice()` or `input_data.as_slice()`?  Should `input_data` not be called `expected_length_bytes` or similar?

runtime-manager/src/managers/execution_engine_manager.rs
```diff
@@ -172,33 +177,52 @@ fn parse_incoming_buffer(
...
-    // `parse_from_bytes()` will return `Ok(parsed)`. In a well-behaving system,
-    // this is reasonable.  In a poorly-behaved system (under attack, clients
-    // just getting confused) it is not reasonable, and can eventually result in
-    // "Out of Memory" or Garbage out.
+    // We return `Ok(None)` as long as the full protocol buffer has not yet been
+    // received. Once received, `parse_from_bytes()` parses it and returns
+    // `Ok(parsed)`.
+    // In a well-behaving system, this is reasonable.  In a poorly-behaved
+    // system (under attack, clients just getting confused) it is not
+    // reasonable, and can eventually result in "Out of Memory" or Garbage out.
```
dominic-mulligan-arm, line 210: Surely we should never allow "garbage out" under any circumstances?  What can be done to stop this?

dominic-mulligan-arm (changes requested):
Thanks, this mostly looks good, just a few small comments about some things that look suspicious.  To confirm, you've run `rustfmt` on the changes using the Git hook, right?

runtime-manager/src/managers/execution_engine_manager.rs
gbryant-arm: After the call to `split_off()`, `input` is reduced to its first 8 bytes and `input_data` contains the rest of the vector.
Agreed that the variable names don't make that very clear. Also I'm doing something ugly with `input` getting assigned the return of an `if` block. I'll make some  changes

runtime-manager/src/managers/execution_engine_manager.rs
gbryant-arm: Agreed. We could enforce a maximum amount of memory per session/enclave as part of the policy, and even check beforehand if the file to be sent by the client can fit in the runtime manager

dominic-mulligan-arm: Ah OK, I misunderstood the behaviour of the `split_off` function: I thought it returned the first 8 bytes and modified the original vector in-place to point-to the resulting tail.

gbryant-arm:
I haven't run the code through `rustfmt` yet, will do

runtime-manager/src/managers/execution_engine_manager.rs
gbryant-arm: I'd really like a function like that but couldn't find it

gbryant-arm: I think adding a new policy field to specify the amount of memory available per session should be part of another PR, as it was already a problem before. I'm opening an issue for that

dominic-mulligan-arm: OK, that sounds like a good plan.

dominic-mulligan-arm (dismissed):

dominic-mulligan-arm:
The TZ failure seems spurious, but the Linux and IceCap failures seem to be a bug in the changeset: before calling `.split_at(8)` we probably need to check that the length of the incoming bytes is at least 8 elements wrong.  Though, is this now an invariant that every one of these messages is preceded by its encoded length?

dominic-mulligan-arm:
(In light of that, I'm going to unapproved this until the issue is fixed.)

gbryant-arm:
I noticed the tests failing.
There should be a way to enforce adding the length prefix to every protocol buffer, or at least every `RuntimeManagerRequest` message.

gbryant-arm:
It's been successfully integrated to the experimental benchmarking branch but it's on hold until I find time to spend on code quality

dominic-mulligan-arm:
What's the status of this PR, @gbryant-arm?  Are you going to come back to it, or is it now defunct?

dominic-mulligan-arm:
Converting this to a draft until it's one a fit state for review...

transport-protocol/src/custom.rs
```diff
@@ -23,35 +35,157 @@ pub enum TransportProtocolError {
...
+    BufferNotFound(u32),
+    #[error(
+        display = "TransportProtocol: Session {:?}: Can't get mutex for incoming buffer",
+        _0
+    )]
+    MutexError(u32),
 }
 type TransportProtocolResult = Result<std::vec::Vec<u8>, TransportProtocolError>;
 
+/// General documentation on the network stack (TODO: move it somewhere else)
```
dominic-mulligan-arm, line 58: Is this TODO in the issue tracker so it is not forgotten?

transport-protocol/src/custom.rs
```diff
@@ -23,35 +35,157 @@ pub enum TransportProtocolError {
...
+/// to parse the buffer, and returns `Ok(None)`.
+/// This technique significantly decreases message parsing time hence latency
+/// when dealing with large messages, e.g. files.
+
+/// Strip the length prefix from the input buffer.
+/// Return the length and the stripped buffer.
+/// This function must be called before deserializing a `protobuf` message
+fn get_length_prefix(buffer: &[u8]) -> (u64, &[u8]) {
+    let mut length_bytes: [u8; LENGTH_PREFIX_SIZE] = [0; LENGTH_PREFIX_SIZE];
+    length_bytes.copy_from_slice(&buffer[..LENGTH_PREFIX_SIZE]);
```
dominic-mulligan-arm, line 76: What if `buffer` is smaller than `LENGTH_PREFIX_SIZE`?  Should there be an explicit `assert` at the beginning of the function to check that precondition holds?  Though, as this is in the trusted runtime, shouldn't an explicit error be propagated rather than causing a runtime failure?

transport-protocol/src/custom.rs
```diff
@@ -23,35 +35,157 @@ pub enum TransportProtocolError {
...
+
+/// Strip the length prefix from the first chunk received and append subsequent
+/// chunks to the session's incoming buffer until the protocol buffer is
+/// complete and can be deserialized into a message.
+/// This function must be called everytime a new chunk is received.
+/// Take a session id and chunk as input.
+/// Return the complete buffer, or an error indicating that the buffer is
+/// partial, or any other error
+///
+/// TODO: harden this against potential malfeasance.  See the note, below.
```
dominic-mulligan-arm, line 98: Why is this not already done?

transport-protocol/src/custom.rs
```diff
@@ -23,35 +35,157 @@ pub enum TransportProtocolError {
...
+    };
+
+    // Append chunk to incoming buffer
+    incoming_buffer.append(&mut input.to_vec());
+
+    // We return `Ok(None)` as long as the full protocol buffer has not yet been
+    // received. Once received, we return `Ok(buffer)`.
+    // In a well-behaving system, this is reasonable.  In a poorly-behaved
+    // system (under attack, clients just getting confused) it is not
+    // reasonable, and can eventually result in "Out of Memory" or Garbage out.
```
dominic-mulligan-arm, line 135: Garbage out should never be possible, surely!

transport-protocol/src/custom.rs
```diff
@@ -23,35 +35,157 @@ pub enum TransportProtocolError {
...
+    //
+    // TODO: It would be nice to check the error, and then determine if this
```
dominic-mulligan-arm, line 137: Why are there so many TODOs left in this PR?

veracruz-server/src/veracruz_server_nitro.rs
```diff
@@ -286,7 +286,7 @@ pub mod veracruz_server_nitro {
 
         let body_vec =
             base64::decode(&received_body).map_err(|err| VeracruzServerError::Base64Decode(err))?;
-        let response = transport_protocol::parse_proxy_attestation_server_response(&body_vec)
+        let response = transport_protocol::parse_proxy_attestation_server_response(None, &body_vec)
```
dominic-mulligan-arm, line 289: When are any of these `Option` arguments to `parse_` functions something other than `None`?  If never, what's the point of the argument?

dominic-mulligan-arm (changes requested):

veracruz-server/src/veracruz_server_nitro.rs
gbryant-arm: When the caller has a session id. There's no session id in the tests, which defaults to 0

transport-protocol/src/custom.rs
gbryant-arm: All the TODOs except the one to move the network stack documentation already existed, I'm just maintaining the status quo

veracruz-server/src/veracruz_server_nitro.rs
dominic-mulligan-arm: But when does the caller ever actually have a session id?

transport-protocol/src/custom.rs
gbryant-arm: The client could saturate the enclave's memory by sending a large file. The amount of enclave memory should be specified in the policy file (#280).
I can't think of any other immediate vulnerabilities here.
Anyway, this PR isn't about hardening the network stack

transport-protocol/src/custom.rs
gbryant-arm: Done

transport-protocol/src/custom.rs
gbryant-arm: Cf. comment above

veracruz-server/src/veracruz_server_nitro.rs
gbryant-arm: The receiver uses the sender's session id if it exists, otherwise it defaults to 0

gbryant-arm: Examples:
https://github.com/veracruz-project/veracruz/pull/277/files#diff-9ed7046a7dd9d07f02d7decd3e701b3fac4def1894e61a6b23fe90ad6c6244e4R151
https://github.com/veracruz-project/veracruz/pull/277/files#diff-4ee6def7b8a49cbfb9f90ef51d0055b41f67727a78eedd55ef26b5f9677076eeR242
https://github.com/veracruz-project/veracruz/pull/277/files#diff-4ee6def7b8a49cbfb9f90ef51d0055b41f67727a78eedd55ef26b5f9677076eeR270

transport-protocol/src/custom.rs
gbryant-arm: Good catch. Working on a fix

gbryant-arm:
The benchmarks validated the changes. File transfer is somehow ~ 2s faster on vc-linux

transport-protocol/src/custom.rs
dominic-mulligan-arm: OK, fair enough!  Resolving this...

veracruz-server/src/veracruz_server_nitro.rs
dominic-mulligan-arm: Ok, great.

transport-protocol/src/custom.rs
gbryant-arm: Okay. Moving it to `get_length_prefix()`

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
+1 reached, merging.

PR #283, opened 2021-11-30
Test if we can call psa_crypto::init.

egrimley-arm:


egrimley-arm:
This is replaced by #346.

PR #284, opened 2021-11-30
DST

egrimley-arm:


mathias-arm:
This is going to be interesting to integrate with the future workspace. So far I have split my work into two steps:
- [First](https://github.com/mathias-arm/veracruz/compare/feature/1-runtime-manager-linux..feature/2-workspaces-linux) add workspaces
- [Second](https://github.com/mathias-arm/veracruz/compare/feature/2-workspaces-linux..feature/3-disable-tz-sgx) disable SGX/TZ and update Rust toolchain, remove as many vendored crates and version pins.

In the second part, I just commented out the `sgx` and `tz` features as well as their dependencies. This PR will have a lot of conflicts but that will be worth it.

egrimley-arm:
> * [Second](https://github.com/mathias-arm/veracruz/compare/feature/2-workspaces-linux..feature/3-disable-tz-sgx) disable SGX/TZ and update Rust toolchain, remove as many vendored crates and version pins.

That sounds like it could quite naturally be split into two stages, or even three?

I have a patch that let me build `freestanding-execution-engine` with `1.56.1`, changing `veracruz` and three other repos.

egrimley-arm:
Moved to #287.

PR #285, opened 2021-11-30
veracruz-utils/src/csr.rs: Generate certificate valid till 2031-02-01.

egrimley-arm:


geky:
Judging from the code comments, it looks like this is just a temporary solution until the root enclaves are removed.

Pushing back the date is probably fine, but maybe we should only push it back a year? With the assertion that that if we haven't fixed this next year we should replace it with something that generates certs relative to the current date?

egrimley-arm:
You think CI breaking in an obscure fashion is a good way to set ourselves a reminder to update the code? I would have thought that an issue database or a calendar entry would be a better way of doing that!

egrimley-arm:
So as not to lose sight of the checks running on this commit, I've also created a slightly modified PR #286.

dominic-mulligan-arm:
Is this a duplicate of #286?

egrimley-arm:
> Is this a duplicate of #286?

Yes! Closing ...

PR #287, opened 2021-12-01
Completely remove SGX and TrustZone.

egrimley-arm:


mathias-arm (dismissed):

dominic-mulligan-arm (approved):
Modulo those two small comments, this looks good to me (and a big simplification)!

PR #293, opened 2021-12-06
Remove Linux Root Enclave and reconfigure attestation process

dominic-mulligan-arm:


dominic-mulligan-arm:
This is ready for review now.

veracruz-utils/src/platform/vm.rs
```diff
@@ -57,24 +57,15 @@ pub enum RuntimeManagerMessage {
...
-    /// A request to initialize the Runtime Manager enclave with the provided
-    /// policy (for Linux).  Parameters (in order):
-    ///
-    /// - The JSON policy in String format,
-    /// - The bytes of the challenge,
-    /// - The challenge ID.
-    Initialize(String, Vec<u8>, i32),
+    /// A request to initialize the Runtime Manager enclave.
+    Initialize,
     #[cfg(feature = "linux")]
```
dreemkiller, line 62: I was thinking that by changing Linux to perform `Initialize` without a policy (matching with Nitro), we could make `nitro` and `linux` match better (removing the `#[cfg(feature =...)]`s that are interspersed throughout this file.

What prevented that for this change?

dominic-mulligan-arm: I think that can be changed, thinking more...

mathias-arm (approved):

veracruz-utils/src/platform/vm.rs
dominic-mulligan-arm: I've pushed changes to move the Linux attestation process closer to Nitro's.

veracruz-server/src/veracruz_server_linux.rs
```diff
@@ -12,109 +12,111 @@
...
+    /// to contact it.
+    const RUNTIME_ENCLAVE_SPAWN_DELAY: u64 = 2;
     /// IP address to use when communicating with the Runtime Manager enclave.
     const RUNTIME_MANAGER_ENCLAVE_ADDRESS: &'static str = "127.0.0.1";
-    /// Delay (in seconds) to use when spawning the Linux root enclave to
-    /// ensure that everything is started before proceeding with communication
-    /// between the server and enclave.
-    const LINUX_ROOT_ENCLAVE_SPAWN_DELAY_SECONDS: u64 = 2;
+    /// Port to communicate with the Runtime Manager enclave on.
+    const RUNTIME_MANAGER_ENCLAVE_PORT: &'static str = "6000";
```
ShaleXIONG, line 50: Maybe we need to document the default port somewhere, e.g. README.md; or even provide a configuration option in the policy file.

ShaleXIONG (approved):

ShaleXIONG (approved):

mathias-arm (approved):

dreemkiller (approved):

dominic-mulligan-arm:
Oops, was calling `init_session_manager` too late in the flow.  Should be fixed now...

ShaleXIONG (approved):

egrimley-arm:
Was the change to `third-party/rust-psa-crypto` in this PR just a mistake, caused by `git commit -a`?

It broke the IceCap build, I think.

dominic-mulligan-arm:
Yes, I think it was included by mistake and somehow made it into `main`.  I will revert.

PR #309, opened 2021-12-14
v1: Fix CLI quickstarts

geky:
These just fell a bit out of date due to VFS changes made since the CLI quickstarts were introduced.

The quickstarts should really be in CI, but this makes the most sense after changing the quickstarts to use a non-hardware platform (Linux?). This is less high-priority and can be done later.

I'm going to make one PR directly against the v1 branch, and another against main, in case there is a different merge pattern we want to use in this situation.

The only changes are to the files CLI_INSTRUCTIONS.md and CLI_QUICKSTART.md, which, since they're not in CI, should have no impact on existing tests or CI.

### DIFF
--- a/CLI_INSTRUCTIONS.markdown
+++ b/CLI_INSTRUCTIONS.markdown
@@ -23,7 +23,7 @@ rules, allowing you to maximize the time you have to get coffee while the code
 is compiling:
 
 ``` bash
-$ make sgx-cli-install sdk
+$ make sdk sgx-cli-install
 ...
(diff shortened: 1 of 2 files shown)


dominic-mulligan-arm:
Build failure looked spurious, with a failing `git fetch`, so I've restarted it.

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
+1 reached, merging.

PR #315, opened 2021-12-16
An experiment to see how much disc space we have.

egrimley-arm:


egrimley-arm:
I think the result of the experiment was 22 GiB: it was possible to `dd` 1 GiB 22 times.

(It doesn't seem to be possible to access the logs afterwards: you need to have been there, watching them as it's running.)

PR #312, opened 2021-12-15
Update toolchain.

egrimley-arm:
TODO:

- Update wasmtime and remove final reference to github.com/egrimley-arm

Draft log message for this PR:

Update Rust toolchain for Icecap and use stable compiler for rest.

Icecap now uses nightly-2021-08-25 and the rest uses stable 1.56.1.

In the new submodule icecap/sysroot/rust:

- Updates for internal API changes.
- A HACK: some locks are disabled.
- The submodule src/llvm-project is removed to save disc space.

In this repo changes include:

- Various features are removed to enable use of stable compiler.
- Crate versions updated, including: hex, lexical-core, x509-parser
- HACK to save disc space in icecap/Makefile: rm -rf ../sdk

PR #318, opened 2021-12-17
Try with "rm -rf ../sdk" to save disc space.

egrimley-arm:


egrimley-arm:
This was a draft for #312.

Issue #303, opened 2021-12-10
Improve coverage of strace feature


egrimley-arm:
See the places marked with `xx` in https://github.com/veracruz-project/veracruz/blob/main/execution-engine/src/wasi/strace.rs

PR #319, opened 2021-12-17
Implement "strace" output functions.

egrimley-arm:
Output is written to stderr so that nothing is lost if the process
exits unexpectedly. Currently not all types of argument are printed
in a useful way.

### DIFF
--- a/execution-engine/src/wasi/strace.rs
+++ b/execution-engine/src/wasi/strace.rs
@@ -9,32 +9,113 @@
 //! See the `LICENSE_MIT.markdown` file in the Veracruz root directory for
 //! information on licensing and copyright.
 
-#![allow(unused_variables)] // while most of the functions are unimplemented
-
 use super::common::MemoryHandler;
 use crate::fs::FileSystemResult;
 use std::fmt;
 
-pub struct Strace {}
+/// How many characters to display from a string or memory buffer.
+const BUFFER_DISPLAY_LEN: usize = 32;
+
+/// State of strace structure.
+enum TraceState {
+    /// Initial state: we have only seen the function name.
+    Func,
+    /// We have seen one or more arguments.
+    Args,
+    /// We have seen the result of the function all.
+    Done,
+}
+
+pub struct Strace {
+    enabled: bool,
+    state: TraceState,
+}
+
+// Convert a vector of bytes into a printable ASCII string.
+// The string may contain readable text, which is useful to display,
+// but it may also contain binary data, so we cannot decode it as UTF-8.
+fn strace_string(bytes: &[u8], max: usize) -> String {
+    let mut res = String::from("\"");
+    let n = if bytes.len() > max { max } else { bytes.len() };
+    for i in 0..n {
+        if 0x20 <= bytes[i] && bytes[i] < 0x7f {
+            if bytes[i] == b'\\' || bytes[i] == b'"' {
+                res.push_str("\\");
+            }
+            res.push_str(&String::from_utf8_lossy(&bytes[i..i + 1].to_vec()))
+        } else if bytes[i] == 9 {
+            res.push_str("\\t")
+        } else if bytes[i] == 10 {
+            res.push_str("\\n")
+        } else if i + 1 < n && b'0' <= bytes[i + 1] && bytes[i + 1] <= b'9' {
+            // The following character is a digit, so use three octal digits.
+            res.push_str(&format!("\\{:03o}", bytes[i]))
+        } else {
+            res.push_str(&format!("\\{:o}", bytes[i]))
+        }
+    }
+    res.push_str("\"");
+    if bytes.len() > max {
+        res.push_str("...")
+    }
+    res
+}
 
 impl Strace {
+    /// Start generating strace output, given function name.
     pub fn func(enabled: bool, name: &str) -> Self {
-        // NOT YET IMPLEMENTED
-        Strace {}
+        if enabled {
+            eprint!("{}(", name)
+        };
+        Strace {
+            enabled,
+            state: TraceState::Func,
+        }
     }
 
+    /// Common code for handling arguments: prints comma as required.
+    fn arg_print_comma(&mut self) {
+        match self.state {
+            TraceState::Func => self.state = TraceState::Args,
+            TraceState::Args => {
+                eprint!(", ");
+            }
+            TraceState::Done => eprint!("\nUnexpected strace arg: "),
+        }
+    }
+
+    /// Handle argument that is a memory buffer.
     pub fn arg_buffer<T: MemoryHandler>(&mut self, mem: &mut T, adr: u32, len: u32) {
-        // NOT YET IMPLEMENTED
+        if !self.enabled {
+            return;
+        }
+        self.arg_print_comma();
+        let mut bytes = vec![0u8; len as usize];
+        match mem.read_buffer(adr, &mut bytes) {
+            Ok(()) => eprint!("{}", strace_string(&bytes, BUFFER_DISPLAY_LEN)),
+            Err(_) => eprint!("BAD_MEM_REF"),
+        }
     }
 
+    /// Handle argument as decimal value.
     pub fn arg_dec<T: fmt::Display>(&mut self, n: T) {
-        // NOT YET IMPLEMENTED
+        if !self.enabled {
+            return;
+        }
+        self.arg_print_comma();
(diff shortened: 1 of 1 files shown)


execution-engine/src/wasi/strace.rs
```diff
@@ -9,30 +9,99 @@
...
 //! information on licensing and copyright.
 
-#![allow(unused_variables)] // while most of the functions are unimplemented
-
 use super::common::MemoryHandler;
 use crate::fs::FileSystemResult;
 use std::fmt;
 
-pub struct Strace {}
+enum TraceState {
```
dominic-mulligan-arm, line 16: Can you also add `rustdoc`-style comments to this enum and its fields explaining what they are?

execution-engine/src/wasi/strace.rs
```diff
@@ -9,30 +9,99 @@
...
+    Func,
+    Args,
+    Result,
+}
+
+pub struct Strace {
```
dominic-mulligan-arm, line 22: Can you add `rustdoc`-style comments to this struct and its fields, explaining what they are?

execution-engine/src/wasi/strace.rs
```diff
@@ -9,30 +9,99 @@
...
+    }
+    res.push_str("\"");
+    if bytes.len() > max {
+        res.push_str("...")
+    }
+    res
+}
 
 impl Strace {
     pub fn func(enabled: bool, name: &str) -> Self {
```
dominic-mulligan-arm, line 58: Can you add `rustdoc`-style comments to all of these functions?

execution-engine/src/wasi/strace.rs
```diff
@@ -9,30 +9,99 @@
...
     }
 
     pub fn arg_buffer<T: MemoryHandler>(&mut self, mem: &mut T, adr: u32, len: u32) {
-        // NOT YET IMPLEMENTED
+        if self.arg() {
+            return;
+        }
+        let mut bytes = vec![0u8; len as usize];
+        match mem.read_buffer(adr, &mut bytes) {
+            Ok(()) => eprint!("{}", strace_string(&bytes, 32)),
```
dominic-mulligan-arm, line 88: Where has 32 as a maximum length come from?  Moreover, if this is a constant can you float this out to the top of the file as a `const` declaration with a comment?

execution-engine/src/wasi/strace.rs
```diff
@@ -9,30 +9,99 @@
...
+    state: TraceState,
+}
+
+// Convert a vector of bytes into a printable ASCII string.
+// The string may contain readable text, which is useful to display,
+// but it may also contain binary data, so we cannot decode it as UTF-8.
+fn strace_string(bytes: &[u8], max: usize) -> String {
+    let mut res = String::from("\"");
+    let n = if bytes.len() > max { max } else { bytes.len() };
+    for i in 0..n {
```
dominic-mulligan-arm, line 33: I wonder whether this can be refactored to use the `windows` method on slices:

```rust
for &[current, next] in bytes.iter().windows(2) {
  if 0x20 <= current && current <= 0x7f {
    ...
  }
  ...
  } else if b'0' <= next && next <= b'9' {
    // The following character is a digit, so use three octal digits.
    res.push_str(&format!("\\{:03o}:, current);
  }
  ...
}
```

...or something similar.

execution-engine/src/wasi/strace.rs
```diff
@@ -9,30 +9,99 @@
...
+        if enabled {
+            eprint!("{}(", name)
+        };
+        Strace {
+            enabled,
+            state: TraceState::Func,
+        }
+    }
+
+    fn arg(&mut self) -> bool {
```
dominic-mulligan-arm, line 68: This `returns` true when an argument is *not* expected, given the current state of the `Strace`, right?  I wonder if renaming this function to something more suggestive (e.g., `is_arg_expected()`) and introducing a special-purpose two-element type

```rust
enum ArgExpected {
  Expected,
  NotExpected
}
```

may not be a good idea, here?

All the checks at the beginning of functions, of the form:

```rust
if self.arg() {
    return;
}
```

then become something like:

```rust
if self.is_arg_expected().is_not_expected() {
  return;
}
```

where `is_not_expected()` is defined on `ArgExpected` in the obvious way...

execution-engine/src/wasi/strace.rs
```diff
@@ -42,23 +111,42 @@ impl Strace {
...
+        }
         // NOT YET IMPLEMENTED
+        eprint!("DIRENTS")
     }
 
     pub fn arg_events<T: MemoryHandler>(&mut self, _mem: &mut T, _events: u32, _size: u32) {
+        if self.arg() {
+            return;
+        }
         // NOT YET IMPLEMENTED
```
dominic-mulligan-arm, line 125: Are these going to be implemented before the PR is merged?

execution-engine/src/wasi/strace.rs
```diff
@@ -69,23 +157,69 @@ impl Strace {
...
+            return;
+        }
+        if !res.is_ok() {
+            eprint!("_");
+            return;
+        }
+        if let Ok(len) = memory_ref.read_u32(address) {
+            // This inefficiently copies everything, but it's only used for tracing.
+            if let Ok(bufs) = memory_ref.unpack_iovec(base, count) {
+                let mut buf: Vec<u8> = Vec::new();
```
dominic-mulligan-arm, line 170: I think this can be rewritten as `bufs.iter().flatten().collect()`.

execution-engine/src/wasi/strace.rs
egrimley-arm: But how would I handle the last character of the slice?

execution-engine/src/wasi/strace.rs
egrimley-arm: Done.

execution-engine/src/wasi/strace.rs
egrimley-arm: Done.

execution-engine/src/wasi/strace.rs
egrimley-arm:
```
/// How many characters to display from a string or memory buffer.
const BUFFER_DISPLAY_LEN: usize = 32;
```

execution-engine/src/wasi/strace.rs
egrimley-arm: This was a hack to make the code more concise. I've separated the function from the test.

execution-engine/src/wasi/strace.rs
egrimley-arm: I couldn't get that to work.

execution-engine/src/wasi/strace.rs
egrimley-arm: Probably not?

execution-engine/src/wasi/strace.rs
dominic-mulligan-arm: What's the error?

execution-engine/src/wasi/strace.rs
egrimley-arm: Done.

execution-engine/src/wasi/strace.rs
egrimley-arm:
```
193 |                 let x = bufs.iter();
    |                              ^^^^ method not found in `IoVecSlices<'_, common::Bound<'_, <T as MemoryHandler>::Slice>>`
```

dominic-mulligan-arm: I think you need an `as_ref()` on `bufs`.

egrimley-arm:
```
193 |                 let x = bufs.as_ref().iter().flatten();
    |                                              ^^^^^^^ `&&[u8]` is not an iterator
```

dominic-mulligan-arm: Hmm, OK, I think we're beyond the point of marginal return trying to get this to work, lol.

ShaleXIONG (approved):

execution-engine/src/wasi/strace.rs
dominic-mulligan-arm: But are they tracked in the issue tracker, at least?

egrimley-arm: Yes, it's #303.

PR #322, opened 2021-12-21
Switch to unmodified "byteorder".

egrimley-arm:
You know how I said I prefer small PRs, well this one is really small ...

### DIFF
--- a/execution-engine/Cargo.toml
+++ b/execution-engine/Cargo.toml
@@ -20,7 +20,7 @@ std = [
 
 [dependencies]
 anyhow = { version = "1.0", default-features = false }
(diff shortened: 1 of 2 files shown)


mathias-arm:
This changes from version `1.3.4` to `1.4.3` (which is the latest). Can you also upgrade the dependency in `io-utils`?

egrimley-arm:
> This changes from version `1.3.4` to `1.4.3` (which is the latest). Can you also upgrade the dependency in `io-utils`?

Yes, I missed that. New version just pushed.

mathias-arm (approved):

PR #325, opened 2021-12-31
Enable SIMD proposal in wasmtime

gbryant-arm:
Enable the SIMD proposal in wasmtime so that wasmtime can parse WASM bytecode containing SIMD instructions and output native SIMD instructions.
This has the benefit of improving performance. It brings the overhead of the video object detection algo from 4x down to 1.4x compared to native.
The default engine configuration is overriden with `wasm_simd` set to true.


dominic-mulligan-arm (approved):

gbryant-arm:
The code was too complicated, I've made a small change to remedy that

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
+1 reached, merging.

PR #326, opened 2022-01-04
Remove old files in sysroot-install target of icecap/mk/realm.mk.

egrimley-arm:
Using a nasty bit of inline shell.

Also update icecap/sysroot/workspace/Cargo.lock, which failed to
take into account a last-minute change to corresponding Cargo.toml.


mathias-arm (approved):
I encountered the same Cargo.lock issue

PR #328, opened 2022-01-06
rust-examples/read-file: Handle Result<DirEntry> being Err(_).

egrimley-arm:


egrimley-arm:
I don't know whether we need to handle `Result<DirEntry>` being `Err(_)`. It seems to happen if you have `FD_READ` without `FD_READDIR`, or something like that. Whether this is correct behaviour by the VFS is hard to say when the WASI spec is so vague.

PR #329, opened 2022-01-10
Use workspaces for CI

mathias-arm:
Currently a draft to get some feedback. Initially I created a single CI image that would work for `linux`, `nitro` and `icecap`. I ended up running out of space for `icecap` build. Instead it now runs with 3 separate dedicated images. I remove the `check-repo-and-compile-sdk ` job because it did not seem to save any wallclock time.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
   linux:
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/veracruz-project/veracruz/veracruz:ci-v2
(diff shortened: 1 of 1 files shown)


egrimley-arm (approved):
It's not ideal how the `.nix` files have been moved away from the code they build, but I don't think that's a reason for delaying this. We really don't want to continue for too long with two different build infrastructures in use.

ShaleXIONG (approved):
Several MR seems to change this file repeatedly?

egrimley-arm:
> Several MR seems to change this file repeatedly?

Not sure which file you mean?

Mathias, do you want to merge this now?

mathias-arm:
> Mathias, do you want to merge this now?

Yes

PR #332, opened 2022-01-12
pre-commit: Run rustfmt without "--skip-children --unstable-features".

egrimley-arm:
Now we are using a stable compiler people are likely to have the
stable rustfmt, which does not have "--skip-children". Also we have
applied rustfmt uniformly across the codebase so it should not hurt to
apply rustfmt more broadly.

This patch also adds code to check that rustfmt ran correctly. The
problem with rustfmt failing was previously easy to ignore.

### DIFF
--- a/githooks/pre-commit
+++ b/githooks/pre-commit
@@ -15,13 +15,17 @@
 # See the `LICENSE_MIT.markdown` file in the Veracruz repository root
 # directory for copyright and licensing information.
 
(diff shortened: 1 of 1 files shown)


dominic-mulligan-arm (approved):

dominic-mulligan-arm:
Thanks for fixing this: I noticed the issue with `--skip-children` when pushing changes earlier today.

+1 reached, merging.

PR #333, opened 2022-01-19
Replace psa-attestation/lib/mbed-crypto with rust-psa-crypto

egrimley-arm:


egrimley-arm:
This is replaced by #346.

PR #334, opened 2022-01-20
Remove unused duplicates: host.nix realm.nix sdk-and-test-collateral.nix

egrimley-arm:
There remain some duplicates where both copies seem to be used:
```
config.nix
default.nix
instance.nix
targets.nix
```

This will be tidied up at some point.


egrimley-arm:
Another `error: writing to file: No space left on device`.

egrimley-arm:
One of the deleted files is under `workspaces`, though.

mathias-arm:
I don't see an issue with removing the unused duplicates. The `FIXME` changes seem superfluous as we are gearing towards removing the non-workspace build system, I would prefer removing the files in question from `icecap/` and leave untouched the ones in `workspace/`.

egrimley-arm:
I think I've worked out how to remove those files from `icecap/`. I hope that a new version of this PR will arrive soon.

egrimley-arm:
This is replaced by #354.

PR #336, opened 2022-01-21
Build Wasm programs with PROFILE=release.

egrimley-arm:
This might make CI faster.


mathias-arm:
The part  you removed from `workspaces/common.mk` is needed.

egrimley-arm:
> The part you removed from `workspaces/common.mk` is needed.

I can vaguely remember removing those lines, but I can't now remember why. I've restored them (and CI failed for a probably unrelated reason).

egrimley-arm:
I think Wasm programs are now being optimised well enough without this change.

PR #339, opened 2022-02-02
Fix runtime manager's Makefiles across workspaces

gbryant-arm:
Currently the build system doesn't recompile the runtime manager when its dependencies, e.g. `transport-protocol`, are modified.
This makes debugging tedious and error prone: many times I was trying to figure out the reason for network issues or crashes, before realising the binaries had not been recompiled.  @egrimley-arm faced similar issues.
This PR fixes that by calling Cargo to look for changes in the source of  runtime-manager and all its dependencies, instead of relying on traditional Makefile targets.

### DIFF
--- a/workspaces/linux-runtime/Makefile
+++ b/workspaces/linux-runtime/Makefile
@@ -19,14 +19,10 @@ include $(WORKSPACE_DIR)/common.mk
 
 all: linux
 
-linux: runtime-manager-enclave css-linux.bin
+linux: runtime-manager-enclave
 
 runtime-manager-enclave:
 	cargo build $(PROFILE_FLAG) $(V_FLAG) --features linux -p runtime_manager_enclave
-	cp target/$(PROFILE_PATH)/runtime_manager_enclave .
-
-css-linux.bin: runtime-manager-enclave
-	cp target/$(PROFILE_PATH)/runtime_manager_enclave $@
 
 doc:
 	cargo doc
@@ -36,7 +32,6 @@ fmt:
 
 clean:
 	@cargo clean
-	@rm -f css-linux.bin
 
 clean-cargo-lock:
 	rm -f Cargo.lock
--- a/workspaces/linux-runtime/Makefile
+++ b/workspaces/linux-runtime/Makefile
@@ -25,7 +25,7 @@ runtime-manager-enclave:
 	cargo build $(PROFILE_FLAG) $(V_FLAG) --features linux -p runtime_manager_enclave
 
 css-linux.bin: runtime-manager-enclave
-	cp $< $@
+	cp target/$(PROFILE_PATH)/runtime_manager_enclave $@
 
 doc:
 	cargo doc
--- a/workspaces/linux-runtime/Makefile
+++ b/workspaces/linux-runtime/Makefile
@@ -23,6 +23,7 @@ linux: runtime-manager-enclave css-linux.bin
 
 runtime-manager-enclave:
 	cargo build $(PROFILE_FLAG) $(V_FLAG) --features linux -p runtime_manager_enclave
+	cp target/$(PROFILE_PATH)/runtime_manager_enclave .
 
 css-linux.bin: runtime-manager-enclave
 	cp target/$(PROFILE_PATH)/runtime_manager_enclave $@
--- a/workspaces/linux-runtime/Makefile
+++ b/workspaces/linux-runtime/Makefile
@@ -9,7 +9,7 @@
 # See the `LICENSE_MIT.markdown` file in the Veracruz root director for licensing
 # and copyright information.
 
-.PHONY: all clean clean-cargo-lock doc fmt linux
+.PHONY: all clean clean-cargo-lock doc fmt linux runtime-manager-enclave
 
 default: all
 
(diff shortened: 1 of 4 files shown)


runtime-manager/Makefile
```diff
@@ -39,9 +39,10 @@ css-linux.bin: target/release/runtime_manager_enclave
 
 
 ############# AWS Nitro Enclaves ###################
-Nitro_Src = $(COMMON_Src) src/runtime_manager_nitro.rs src/main.rs
+nitro: build-nitro runtime_manager.eif
 
-nitro: runtime_manager.eif
+build-nitro:
```
egrimley-arm, line 44: You should add `build-nitro` to the `.PHONY` dependencies.

dreemkiller:
Cargo, even when it's doing nothing, it a relatively slow process (I think 0.5s to check dependencies).
Make is much faster. Is it possible to modify the Make solution to pick up the relevant changes instead? If not, this will work.

gbryant-arm:
Agreed, Cargo is slow, but mainly at compiling, which we can't avoid, right?
Plus it automatically keeps track of the sources, fetches updates for the crate's dependencies, etc.
I don't think using Cargo here will add much overhead to the build process

gbryant-arm:
AFAIK the RM is the only component of the build (or at least the only one I had issues with) that is currently relying on Make to detect changes, so that would be 0.5 additional seconds per build, which seems acceptable.
The main Makefile directly calls cargo for every other component, and I don't think we want to change that, right?

mathias-arm (changes requested):
This `Makefile` is destined to be removed. Look at the ones in `workspaces/linux-runtime` and `workspaces/nitro-runtime`

workspaces/linux-runtime/Makefile
```diff
@@ -9,30 +9,25 @@
...
-target/$(PROFILE_PATH)/runtime_manager_enclave target/$(PROFILE_PATH)/linux-root-enclave: \
-	Cargo.toml $(Linux_Src) crates/linux-root-enclave/src/main.rs
-	cargo build $(PROFILE_FLAG) --features linux \
-		-p runtime_manager_enclave -p linux-root-enclave
+linux-root-enclave:
+	cargo build $(PROFILE_FLAG) --features linux -p linux-root-enclave
 
-css-linux.bin: target/$(PROFILE_PATH)/runtime_manager_enclave
+css-linux.bin: runtime-manager-enclave
 	cp $< $@
```
mathias-arm, line 31: This does not seem to be right. 
`runtime-manager-enclave` rule builds `target/$(PROFILE_PATH)/runtime_manager_enclave `, and it's the latter that should be copied to `css-linux.bin`

mathias-arm:
@gbryant-arm: regarding @dreemkiller's comments, have you checked that with multiple `make` invocations do not result in superfluous rebuilds (i.e. `cargo` figures out there is no need to recompile).

dominic-mulligan-arm:
What is the status of this PR @gbryant-arm?

gbryant-arm:
WIP

dominic-mulligan-arm:
Converted the PR to a draft, in that case.

mathias-arm:
Looking at the nitro changes, it seems `nitro-cli` might be run even without any changes to the `runtime-manager` which brings me closer to @dreemkiller's concerns. Looking back at the original issue, what kind of changes were not picked up?

mathias-arm:
Looking at the nitro changes, it seems `nitro-cli` might be run even without any changes to the `runtime-manager` which brings me closer to @dreemkiller's concerns. Looking back at the original issue, what kind of changes were not picked up?

gbryant-arm:
> Looking back at the original issue, what kind of changes were not picked up?

For example, `runtime manager` doesn't get recompiled when `transport protocol` is changed.


> nitro-cli might be run even without any changes to the runtime-manager

I can't find `nitro-cli` anywhere. Does that target still exist in the workspace Makefiles?

gbryant-arm:
> @gbryant-arm: regarding @dreemkiller's comments, have you checked that with multiple make invocations do not result in superfluous rebuilds (i.e. cargo figures out there is no need to recompile).

Checked. Takes cargo about 0.2s to look for changes and realize there's nothing to build.

mathias-arm:
For the nitro part, it makes sense to re-run `cargo build` each time. I am not sure what now happens with the rules for `runtime_manager.eif` and `PCR0`, does `nitro-cli` get called appropriately?

gbryant-arm:
> `runtime_manager.eif` and `PCR0`, does `nitro-cli`

`nitro-cli` gets called whenever the `runtime_manager.eif` or (and?) `PCR0` targets get executed. The CI logs confirm that

mathias-arm:
> > `runtime_manager.eif` and `PCR0`, does `nitro-cli`
> 
> `nitro-cli` gets called whenever the `runtime_manager.eif` or (and?) `PCR0` targets get executed. The CI logs confirm that

The issue I was thinking about is on subsequent builds, if those rules are called regardless of whether `runtime_manager_enclave` was updated.

mathias-arm:
> One hacky way to tell is to check the creation date of the produced binaries. Ugly

This is one of the main feature of Make.

gbryant-arm:
> This is one of the main features of Make.

It is, but I don't think we can use that here, since the target is already used. We'd have a cycle of targets

gbryant-arm:
The goal of this PR is to replace Make with cargo to track Rust-related dependencies

gbryant-arm:
@mathias-arm Are you happy with this PR? I addressed your request for changes.

dominic-mulligan-arm:
What is the status of this PR?

gbryant-arm:
> What is the status of this PR?
Waiting for @mathias-arm 's approval. It needs rebasing too

dominic-mulligan-arm:
Can you go ahead and rebase in the meantime?  @mathias-arm are you happy with this PR?  @gbryant-arm claims to have addressed all concerns.

mathias-arm:
#339 and #375 are co-dependent because the second chunk of [b1939e4](https://github.com/veracruz-project/veracruz/pull/375/commits/b1939e4c19200e7b9251e13d3ff06730ec90ac52) relies on [0ac34f3](https://github.com/veracruz-project/veracruz/pull/339/commits/0ac34f3368039d3b8a1788feb81b6db205d8c86c). I think both should be removed from their respective PR.

dominic-mulligan-arm:
OK, trying to break the deadlock because this has been dragging on for way too long:

- @mathias-arm: can you be more specific about what's wrong with 14d5006?
- @gbryant-arm: please combine PRs #339 and #375 into a single PR.  Also, can you be more specific about why this change is needed, and why the same effect cannot be obtained using `make`?  That's not clear at all from the thread above, just a comment about "circular targets".  Isn't one of the main features of `make` to detect changes in files and act on them?

mathias-arm:
Regarding https://github.com/veracruz-project/veracruz/commit/14d5006060d6763042dcb1acc1de62c0157e0878 it's a commit titled `Fix Makefile` that deletes the `docker` directory. I assume this is a mistake which happened because `docker` switched from being a submodule to a regular directory.

My take on @gbryant-arm's change is that in order to be sure things are properly rebuilt, we switch from expressing dependencies on the existence of files to phony targets that will force cargo invocations (which should be quick hopefully). Even if @dreemkiller and I have expressed concern about the consequence on build times, the principle is sound. My remaining objections is that the two PRs rely on `$(WORKSPACE_DIR)/linux-runtime/target/$(PROFILE_PATH)/runtime_manager_enclave` being copied to `$(WORKSPACE_DIR)/linux-runtime/runtime_manager_enclave` which I think is superfluous and confusing.

gbryant-arm:
I'm fixing https://github.com/veracruz-project/veracruz/issues/383 for Linux and Nitro as well as this is related

gbryant-arm:
Okay ready for review

mathias-arm (approved):

PR #341, opened 2022-02-02
Veracruz Server Nitro: Remove 10s sleep after launching the runtime manager

gbryant-arm:
There's a 10s wait after launching the runtime manager, which has been shown empirically on hundreds of runs to not make any difference.

NB: There's also a 2s sleep in Veracruz Server Linux (https://github.com/veracruz-project/veracruz/blob/main/veracruz-server/src/veracruz_server_linux.rs#L50).
We could get rid of that along with the Linux Root Enclave (PR https://github.com/veracruz-project/veracruz/pull/293)

### DIFF
--- a/veracruz-server/src/veracruz_server_nitro.rs
+++ b/veracruz-server/src/veracruz_server_nitro.rs
@@ -20,6 +20,10 @@ pub mod veracruz_server_nitro {
     use std::env;
     use veracruz_utils::platform::vm::{RuntimeManagerMessage, VMStatus};
 
+    /// Delay to apply between launching the server and trying to contact the runtime manager
+    /// enclave (in milliseconds).
+    const SERVER_DELAY_IN_MILLIS: u64 = 2000;
+    /// Path of the Runtime Manager enclave EIF file.
     const RUNTIME_MANAGER_EIF_PATH: &str = "../runtime-manager/runtime_manager.eif";
     /// The protocol to use when interacting with the proxy attestation server.
     const PROXY_ATTESTATION_PROTOCOL: &str = "nitro";
@@ -69,7 +73,8 @@ pub mod veracruz_server_nitro {
                 enclave: runtime_manager_enclave,
             };
             println!("VeracruzServerNitro::new Runtime Manager instantiated. Calling initialize");
-            std::thread::sleep(std::time::Duration::from_millis(10000));
+
+            std::thread::sleep(std::time::Duration::from_millis(SERVER_DELAY_IN_MILLIS));
 
             let attesstation_doc = {
                 let attestation = RuntimeManagerMessage::Attestation(challenge, challenge_id);


dreemkiller:
What platforms has removing this shown to be working?
It was added for SGX, which is deprecated.
I would think startup time for Nitro would be significant (it is booting Linux, after all).

dreemkiller:
Nevermind, this is the Nitro-specific code. Duh. 
Anyway, I'm still surprised. 10 seconds is probably too long, but some delay seems like it would be necessary.

egrimley-arm:
If you _had_ to implement this without any arbitrary delays, what would you do? (I'm assuming "keep sending requests" involves an arbitrary delay between requests so that's not a "proper" solution either.)

gbryant-arm:
> I'm assuming "keep sending requests" involves an arbitrary delay between requests so that's not a "proper" solution either

Agreed, that's not a proper solution but that's safer than trying to establish a connection once and for all.

> If you had to implement this without any arbitrary delays, what would you do?

Ideally the connection should be established in a blocking fashion: the delivery of the first message is pending until the RM picks it up.
Or the RM should somehow signal that it's ready to communicate.
If that's not straightforward to implement then I think it's acceptable to wait for e.g. 100ms, that's not much compared to the overall onboarding time (about 3s on Nitro excluding provisioning and the 10s wait).

egrimley-arm:
I think probably in general the command to start a server should not exit until the server really is started, i.e. ready to accept connections. I think I did this once with a program that wrote a `'\n'` when it was ready and a test script that did something like `./server | head -n 1`. And on another occasion I ran someone else's server with some kind of logging enabled and had the test script grep the log for a message that implied the server was ready.

dominic-mulligan-arm:
@dreemkiller are you happy with this change?  I'm also surprised that *no* delay is required at all and wonder how representative testing on one machine actually is, in practice.  I think if we're going to remove the delay, then there should be some robust method implemented to signal that the server is up-and-running to the client...

dreemkiller:
I'm surprised that it works.

Ideally, yes, we would have a mechanism to message from the enclave to the non-secure server when it is up. Barring that, if removal of the delay doesn't cause problems, then I'm fine taking it out.

dominic-mulligan-arm:
On further reflection, I don't believe that completely removing the delay is robust enough.  I think we need to keep a delay, albeit the 10 seconds can probably be reduced to 2-3 seconds or so.

veracruz-server/src/veracruz_server_nitro.rs
```diff
@@ -68,7 +68,6 @@ pub mod veracruz_server_nitro {
                 enclave: runtime_manager_enclave,
             };
             println!("VeracruzServerNitro::new Runtime Manager instantiated. Calling initialize");
-            std::thread::sleep(std::time::Duration::from_millis(10000));
```
dominic-mulligan-arm, line 71: Can you keep this delay, albeit reduce it to 2-3 seconds?

dominic-mulligan-arm (changes requested):

veracruz-server/src/veracruz_server_nitro.rs
gbryant-arm: Sounds like the easiest and most sensible thing to do

dreemkiller (approved):

PR #342, opened 2022-02-04
icecap: Bump IceCap source

egrimley-arm:
This is an updated version of https://github.com/veracruz-project/veracruz/pull/338

PR #343, opened 2022-02-08
icecap: Bump IceCap source, plus Use workspaces for CI

egrimley-arm:
This is #342 with #329 on top.


egrimley-arm:
This was just https://github.com/veracruz-project/veracruz/pull/342 with https://github.com/veracruz-project/veracruz/pull/329 on top.

PR #344, opened 2022-02-24
Fix standard streams

gbryant-arm:
Fix a bug in the way standard streams are handled: as of now, a forward slashed is enforced to every file, making it impossible to use `stdin`, `stdout` and `stderr`


egrimley-arm:
CI fails because of `nsm` change, which I am working on.

ShaleXIONG (approved):

policy-utils/src/parsers.rs
```diff
@@ -67,7 +67,13 @@ pub fn parse_renamable_paths(
 /// mistakes.
 #[cfg(feature = "std")]
 pub fn enforce_leading_backslash(path: &str) -> Cow<str> {
-    if !path.starts_with('/') {
+    let is_special_file = match path {
+        "stdin" => true,
```
dominic-mulligan-arm, line 71: Is this not clearer as

```rust
let is_special_file = path == "stdin" || path == "stdout" || path == "stderr";
```

policy-utils/src/parsers.rs
```diff
@@ -67,7 +67,13 @@ pub fn parse_renamable_paths(
...
+        "stderr" => true,
```
dominic-mulligan-arm, line 72: Also, I wonder whether these "special file" names should be constants, rather than matching on string literals (and potentially introducing typos)?  e.g.

```rust
let is_special_file = path == STDIN_FILE_PATH || path == STDOUT_FILE_PATH || path == STDERR_FILE_PATH;
```

runtime-manager/src/managers/mod.rs
```diff
@@ -130,9 +130,13 @@ impl ProtocolState {
                 }
             }
         }
-        self.vfs
-            .spawn(client_id)?
-            .write_file_by_absolute_path(file_name, data, false)?;
+        let mut vfs = self.vfs.spawn(client_id)?;
+        match file_name {
+            "stdin" => {
```
dominic-mulligan-arm, line 135: Is this not clearer as an if-else?

```rust
if file_name == STDIN_FILE_PATH {
  self.vfs.spawn(client_id)?.write_stdin(&data)?;
} else {
  self.vfs.spawn(client_id)?.write_file_by_absolute_path(file_name, data, false)?;
}
```

dominic-mulligan-arm:
@ShaleXIONG can you take a look at this?  I pushed changes yesterday so cannot approve myself.

ShaleXIONG (approved):

dominic-mulligan-arm:
+1 reached, merging.

PR #346, opened 2022-02-25
Replace psa-attestation/lib/mbed-crypto with mbedtls from rust-psa-crypto.

egrimley-arm:
This is on top of two other PRs that I expect to be merged first: #345, #329

The crate rust-psa-crypto is slightly modified from upstream, primarily to configure mbedtls differently: "baremetal" with entropy obtained with the C-library function getrandom.

Also, rust-psa-crypto/Cargo.toml is removed to avoid nested workspaces.

This uses a different version of the C library with a slightly different API. In particular, key handles are now uint32_t instead of uint16_t. This accounts for most of the changes in psa-attestation and t_cose.

There are various IceCap changes to enable building rust-psa-crypto. These include adding a bogus implementation of getrandom()!


mathias-arm:
Looking at `rust-psa-crypto`, it might be interesting to take inspiration from `psa-crypto-sys` to improve `psa-attestation` build.

mathias-arm:
Instead of building `QCBOR` and `t_cose` libraries separately and having to hack their build system, why not compile their sources as part of the `psa-attestation` static C library?

psa-attestation/Cargo.toml
```diff
@@ -18,8 +18,11 @@ nitro = []
 
 [dependencies]
 libc = "0.2.69"
+# We are not really using psa-crypto but we are using the C libraries
+# libmbedcrypto.a and libshim.a that psa-crypto indirectly uses.
+psa-crypto = { path = "../third-party/rust-psa-crypto/psa-crypto" }
 
 [build-dependencies]
-bindgen = "0.59"
+bindgen = "0.57.0"
```
mathias-arm, line 26: The `icecap-sel4-sys` crate uses `bindgen = "*"` (which is currently `0.59.2`) should we all align on 0.57 or 0.59?

egrimley-arm: `0.57` is what I've tested!

psa-attestation/Cargo.toml
```diff
@@ -18,8 +18,11 @@ nitro = []
 
 [dependencies]
 libc = "0.2.69"
+# We are not really using psa-crypto but we are using the C libraries
+# libmbedcrypto.a and libshim.a that psa-crypto indirectly uses.
+psa-crypto = { path = "../third-party/rust-psa-crypto/psa-crypto" }
```
mathias-arm, line 23: Given the comment should the dependency rather be on `psa-crypto-sys`?

egrimley-arm:
> Instead of building `QCBOR` and `t_cose` libraries separately and having to hack their build system, why not compile their sources as part of the `psa-attestation` static C library?

Would that have to be part of this PR?

mathias-arm:
> Would that have to be part of this PR?

It is not a requirement. Some of the changes in our `t_cose` fork and all of the ones for `QCBOR` are related to their build systems. The suggested change will minimize the difference to upstream and avoid doing double commits (one in the submodule and one in `veracruz`) for build-related changes.

psa-attestation/Cargo.toml
egrimley-arm: I've changed that to `0.59.2`.

psa-attestation/Cargo.toml
egrimley-arm: It probably doesn't make any difference. I'll try changing it.

PR #347, opened 2022-02-25
Test of: Replace psa-attestation/lib/mbed-crypto with mbedtls from rust-psa-crypto.

egrimley-arm:
This is #346 with some additional tests on top.


egrimley-arm:
This was a hacked together test of #346, which is now merged, so I'm going to close the PR now to declutter things a bit.

I hope that we will have a proper test of the `psa-crypto` functionality before too long.

Issue #349, opened 2022-03-02
There's recently been a regression on Nitro


dreemkiller:
**Describe the bug**
When veracruz-server-test is running on a nitro-enabled, platform, we see the following error:
`VeracruzServerNitro::new complete. Returning
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Unspecified', /usr/local/cargo/git/checkouts/rustls-59aebdcc975fb2c1/39f3996/src/rand.rs:14:10`

**To Reproduce**
Start with a fresh environment, and execute:
`make nitro-veracruz-server-test`
Most of the tests will fail.

When you run one test by itself, for example `test_phase3_linear_regression_one_data_with_attestation`, you will see the above error message

**Expected behaviour**
Tests should pass

**Environment, Git commit hash, and configuration**
Working from the docker file from the repo. Executing inside the container as root.

mathias-arm:
This might not fix the issue bug you should be using:
```
make -C `workspaces/nitro-host` test-server
```

dreemkiller:
Have the READMEs been updated to reflect that? I haven't seen that on main.

egrimley-arm:
The `nitro-veracruz-server-test` target was removed on 4 Mar 2022 and `NITRO_INSTRUCTIONS.markdown` has been recently updated (#481) and tested. So I'm closing this issue.

(I'll just mention in passing that the instructions specify using `PROFILE=release`. I think it doesn't work without that, so perhaps that should be investigated or the default should be changed.)

PR #350, opened 2022-03-02
WIP: Added a new feature for veracruz-utils to specify it should be built …

dreemkiller:
…for inside a nitro enclave

This is a fix for #349.

The issue was that for nitro platforms, even when outside of the enclave, the `ring` crate was being built to use the AWS nitro `nsm_lib` crate for random number generation. This is only supported inside an enclave.

Added an additional feature to `veracruz-utils`, `nitro_enclave` to clarify that it is being built for execution inside the enclave.

This is certainly one of those bugs that makes you think "How did this ever work". I suppose, but have not investigated, that it has something to do with how `cargo` resolved dependencies.

### DIFF
--- a/runtime-manager/Cargo.toml
+++ b/runtime-manager/Cargo.toml
@@ -58,6 +58,7 @@ nitro = [
   "ring/nitro",
   "ring/std",
   "session-manager/nitro",
+  "veracruz-utils/nitro_enclave",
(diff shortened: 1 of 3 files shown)


mathias-arm:
The version "patch" to point to our git forks is done in the workspace `Cargo.toml` file.

PR #351, opened 2022-03-03
Change bindgen version from 0.59.2 to 0.57.

egrimley-arm:
The toolchain version currently used for the IceCap build does not
accept the code generated by bindgen 0.59.2.

### DIFF
--- a/psa-attestation/Cargo.toml
+++ b/psa-attestation/Cargo.toml
@@ -23,6 +23,6 @@ libc = "0.2.69"
 psa-crypto-sys = { path = "../third-party/rust-psa-crypto/psa-crypto-sys" }
 
 [build-dependencies]
-bindgen = "0.59.2"
+bindgen = "0.57"
 cfg-if = "1.0"
 cmake = "0.1"


egrimley-arm:
I'm now not sure that this change is required. I couldn't reproduce the failure after `make clean` and `rm */Cargo.lock` just now, though I thought I had before. Is there any state that survives through  `make clean` and `rm */Cargo.lock`?

dominic-mulligan-arm:
I don't think `rm */Cargo.lock` is sufficient to remove all state, as we have deeply nested directory trees.  You should use `make clean-cargo-lock` which removes all lock files instead.

Nevertheless, I think there's no harm in merging this, even if it apparently is no longer strictly needed, right?

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
Let's wait for a second opinion, though, before merging this.  @mathias-arm are you happy with the version bump?

ShaleXIONG (approved):

dominic-mulligan-arm:
Second opinion obtained.  +1+1 = +1 reached, merging.

egrimley-arm:
> I don't think `rm */Cargo.lock` is sufficient to remove all state, as we have deeply nested directory trees. You should use `make clean-cargo-lock` which removes all lock files instead.

`make clean-cargo-lock` seems to just do `rm -f */Cargo.lock`:

https://github.com/veracruz-project/veracruz/blob/59864749459683d240b617d7396b653a0200b157/workspaces/Makefile#L56

dominic-mulligan-arm:
Ah, I was thinking of the old build system where the top-level `Makefile` used to recursively call `clean-cargo-lock` for various system components.  With the move to workspaces, it seems the situation has become simpler...

PR #352, opened 2022-03-04
Fixed the Nitro tests for the new workspace set up

dreemkiller:
The Nitro tests is the new workspace setup were failing because the EIF file was getting picked up from the wrong place.
Added a set to the proper environment variable before calling `cargo test`.


mathias-arm (changes requested):
- I think it would be better to replace `/work/veracruz/workspaces/nitro-runtime/runtime_manager.eif` with a make variable assigned `$(abspath $(WORKSPACE_DIR)/nitro-runtime/runtime_manager.eif)`.
- I think if we should move `nitro-terminate.sh` into `workspaces/nitro-host` or alternatively invoke it  with `./crates/veracruz-server-test/nitro-terminate.sh`.

mathias-arm (approved):

mathias-arm (approved):

PR #354, opened 2022-03-08
Remove icecap/targets.nix and icecap/nix/.../*.nix.

egrimley-arm:
Also some changes under workspaces/ to stop using those files.


dominic-mulligan-arm:
This looks good to me, though let's wait for @mathias-arm's opinion.

mathias-arm (approved):

PR #358, opened 2022-03-10
Add workspace Cargo.lock files

### LINKED ISSUE #355: Check in Cargo.lock files
**Requested feature**

All `Cargo.lock` files should be checked into the repository.

**Motivation**

This guards against dependencies silently updating and bumping the MSRV, and also goes some way to address reproducible builds for Veracruz.

mathias-arm:
Fixes #355


egrimley-arm:
If #359 works it might be nice to redo this on top of that, which gets rid of `rouille`, for example.

mathias-arm:
I think it would make more sense to rebase #359 after #358. This way we can see the impact of efforts like #359 on the dependencies.

egrimley-arm:
What's the procedure for updating the `Cargo.lock` files?

egrimley-arm:
If I don't do anything special, #359 will have no effect on the `Cargo.lock` files: unused crates will still be mentioned there. If I regenerate the `Cargo.lock` files, there will be lots of changes that have nothing to do with #359.

mathias-arm:
If you make a change to `Cargo.toml` files that will result in changes in the `Cargo.lock` files, you should probably use the `localci` docker image, and run `make -C workspaces clean all`. This will insure all Cargo.lock files are updated. I removed `Cargo.lock` and `*.lock` from `.gitignore` specifically so that they would show up in `git diff` / `git status`.

We have agreed that we would periodically update the dependencies. But the rest of the time you should not re-generate (which I guess means you delete the file and let Cargo create from scratch) or run `cargo update`.

egrimley-arm (dismissed):
If I regenerate the `Cargo.lock` files I get version `1.3.2` of `bitflags` instead of `1.2.1`. Perhaps that crate was updated today.

mathias-arm:
As reminded by @geky, we could add a file to track the commit ID of the crates.io index. I don't think we should be using that for the build in general, but for reproducible builds or in this case help track updates.

mathias-arm:
> If I regenerate the `Cargo.lock` files I get version `1.3.2` of `bitflags` instead of `1.2.1`. Perhaps that crate was updated today.

That update has existed for a while. Is that with #358 or #358 + #359? In the latter case if you remove dependencies that were constraining the version of a crate, that might allow Cargo to use a newer version.

egrimley-arm:
> That update has existed for a while. Is that with #358 or #358 + #359? In the latter case if you remove dependencies that were constraining the version of a crate, that might allow Cargo to use a newer version.

It was without #359, I think.

egrimley-arm:
A makefile target to quickly regenerate the `Cargo.lock` files without the slower parts of building would be nice to have.

egrimley-arm (approved):

geky (approved):
Looks good to me 👍

PR #359, opened 2022-03-10
Remove some unused imports and dependencies.

egrimley-arm:


mathias-arm:
As mentioned in #358, I think it would be valuable for this PR to be rebase after #358 is base to see the impact. Otherwise good clean-ups.

mathias-arm (approved):

execution-engine/src/fs.rs
```diff
@@ -21,13 +21,14 @@ use policy_utils::{
     CANONICAL_STDERR_FILE_PATH, CANONICAL_STDIN_FILE_PATH, CANONICAL_STDOUT_FILE_PATH,
 };
 #[cfg(not(feature = "icecap"))]
+use std::ffi::OsString;
```
dominic-mulligan-arm, line 24: Can this be simplified to a single import?

```rust
#[cfg(not(feature = "icecap"))]
use std::{ffs::OsString, os::unix::ffi::{OsStrExt, OsStringExt}};
```

dominic-mulligan-arm (commented):
Looks good to me, barring a small change.

execution-engine/src/fs.rs
egrimley-arm: Done, but I must check `icecap` on `localci` after the rebase.

dominic-mulligan-arm (approved):

execution-engine/src/fs.rs
dominic-mulligan-arm: Great, thanks!

PR #364, opened 2022-03-11
Remove crate_name pragmas from system components

### LINKED ISSUE #361: Remove crate_name pragmas from component crates
**Requested feature**

Various `crate_name` pragmas appear in component crates which change the names of the crates being built.  These should be removed.

**Motivation**

The pragmas rename the crates, making the built crate name differ to its expected name.  They are also unnecessary when building with Cargo.

dominic-mulligan-arm:
Closes #361.

### DIFF
--- a/runtime-manager/src/lib.rs
+++ b/runtime-manager/src/lib.rs
@@ -9,6 +9,4 @@
 //! See the `LICENSE_MIT.markdown` file in the Veracruz root directory for
 //! information on licensing and copyright.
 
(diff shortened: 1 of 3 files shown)


dominic-mulligan-arm:
Marking this as ready to review: I thought it would require `Makefile` changes, but apparently not --- Linux has just built locally on my machine, and Nitro has built via CI.

dreemkiller (approved):

mathias-arm (approved):

PR #367, opened 2022-03-15
Change IceCap platform to target RealmOS

mathias-arm:
Update Veracruz IceCap support to use the `realmos` branch of IceCap.


geky:
Well we know the timeouts work now.

GitHub actions tend to run slower than locally, is it possible this is a legitimate timeout and we just need to increase the timeout >2 minutes?

geky (approved):

dominic-mulligan-arm (approved):
Gone through and it looks good to me.

PR #368, opened 2022-03-17
Delayed load of policy in IceCap attestation

### LINKED ISSUE #363: Streamline IceCap attestation
**Requested feature**

IceCap attestation flow currently sends the policy to the Runtime Manager enclave as part of the initialisation process.  This should be delayed until later in the flow.

**Motivation**

This was a feature request from @alexandref75 who noted that delaying the provisioning of the policy allowed a policy-less enclave to be setup which simplified the Veracruz FaaS work.

dominic-mulligan-arm:
Deferred provisioning of JSON policy during IceCap attestation to make the process match attestation on Nitro and Linux.

Closes #363.

### DIFF
--- a/veracruz-utils/src/platform/icecap/message.rs
+++ b/veracruz-utils/src/platform/icecap/message.rs
@@ -17,14 +17,12 @@ pub type SessionId = u32;
 /// Type of requests from the host to the realm
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum Request {
-    Initialize {
-        policy_json: String,
-    },
     Attestation {
         device_id: i32,
         challenge: Vec<u8>,
     },
-    CertificateChain {
+    Initialize {
+        policy_json: String,
         root_cert: Vec<u8>,
         compute_cert: Vec<u8>,
     },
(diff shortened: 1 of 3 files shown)


dreemkiller (dismissed):

mathias-arm (dismissed):

dominic-mulligan-arm:
This is now ready for review.

veracruz-utils/src/platform/icecap/message.rs
```diff
@@ -17,14 +17,12 @@ pub type SessionId = u32;
 /// Type of requests from the host to the realm
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum Request {
```
dreemkiller, line 19: Are we now at a point where we can unify this file with `RuntimeManagerMessage` in `veracruz-utils/src/platform/vm.rs`?

dominic-mulligan-arm: I think so, but I think there's a slightly more involved change that I want to make, namely splitting `RuntimeManagerMessage` up into a `RuntimeManagerRequest` and `RuntimeManagerResponse` type and then unifying all of the types.  I will do that next: see #374.

gbryant-arm (approved):

dominic-mulligan-arm:
+1 reached, merging.

PR #369, opened 2022-03-17
Start replacing "ring".

egrimley-arm:


egrimley-arm:
The stuff in `der.rs` is far from perfect. Mostly I've just moved the dodgy code into one place.

mathias-arm:
I find it a little bit strange to have `sha256` in `ecdsa`.

Those this PR results in uncommitted changes to `Cargo.lock` files?

egrimley-arm:
> I find it a little bit strange to have `sha256` in `ecdsa`.

Yes ... but it also felt strange to have a separate file for something so small.

> Those this PR results in uncommitted changes to `Cargo.lock` files?

Should I push another commit with the `Cargo.lock` changes?

mathias-arm:
> Should I push another commit with the `Cargo.lock` changes?

I think so. As you suggest, it seems interesting to have those changes as a separate commit at the end in a PR. In this case there is only one commit, but in other cases there might be multiple changes to `Cargo.toml` files across multiple commits, and if we'd add `Cargo.lock` changes it would probably just add superfluous diffs.

mathias-arm:
The resulting `Cargo.lock` changes are almost exclusively negative: additional crates or additional crate versions. Is it worth merging at this stage? I would suggest waiting until the benefits are larger.

egrimley-arm:
I'm now wasting another few hours rebasing this PR.

dreemkiller (dismissed):

veracruz-client/src/veracruz_client.rs
```diff
@@ -158,42 +150,46 @@ impl VeracruzClient {
         let client_priv_key = Self::read_private_key(&client_key_filename)?;
 
         // check if the certificate is valid
-        let key_pair = ring::signature::RsaKeyPair::from_der(client_priv_key.0.as_slice())
-            .map_err(|err| VeracruzClientError::RingError(format!("from_der failed:{:?}", err)))?;
-        Self::check_certificate_validity(&client_cert_filename, key_pair.public_key().as_ref())?;
+        let public_key = der::extract_public(client_priv_key.0.as_slice())
+            .map_err(|_| VeracruzClientError::RingError(format!("extract_public failed")))?;
```
gbryant-arm, line 154: We should probably rename `RingError` to something like `KeyError` or `CryptoError`

veracruz-utils/src/der.rs
```diff
@@ -0,0 +1,151 @@
...
+//! ## Licensing and copyright notice
+//!
+//! See the `LICENSE_MIT.markdown` file in the Veracruz root directory
+//! for information on licensing and copyright.
+
+/// Convert a raw signature (64 bytes) into DER encoding:
+/// two INTEGERs in a SEQUENCE.
+pub fn sig_to_der(sig: Vec<u8>) -> Result<Vec<u8>, ()> {
+    if sig.len() != 64 {
+        Err(())
```
gbryant-arm, line 16: Should we define an error here?

veracruz-utils/src/der.rs
```diff
@@ -0,0 +1,151 @@
...
+        der.push(128 + der.len() as u8);
+        der.reverse();
+        der
+    }
+}
+
+/// Extract DER length. Argument is updated to remaining input.
+fn parse_der_length(der: &mut &[u8]) -> Result<usize, ()> {
+    if der.len() < 1 || der[0] == 128 || der[0] == 255 {
+        return Err(());
```
gbryant-arm, line 58: Same as above

veracruz-client/src/veracruz_client.rs
```diff
@@ -12,19 +12,15 @@
...
-};
+use rustls::{Certificate, ClientConnection, PrivateKey};
 use rustls_pemfile;
 use std::{
     convert::TryFrom,
     io::{Read, Write},
     path::Path,
     str::from_utf8,
 };
+use veracruz_utils::der;
```
dominic-mulligan-arm, line 23: Can you group the related imports together, like so:

```rust
use veracruz_utils::{der, VERACRUZ_RUNTIME_HASH_EXTENSION_ID};
```

veracruz-client/src/veracruz_client.rs
```diff
@@ -158,42 +150,46 @@ impl VeracruzClient {
...
             let certs_pem = policy.proxy_service_cert();
-            let certs =
-                rustls_pemfile::certs(&mut certs_pem.as_bytes()).map_err(|_| {
-                    VeracruzClientError::X509ParserError(format!(
-                        "rustls_pemfile::certs found not certificates"
-                    ))
-                })?;
+            let certs = rustls_pemfile::certs(&mut certs_pem.as_bytes()).map_err(|_| {
+                VeracruzClientError::X509ParserError(format!(
+                    "rustls_pemfile::certs found not certificates"
```
dominic-mulligan-arm, line 165: sp: "found not" --> "found no"

(It's not your error, but you're touching the line here, so it may as well get fixed.)

veracruz-utils/src/der.rs
dominic-mulligan-arm: Yes...

veracruz-utils/src/der.rs
```diff
@@ -0,0 +1,151 @@
...
+
+/// Convert a raw signature (64 bytes) into DER encoding:
+/// two INTEGERs in a SEQUENCE.
+pub fn sig_to_der(sig: Vec<u8>) -> Result<Vec<u8>, ()> {
+    if sig.len() != 64 {
+        Err(())
+    } else {
+        let x = &sig[0..32];
+        let y = &sig[32..64];
+        let xlen = 32 + x[0] / 128;
```
dominic-mulligan-arm, line 20: Where do all of these constants come from?  Can they be hoisted out to the top-level as defined constants so that they're self-documenting?

veracruz-utils/src/der.rs
```diff
@@ -0,0 +1,151 @@
...
+    };
+    let mut der_public = vec![3, 66, 0];
+    der_public.extend_from_slice(&raw_public);
+
+    let mut der_private = vec![
+        48, 129, 135, 2, 1, 0, 48, 19, 6, 7, 42, 134, 72, 206, 61, 2, 1, 6, 8, 42, 134, 72, 206,
+        61, 3, 1, 7, 4, 109, 48, 107, 2, 1, 1, 4, 32,
+    ];
+    der_private.extend_from_slice(&raw_private);
+    der_private.extend_from_slice(vec![161, 68].as_slice());
```
dominic-mulligan-arm, line 143: Can't you use `&[161, 68]`?

veracruz-utils/src/ecdsa.rs
```diff
@@ -0,0 +1,70 @@
...
+//! The Veracruz Development Team.
+//!
+//! ## Licensing and copyright notice
+//!
+//! See the `LICENSE_MIT.markdown` file in the Veracruz root directory
+//! for information on licensing and copyright.
+
+use crate::der;
+use crate::sha256::sha256;
+use psa_crypto;
```
dominic-mulligan-arm, line 14: Can you group all of these related imports together?

veracruz-utils/src/ecdsa.rs
```diff
@@ -0,0 +1,70 @@
...
+                hash_alg: Hash::Sha256.into(),
+            }
+            .into(),
+        },
+    }
+}
+
+/// Generate ECDSA keypair (public, private) in DER encoding.
+pub fn generate() -> (Vec<u8>, Vec<u8>) {
+    psa_crypto::init().unwrap();
```
dominic-mulligan-arm, line 43: We should avoid `unwrap()`.  When would this fail?  (And all other uses of `unwrap()` in this file...). Can we not make these functions return an error type?

veracruz-utils/src/der.rs
```diff
@@ -0,0 +1,151 @@
...
+    public.extend_from_slice(&encode_der_length(int1.len() + int2.len()));
+    public.extend_from_slice(int1);
+    public.extend_from_slice(int2);
+    Ok(public)
+}
+
+/// Convert a keypair (public, private) into a particular DER encoding.
+pub fn keypair_to_der(raw_public: &[u8], raw_private: &[u8]) -> (Vec<u8>, Vec<u8>) {
+    if raw_public.len() != 65 || raw_private.len() != 32 {
+        panic!(
```
gbryant-arm, line 129: Can we avoid the `panic!` and return an error instead?

dominic-mulligan-arm (changes requested):

veracruz-utils/src/ecdsa.rs
gbryant-arm, line 43: Can we handle those errors without panicking?

gbryant-arm (changes requested):
Looks good besides error handling and some minor details

veracruz-utils/src/der.rs
```diff
@@ -0,0 +1,151 @@
+//! Some simple manipulations of Distinguished Encoding Rules (DER).
```
dreemkiller, line 1: Before we merge this, can we confirm that MbedTLS doesn't offer these DER-specific functions?

egrimley-arm: > Before we merge this, can we confirm that MbedTLS doesn't offer these DER-specific functions?

It does, but I hadn't merged MbedTLS yet. Now I have, and this PR has become much simpler!

gbryant-arm (approved):
LGTM

dominic-mulligan-arm (approved):
This looks good to me, though let's wait for final approval from @dreemkiller given the subject matter.

dreemkiller (approved):

dominic-mulligan-arm (approved):

PR #372, opened 2022-03-23
Add `RUNTIME_ENCLAVE_BINARY_PATH` environment variable

gbryant-arm:
Add `RUNTIME_ENCLAVE_BINARY_PATH` environment variable.
Allows to decouple execution from the build environment


dominic-mulligan-arm:
Is the `RUNTIME_ENCLAVE_BINARY_PATH` variable already set as part of the build process?

dominic-mulligan-arm:
Actually, looking at the changes again there's no overlap with IceCap so the build should not be affected.  Will review now.

PR #373, opened 2022-03-24
Remove superfluous dependencies from Proxy Attestation Server

### LINKED ISSUE #362: Remove superfluous dependencies in Proxy Attestation Service
**Requested feature**

Now that the Linux Root Enclave is almost removed, various dependencies of the Proxy Attestation Service are not needed (e.g., Diesel, and similar).  These should be removed.

**Motivation**

The removal of the Linux Root Enclave has simplified the Proxy Attestation Service, and the SQL database maintained there is now no longer needed.  All material related to that should be removed.

dominic-mulligan-arm:
Closes #362.


mathias-arm:
There are also calls to the `diesel` CLI that should removed from `Makefile` (which in turn should result in some cleanups in the docker images)

dominic-mulligan-arm:
This is now ready for review.

dreemkiller (approved):

dominic-mulligan-arm:
+1 reached, merging.

PR #375, opened 2022-03-28
Add `RUNTIME_ENCLAVE_BINARY_PATH` environment variable (2/2)

gbryant-arm:
Fix for #372

### DIFF
--- a/workspaces/linux-host/Makefile
+++ b/workspaces/linux-host/Makefile
@@ -18,7 +18,6 @@ WORKSPACE_DIR = $(abspath ..)
 OUT_DIR ?= $(abspath test-collateral)
 MEASUREMENT_FILE = $(abspath ../linux-runtime/css-linux.bin)
 MEASUREMENT_PARAMETER = --css-file $(MEASUREMENT_FILE)
-ENCLAVE_BINARY_PATH = $(abspath $(WORKSPACE_DIR)/linux-runtime/runtime_manager_enclave)
 
 include $(WORKSPACE_DIR)/common.mk
 include $(WORKSPACE_DIR)/shared.mk
@@ -30,7 +29,7 @@ TEST_PARAMETERS = DATABASE_URL=$(PROXY_ATTESTATION_SERVER_DB) \
 	VERACRUZ_TRUST_DIR=$(OUT_DIR) \
 	VERACRUZ_PROGRAM_DIR=$(OUT_DIR) \
 	VERACRUZ_DATA_DIR=$(OUT_DIR) \
-	RUNTIME_MANAGER_ENCLAVE_PATH=$(RUNTIME_MANAGER_ENCLAVE)
+	RUNTIME_ENCLAVE_BINARY_PATH=$(WORKSPACE_DIR)/linux-runtime/runtime_manager_enclave
 
 all: build test-collateral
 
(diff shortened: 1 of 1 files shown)


dominic-mulligan-arm:
@mathias-arm can you take a look at this and see whether it fixes your concerns?

workspaces/linux-host/Makefile
```diff
@@ -30,7 +29,7 @@ TEST_PARAMETERS = DATABASE_URL=$(PROXY_ATTESTATION_SERVER_DB) \
 	VERACRUZ_TRUST_DIR=$(OUT_DIR) \
 	VERACRUZ_PROGRAM_DIR=$(OUT_DIR) \
 	VERACRUZ_DATA_DIR=$(OUT_DIR) \
-	RUNTIME_MANAGER_ENCLAVE_PATH=$(RUNTIME_MANAGER_ENCLAVE)
+	RUNTIME_ENCLAVE_BINARY_PATH=$(WORKSPACE_DIR)/linux-runtime/runtime_manager_enclave
```
mathias-arm, line 32: With the changes in #339, the file `$(WORKSPACE_DIR)/linux-runtime/runtime_manager_enclave` will not be created anymore. This line should be:
```
	RUNTIME_ENCLAVE_BINARY_PATH=$(RUNTIME_MANAGER_ENCLAVE)
```

mathias-arm (changes requested):

workspaces/linux-host/Makefile
gbryant-arm: Good catch. Looks like not having `runtime_manager_enclave` breaks the tests.
Let's keep copying it from `./target`.
Making changes to #339

gbryant-arm: Fixed in #339

mathias-arm:
It seems the proper fix without a superfluous copy can't be implemented and tested in two separate PRs (#375 and #339). The easiest option is probably to move this change into #339.

gbryant-arm:
> It seems the proper fix without a superfluous copy can't be implemented and tested in two separate PRs (#375 and #339). The easiest option is probably to move this change into #339.

Yeah there seems to be a bunch of unnecessary copies:
+ Nitro: `PCR0` copied to `css-nitro.bin`, `runtime_manager_enclave` copied to `docker`
+ Linux: `runtime_manager_enclave` copied to `./` and `./css-linux.bin`

I suggest we address that in another PR

dominic-mulligan-arm:
What is the status of this PR?

mathias-arm:
My suggestion is to close this PR, and work on this (and the additional copies) as part of #375.

mathias-arm:
Sorry, yes I meant #339. I agree they are orthogonal, but they overlap a lot in what parts of the Makefile they modify. I don't think it would be possible to have two/three PRs that would not conflict. Regarding the copied files, I am worried that it might lead to using stale copies.

gbryant-arm:
> Sorry, yes I meant #339. I agree they are orthogonal, but they overlap a lot in what parts of the Makefile they modify. I don't think it would be possible to have two/three PRs that would not conflict. Regarding the copied files, I am worried that it might lead to using stale copies.

#339 changes the runtime Makefile, #375 was changing the host Makefile. No overlaps.

I opened https://github.com/veracruz-project/veracruz/issues/383 for the copies, I'd like to keep my PRs self-contained if possible

dominic-mulligan-arm:
Reopened until consensus emerges.

dominic-mulligan-arm:
OK, trying to drive a consensus about what should be done with this PR.  From talking with @gbryant-arm:

- Apparently there is an issue affecting @egrimley-arm that this PR fixes.
- Moreover, the changes here are disjoint from #339 and @gbryant-arm wants to keep the PRs self-contained, and has opened an additional PR (#383) trying to resolve the copying issue.

@mathias-arm are you happy if this PR remains open, in light of the above?

gbryant-arm:
I believe @egrimley-arm was affected by the issue described in https://github.com/veracruz-project/veracruz/pull/339/ instead

egrimley-arm:
> I believe @egrimley-arm was affected by the issue described in #339 instead

Yes, that's right, I think.

dominic-mulligan-arm:
What is the status of this PR?

gbryant-arm:
> What is the status of this PR?

Waiting for @mathias-arm 's approval

gbryant-arm:
Merging into #339

PR #380, opened 2022-04-05
WIP: Update rustls, webpki, and ring

dreemkiller:
Updated rustls, webpki, and ring to the latest released versions.
This should have been fairly simple, since the rustls API didn't change. However, the behavior of the existing functions changed quite a bit, and required quite a lot of work to figure out.

As is, the changes we now require in rustls are very minimal: a convenience function. After this is done, I will investigate removing the need for a fork of rustls entirely.

The changes we now require in webpki are limited to support for unrecognized extensions. This would be a generally useful feature. I am going to see how difficult it would be to upstream this change. After that, we may no longer require a fork of webpki.

The changes for ring involve platform-specific support for random number generation. The platforms that we have added are not of general interest and therefore are unlikely to be upstreamed as-is. We have discussed changing the way it is done to make adding platform-specific support easier. If we were to do that, it could be upstreamed (but it's unclear how likely that would be to succeed).

### DIFF
--- a/veracruz-client/Cargo.toml
+++ b/veracruz-client/Cargo.toml
@@ -31,16 +31,17 @@ mockall = { version = "0.5.0", optional = true }
 policy-utils = { path = "../policy-utils", features = ["std"] }
 rand = "0.7.0"
 reqwest = { version = "0.9", default-features = false }
-ring = "0.16.11"
+ring = "0.16.20"
 # The cargo patch mechanism does NOT work when we add function into a macro_rules!
-rustls = "0.16"
+rustls = "0.20.4"
 serde_json = "1.0"
 structopt = { version = "0.3", optional = true, features = ["wrap_help"] }
 transport-protocol = { path = "../transport-protocol" }
 veracruz-utils = { path = "../veracruz-utils", features = ["std"] }
-webpki = "0.21.2"
-webpki-roots = "0.19.0"
+webpki = "0.22.0"
+webpki-roots = "0.22.0"
 x509-parser = "0.12.0"
+rustls-pemfile = "0.3.0"
 
 [dev-dependencies]
 actix-http = "2.2.0"
--- a/workspaces/host/Cargo.toml
+++ b/workspaces/host/Cargo.toml
@@ -42,5 +42,5 @@ lto = true
 opt-level = 3
 
 [patch.crates-io]
-rustls = { path = "crates/third-party/rustls" }
+rustls = { path = "crates/third-party/rustls/rustls" }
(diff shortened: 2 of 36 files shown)


mathias-arm (changes requested):
Great job!

Previously we seemed to be pinning the version of `webpki` in some place. I think it might have been because we did not want a minor update to `webpki` to require an upgraded version of `ring`. It seems all workspaces are patching to use the vendored version of `webpki` so I guess we should be dropping the pinning everywhere.

There are a couple of place where the dependencies are not properly sorted. I will add comments.

I have not looked at the changes in the submodules yes.

workspaces/linux-host/Cargo.toml
```diff
@@ -28,5 +28,6 @@ lto = true
 opt-level = 3
 
 [patch.crates-io]
-rustls = { path = "crates/third-party/rustls" }
+rustls = { path = "crates/third-party/rustls/rustls" }
 webpki = { path = "crates/third-party/webpki" }
+ring = { path = "crates/third-party/ring" }
```
mathias-arm, line 33: Not properly sorted and missing end of line

workspaces/host/Cargo.toml
```diff
@@ -42,5 +42,5 @@ lto = true
 opt-level = 3
 
 [patch.crates-io]
```
mathias-arm, line 44: Should we add `ring` patch?

veracruz-client/Cargo.toml
```diff
@@ -31,16 +31,17 @@ mockall = { version = "0.5.0", optional = true }
...
 serde_json = "1.0"
 structopt = { version = "0.3", optional = true, features = ["wrap_help"] }
 transport-protocol = { path = "../transport-protocol" }
 veracruz-utils = { path = "../veracruz-utils", features = ["std"] }
-webpki = "0.21.2"
-webpki-roots = "0.19.0"
+webpki = "0.22.0"
+webpki-roots = "0.22.0"
 x509-parser = "0.12.0"
+rustls-pemfile = "0.3.0"
```
mathias-arm, line 44: Not properly sorted

mathias-arm:
Update: the changes to the submodules are very clean.

workspaces/host/Cargo.toml
dreemkiller: It's not strictly necessary. The changes we have for `ring` are only needed inside an enclave. `linux-host` is outside.
It might be good just to keep things consistent.

mathias-arm: You make a good point. We can leave it like this and change if we encounter an issue.

PR #385, opened 2022-04-07
CLI: Bring the CLI_QUICKSTART.markdown back up to date

geky:
Unfortunately the CLI_QUICKSTART.markdown example has regressed rather severely since the last update to it due to the ongoing improvements in Veracruz.

This PR does a couple things to hopefully improve this state:
1. Brings CLI_QUICKSTART.markdown up to date
2. Changes CLI_QUICKSTART.markdown to use the Linux runtime, which should be easier for others to reproduce due to no extra hardware requirements
3. Adds CLI_QUICKSTART.markdown to CI to prevent further regressions

Other things to note:
- Added `make install` to the Linux/shared Makefiles, this introduces the `vc-*` programs back into the PATH so Veracruz can be used interactively again.
- Made Veracruz Linux's `vc-server` program idempotent with another `include_bytes`.
- Deduplicated CLI_INSTRUCTIONS.markdown and CLI_QUICKSTART.markdown. These were slightly different, but I'm not sure having two were valuable. It wasn't clear which one to start with, and would require updating two files, further discouraging maintenance.


veracruz-server/src/veracruz_server_linux.rs
```diff
@@ -23,13 +23,18 @@ pub mod veracruz_server_linux {
     use ring::digest::{digest, SHA256};
     use std::{
         env,
-        fs::File,
+        fs::{self, File},
```
dominic-mulligan-arm, line 26: TIL!

dominic-mulligan-arm:
Not sure if the Linux failure was spurious, as nothing should have changed there...  Will try a restart.

CLI_QUICKSTART.markdown
```diff
@@ -1,29 +1,28 @@
...
+Veracruz.
 
 ## Building the Veracruz binaries
 
-First thing is we need to build the actual Veracruz command-line interfaces,
-a WebAssembly toolchain, and the test-collateral artifacts. These can be
-compiled using make, however we need to specify which Trusted Execution
-Environment (TEE) we are building for. So for example, if you are running
-Veracruz on Nitro, you would run `make nitro-cli-install`.
+First lets make sure we've built and installed the Veracruz binaries. For
```
egrimley-arm, line 14: let's

CLI_QUICKSTART.markdown
```diff
@@ -39,113 +38,304 @@ USAGE:
...
-We can request a computation as long as we have read access to the program we
-want to execute:
+## Running the Veracruz Client 
+
+In practice you may integrate the `veracruz-client` Rust crate directly into
+your application, sending data programatically via the Rust API, but for this
+example we'll use the Veracruz Client CLI interface. This provides the same
+functionality, but in a CLI form.
+
+First lets send over the program to our Veracruz server, this requires an
```
egrimley-arm, line 264: lets -> let's

, this -> . This

CLI_QUICKSTART.markdown
```diff
@@ -39,113 +38,304 @@ USAGE:
...
+    --key example/example-program-key.pem \
+    --program /program/example-binary.wasm=example/example-binary.wasm
+Loaded policy example/example-policy.json 645ae94ea86eaf15cfc04c07a17bd9b6a3b3b6c3558fae6fb93d8ee4c3e71241
+Connecting to 127.0.0.1:3017
+Submitting <enclave>/example-binary.wasm from example/example-binary.wasm
 ```
 
-And finally, we can get request the result, as long as our client certificate
-has the permission to do so:
+Then lets send over our data with identities with the "DataProvider" role.
```
egrimley-arm, line 277: let's

veracruz-server/src/veracruz_server_linux.rs
```diff
@@ -263,11 +271,19 @@ pub mod veracruz_server_linux {
...
-                .unwrap_or(RUNTIME_ENCLAVE_BINARY_PATH.to_string());
+            // temporary directory to store image
+            let runtime_enclave_binary_dir = tempfile::tempdir()?;
+
+            let runtime_enclave_binary_path = runtime_enclave_binary_dir.path().join("runtime_enclave_binary");
+            fs::write(&runtime_enclave_binary_path, RUNTIME_ENCLAVE_BINARY_IMAGE)?;
+
+            // make sure our image is executable
+            let mut runtime_enclave_binary_permissions = fs::metadata(&runtime_enclave_binary_path)?.permissions();
+            runtime_enclave_binary_permissions.set_mode(0o544);
```
egrimley-arm, line 282: Why is it `0o544`? Doesn't that mean readable but not executable by other people? Or is this different from `chmod`?

workspaces/host/Makefile
```diff
@@ -9,14 +9,26 @@
...
+BIN_DIR ?= /usr/local/cargo/bin
 include $(WORKSPACE_DIR)/common.mk
 
+
 all: crates/sdk/wasm-checker/bin/wasm-checker
 	cargo build $(PROFILE_FLAG)
 
+install: all
+	# create full names
+	cp -f target/$(PROFILE_PATH)/freestanding-execution-engine $(BIN_DIR)/freestanding-execution-engine
```
egrimley-arm, line 24: Since you're not changing the name you could shorten these commands:
```
cp -f target/$(PROFILE_PATH)/freestanding-execution-engine $(BIN_DIR)/
```
etc

workspaces/linux-host/Makefile
```diff
@@ -44,6 +46,16 @@ build:
 		--features veracruz-server/linux \
 		--features cli
 
+install: build
+	# create full names
+	cp -f target/$(PROFILE_PATH)/proxy-attestation-server $(BIN_DIR)/proxy-attestation-server
```
egrimley-arm, line 51: The same comment: you could shorten these commands.

egrimley-arm (approved):
I've just commented on a few typos and similar trivialities, and asked why it's `0o544`.

egrimley-arm:
The CI failure is: `error: environment variable `RUNTIME_ENCLAVE_BINARY_PATH` not defined`

workspaces/linux-host/Makefile
```diff
@@ -9,13 +9,14 @@
...
 	test-dependencies test-collateral test-server test-client veracruz-test
 
 default: build
 
 WORKSPACE_DIR = $(abspath ..)
 OUT_DIR ?= $(abspath test-collateral)
+BIN_DIR ?= /usr/local/cargo/bin
 MEASUREMENT_FILE = $(abspath ../linux-runtime/css-linux.bin)
 MEASUREMENT_PARAMETER = --css-file $(MEASUREMENT_FILE)
 ENCLAVE_BINARY_PATH = $(abspath $(WORKSPACE_DIR)/linux-runtime/runtime_manager_enclave)
```
egrimley-arm, line 22: I think the `RUNTIME_ENCLAVE_BINARY_PATH` thing should perhaps be added to `TEST_PARAMETERS`.

veracruz-server/src/veracruz_server_linux.rs
geky: I also learned this only recently!

veracruz-server/src/veracruz_server_linux.rs
geky: No specific reason, I just needed to set the executable bit really (yep this is basically `chmod`). It's a temporary file that just needs to be executable and readable, so I figured to keep the permissions conservative, but there isn't any real security benefit to it.

egrimley-arm: So `0o500` might work just as well? If so, it's perhaps worth doing that instead and adding an explanatory comment like `// readable and executable by user`.

geky: Sounds reasonable to me, will do

CLI_QUICKSTART.markdown
geky: I've went ahead and did a `%s/lets/let's/g` on this file :)

geky:
@dominic-mulligan-arm looks like this is CI working as intended 🙂, @egrimley-arm found that the issue was a missing environment variable. Because I've moved the runtime image into the server binary, this means we need that environment variable at build time rather than runtime.

I've pushed up a fix so hopefully this one works.

dominic-mulligan-arm (approved):
This looks good to me, though I'll wait for @egrimley-arm's second vote before merging.

egrimley-arm (approved):

dominic-mulligan-arm:
+1+1 = +1 reached, merging.

PR #386, opened 2022-04-07
Add the Counter mode AES native module

ShaleXIONG:
Add counter mode AES native module. The service is mounted on the special path `/services/aesctr.dat` and uses `psa-crypto` as the backend. The interface between WASM and the module is via a Postcard binary encoding of the following structure.
```
pub(crate) struct AesCtrService {
    key: [u8; 16],
    iv: [u8; 16],
    input_path: PathBuf,
    output_path: PathBuf,
    is_encryption: bool,
}
```

Minor changes:
- Allow the native module interfaces to have side-effects: both `service` and `try_parse` now take a mutable self. This is useful to, for example, buffer intermediate results.
- Add an example to call the AES native module.


execution-engine/Cargo.toml
```diff
@@ -38,6 +38,8 @@ num-traits = { version = "0.2", default-features = false }
 postcard = { version = "0.7.2", features = [ "alloc", "use-std" ] }
 platform-services = { path = "../platform-services" }
 policy-utils = { path = "../policy-utils" }
+psa-crypto = { git= "https://github.com/veracruz-project/rust-psa-crypto.git" }
+psa-crypto-sys = { git= "https://github.com/veracruz-project/rust-psa-crypto.git" }
```
egrimley-arm, line 42: We don't need this, I hope. We should be using `psa-crypto`, not `psa-crypto-sys`.

egrimley-arm:
The CI is failing because it's trying to use version `0.1.48` of `cmake`: search the log for `cmake v`. But *why* is it trying to use version `0.1.48` when I explicitly asked for `cmake = ">=0.1.44, <0.1.46"`?

egrimley-arm:
Ah, I see the problem!

execution-engine/Cargo.toml
```diff
@@ -38,6 +38,8 @@ num-traits = { version = "0.2", default-features = false }
 postcard = { version = "0.7.2", features = [ "alloc", "use-std" ] }
 platform-services = { path = "../platform-services" }
 policy-utils = { path = "../policy-utils" }
+psa-crypto = { git= "https://github.com/veracruz-project/rust-psa-crypto.git" }
```
egrimley-arm, line 41: This should, of course, be `{ path = "../third-party/rust-psa-crypto/psa-crypto" }`!

execution-engine/Cargo.toml
ShaleXIONG: Yes we only need the psa-crypto

execution-engine/src/native_modules/aesctr.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn enc_dec(&mut self, fs: &mut FileSystem) -> FileSystemResult<()> {
+        println!("AesCtr is called");
+        let AesCtrService {
+            key,
+            iv,
+            input_path,
+            output_path,
+            is_encryption,
+        } = self;
+        println!("AesCtr input path: {:?}", input_path);
```
dominic-mulligan-arm, line 86: ...same here.

sdk/rust-examples/aesctr-native/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+    key: [u8; 16],
+    iv: [u8; 16],
+    input_path: PathBuf,
+    output_path: PathBuf,
+    is_encryption: bool,
+}
+
+/// Example to invoke the Counter mode AES service. Encrypt `data.dat` and then 
+/// decrypt it using the `key.dat` and iv.dat`.
+fn main() -> anyhow::Result<()> {
```
dominic-mulligan-arm, line 45: Can we tidy this file up a little, e.g., by adding in some new lines between related material, capitalising log messages, and spelling the messages out fully rather than using abbreviations?

dominic-mulligan-arm (changes requested):

dominic-mulligan-arm (dismissed):

dominic-mulligan-arm:
This looks good to me, though I think it would be good if @dreemkiller or @mathias-arm cast their eyes over it before merge, given the subject matter.

sdk/rust-examples/aesctr-native/Makefile
```diff
@@ -0,0 +1,26 @@
+# Example Makefile
```
mathias-arm, line 1: Do we even need a `Makefile` for each Rust example?

mathias-arm (approved):

sdk/rust-examples/aesctr-native/Makefile
dominic-mulligan-arm: Probably not, but each example currently has one.  This can be fixed in another PR I think.

sdk/rust-examples/aesctr-native/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+anyhow = "1.0.14"
+postcard = { version = "0.7.2", features = [ "alloc", "use-std" ] }
+serde = { version = "1.0.3", features = ["derive"] }
+
+[profile.release]
```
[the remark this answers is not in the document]
dominic-mulligan-arm: @ShaleXIONG can you fix this and then I'll approve quickly?

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
+1 reached, merging.

Issue #393, opened 2022-04-11
Consensus needed to shutdown the enclave


gbryant-arm:
**Describe the bug**
A consensus from every party is needed to shutdown the enclave, which is not aligned with the idea of mutual distrusted parties.
It also makes it harder to play with Veracruz since the enclave has to be manually shutdown after computation, or a new client has to be written.

**To Reproduce**
Run a computation with Veracruz-Client on Linux or Nitro (I haven't tried IceCap). A shutdown request is sent at the end of the computation but the enclave doesn't shutdown (can be verify with `ps` on Linux or `nitro-cli` on Nitro).

**Expected behaviour**
Some options to consider:
+ Allow any party to shutdown the enclave and terminate the computation
+ Specify which parties are allowed to shutdown the enclave

In any case, a party could abuse the system by fetching the results first then shutting down the enclave before other parties get the results. A solution for this specific problem is to delay shutdown until every party gets the results.
There might be other forms of abuse.

ShaleXIONG:
We cannot predict if a party read the result they need, as it may be several files in the FS.

gbryant-arm:
We should definitely terminate the enclave after a timeout: there will always be ways for malicious parties to deny other parties access to the platform but we have to protect the platform against DoS as well.
So how about a timeout and a shutdown right, both specified in the policy file?

ShaleXIONG:
In this case, I do not think we even need to timeout, we just terminate on a single request. Also not sure we can get reliable time in all isolates.

ShaleXIONG:
I think this is done, @gbryant-arm? If so, can you close this issue.

gbryant-arm:
As of now the shutdown condition is still the same: any client can request immediate enclave shutdown.
I suggest to implement a timeout on the server side that kills the enclave some time after the computation is finished. The server isn't part of the TCB but is already trusted to correctly handle the infrastructure (spawn the enclaves). That would protect clients from each other as discussed above.
This mechanism could be combined with the consensus mechanism that we used previously.

Issue #390, opened 2022-04-07
Improve shutdown command


gbryant-arm:
**Requested feature**
Simplify the workflow of the shutdown command:
* Current workflow: Veracruz-Client sends a `request_shutdown` message to the runtime manager, which toggles a flag, sends it to Veracruz-Server, which sends a `ResetEnclave` message to the runtime manager which terminates itself (exits the send/receive loop)
* Proposed workflow: Veracruz-Client sends a `request_shutdown(is_immediate)` message to Veracruz-Server. If `is_immediate` is set, Veracruz-Server kills the enclave. If not, Veracruz-Server sends a `ResetEnclave` message to the runtime manager which terminates itself (exits the send/receive loop), which gives the runtime manager a chance to gracefully terminate

That would allow the client to kill the enclave while the WASM program is running, which isn't possible currently unless the `VeracruzServer` object gets dropped. 
More information in https://github.com/veracruz-project/veracruz/issues/389.

On another note, there are some inconsistencies across backends: the same function is called `teardown()` or `close()` depending on the backend.

**Motivation**
The shutdown workflow is confusing and the command doesn't trigger an immediate enclave shutdown.

Issue #389, opened 2022-04-07
Nitro enclave doesn't shutdown


gbryant-arm:
**Describe the bug**
There are two ways to shutdown an enclave in Veracruz:
* Send a `ResetEnclave` message to the runtime manager, which should end the runtime manager's receiving loop and the main function, resulting in the enclave shutdown. This happens when the runtime manager reports it's not active anymore, after receiving a `request_shutdown` message from Veracruz-Client
* Have Veracruz-Server directly kill the runtime manager's process (Linux and Nitro) or call `nitro-cli terminate terminate-enclave ...` (Nitro). Happens when the `VeracruzServer` object is dropped

Nitro enclave doesn't shutdown for two reasons:
* Nitro's runtime manager doesn't exit when receiving the `ResetEnclave` message. The message simply isn't processed
*  The `VeracruzServerNitro` object never gets dropped for some reason

Relates to https://github.com/veracruz-project/veracruz/issues/220

**To Reproduce**
Use Veracruz-Client to run any program in a Nitro enclave. Veracruz-Client claims at the end of the computation that the enclave is shutdown but `nitro-cli describe-enclaves` reports that the enclave is still running

**Expected behaviour**
Veracruz-Client or Veracruz-Server should be able to shutdown the Nitro enclave

PR #394, opened 2022-04-11
Fix and simplify shutdown mechanism on Nitro and other platforms

gbryant-arm:
Fix for https://github.com/veracruz-project/veracruz/issues/389 and quick temporary fix for https://github.com/veracruz-project/veracruz/issues/393.

Allow any party to request an immediate shutdown of the enclave instead of waiting for all parties to reach a unanimous consensus. This is a temporary fix for https://github.com/veracruz-project/veracruz/issues/393: more cunning proposals are being devised.
Also includes a fix for Nitro's Runtime Manager which wouldn't break from the main loop upon receiving the `ResetEnclave` message.

TODO:
* ~Remove `ResetEnclave` message: the CI fails on Linux because the server tries to terminate the enclave a second time. Relates to https://github.com/veracruz-project/veracruz/issues/390~
* ~Adjust tests~

### DIFF
--- a/runtime-manager/src/managers/mod.rs
+++ b/runtime-manager/src/managers/mod.rs
@@ -188,9 +188,9 @@ impl ProtocolState {
     /// TODO: Do something better (https://github.com/veracruz-project/veracruz/issues/393)
     pub(crate) fn request_and_check_shutdown(
         &mut self,
-        _client_id: u64,
+        client_id: u64,
     ) -> Result<bool, RuntimeManagerError> {
-        Ok(!self.expected_shutdown_sources.is_empty())
+        Ok(self.expected_shutdown_sources.contains(client_id))
     }
 
     /// Execute the program `file_name` on behalf of the client (participant) identified by `client_id`.
--- a/runtime-manager/src/managers/mod.rs
+++ b/runtime-manager/src/managers/mod.rs
@@ -190,7 +190,7 @@ impl ProtocolState {
         &mut self,
         client_id: u64,
     ) -> Result<bool, RuntimeManagerError> {
-        Ok(self.expected_shutdown_sources.contains(client_id))
+        Ok(self.expected_shutdown_sources.contains(&client_id))
     }
 
     /// Execute the program `file_name` on behalf of the client (participant) identified by `client_id`.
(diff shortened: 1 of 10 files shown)


dreemkiller (dismissed):

runtime-manager/src/managers/mod.rs
```diff
@@ -184,14 +184,13 @@ impl ProtocolState {
...
+    /// ID.
+    /// TODO: Do something better (https://github.com/veracruz-project/veracruz/issues/393)
     pub(crate) fn request_and_check_shutdown(
         &mut self,
-        client_id: u64,
+        _client_id: u64,
     ) -> Result<bool, RuntimeManagerError> {
-        self.expected_shutdown_sources.retain(|v| v != &client_id);
-        Ok(self.expected_shutdown_sources.is_empty())
+        Ok(!self.expected_shutdown_sources.is_empty())
```
ShaleXIONG, line 193: The boolean here is not as expected.
- Directly return `Ok(true)`  here? We basically by-pass `expected_shutdown_sources`? 
- Or a better way is to check if a `client_id` is in the `expected_shutdown_sources` and then return `Ok(true)`.

ShaleXIONG (dismissed):

dominic-mulligan-arm (approved):

ShaleXIONG (dismissed):

veracruz-server/src/veracruz_server_linux.rs
```diff
@@ -635,12 +581,20 @@ pub mod veracruz_server_linux {
...
-            info!("Closing...");
-            self.teardown()
+        fn shutdown_isolate(&mut self) -> Result<(), Box<dyn Error>> {
+            info!("Shutting down Linux runtime manager enclave.");
+
+            info!("Closing TCP connection...");
+            self.runtime_manager_socket.shutdown(Shutdown::Both)?;
+
+            info!("Killing and Runtime Manager process...");
+            //self.runtime_manager_process.kill().map_err(|err| VeracruzServerError::EnclaveShutdownError(err))?;
```
dominic-mulligan-arm, line 593: Stray comment?

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
+1 reached, merging.

Issue #396, opened 2022-04-12
Test cases on calling multiple programs in Veracruz


ShaleXIONG:
**Requested feature**
Add test cases on calling multiple programs in sequence in Veracruz.
- Add at least one case in `veracruz-test`.
- Add parameters in `freestanding-execution-engine`.

**Motivation**
This links to #377. This issue provides minimum examples so further examples can be added.

**Additional context**
Requested and reported by @Mo-Fatah

Mo-Fatah:
@ShaleXIONG  what is the policy of the environment variables? 
For examples, If I run multiple programs in FEE and I provided the parameter `--arg 100` , Will this `100` considered an environment variable for the first program ONLY or all the programs ?

ShaleXIONG:
> @ShaleXIONG what is the policy of the environment variables? For examples, If I run multiple programs in FEE and I provided the parameter `--arg 100` , Will this `100` considered an environment variable for the first program ONLY or all the programs ?

It will be passed to all programs actual. In fact we did not have much test on passing environment variables in Veracruz itself.

Issue #377, opened 2022-04-01
Add more examples


ShaleXIONG:
**Requested feature**
We want more examples, particularly, in the following areas, (but not restricted):
- I/O related examples. This can give us more insight into whether our implementation of WASI makes sense and works as expected. 
- Pipeline examples. This will be similar to the Linux pipeline  `command1 | command2 | command3`. In the Veracruz content, the first program, `prog_1`, writes the result to a file, e.g., `result_1.dat` and the second program `prog_2` reads the previous result in `result_1.dat` and then writes the new result to, e.g., `result_2.dat` and so on. In this setting, the WASM programs may be compiled from different source languages such as a combination of Rust and C.
- Examples in other languages. More languages can be compiled to the WASM binaries now. It is worth exploring and providing a guide to compile more languages to WASM/WASI binaries.

Note that, each bullet point above is a large group of examples, and there are other examples we might fail to cover. Everyone is welcome to contribute here. 

**Motivation**

More examples can help us to identify new use-cases and features for Veracruz.

**Additional context**

This is a suitable starting point for people wanting to work on Veracruz. Newcomers will learn how to build and use our docker image under the directory `docker`, and then compile the entire Veracruz project to a backend, (the easiest one is Linux), then use the `sdk/freestanding-execution-engine` to check if the new examples can be executed in Veracruz runtime.

paul-111:
Can i work upon the 3rd example? Providing a guide to compile more languages to WASM/WASI binaries?

ShaleXIONG:
> Can i work upon the 3rd example? Providing a guide to compile more languages to WASM/WASI binaries?

Hi, of course. This issue is open for everyone to contribute.

ShaleXIONG:
> Hi @ShaleXIONG , I am a little confused with the second point (Pipeline examples). How would I feed more than one program to the `freestanding-execution-engine` , while it accepts only one file after the `--program` parameter.
> 
> `USAGE: freestanding-execution-engine --arg <ARG>... --dump-stderr --dump-stdout --enable-clock --input-source <DIRECTORIES>... --output-source <DIRECTORIES>... --program <FILE> `

That is a good question. I do not think `freestanding-execution-engine` can call more than one program. I will patch it soon (in days). However, if you use the full Veracruz, with proper client and server sides, it has no such limitation. Free feel to try before the patch on `freestanding-execution-engine` is live.

PR #398, opened 2022-04-13
Parameter for calling multiple programs in FEE.

ShaleXIONG:
Modify the freestanding execution engine, allowing multiple values in the program parameter. All programs will be called in order. E.g.
```
./freestand-execution-engine -i input programs -o output -p program/FIRST.wasm program/SECOND.wasm ...
```
the engine will execute `FIRST.wasm`, then `SECOND.wasm` and so on. The latter can observe any result of the former if they have permission.
 
Minors:
- Add makefile to `sort-numbers` rust example
- Add a new `random-u32-list` example that generates the expected input for the `sort-numbers`. This is for testing calling of multiple programs

Request changes of #396 and #377

### DIFF
--- a/sdk/freestanding-execution-engine/src/main.rs
+++ b/sdk/freestanding-execution-engine/src/main.rs
@@ -68,8 +68,8 @@ struct CommandLineOptions {
     input_sources: Vec<String>,
     /// The list of file names passed as ouput.
     output_sources: Vec<String>,
-    /// The filename passed as the WASM program to be executed.
-    binary: String,
+    /// The pathes passed as the WASM programs to be executed (in order).
+    program_sources: Vec<String>,
     /// The execution strategy to use when performing the computation.
     execution_strategy: ExecutionStrategy,
     /// Whether the contents of `stdout` should be dumped before exiting.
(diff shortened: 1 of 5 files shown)


sdk/freestanding-execution-engine/src/main.rs
```diff
@@ -123,8 +123,8 @@ fn parse_command_line() -> Result<CommandLineOptions, Box<dyn Error>> {
                 .short("p")
                 .long("program")
                 .value_name("FILE")
-                .help("Path to the WASM binary to be executed.")
-                .multiple(false)
+                .help("Pathes to the WASM binary to be executed. It executes in order.")
```
dominic-mulligan-arm, line 126: sp: Pathes --> Paths

sdk/freestanding-execution-engine/src/main.rs
```diff
@@ -334,61 +341,66 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-    let program = vfs.read_file_by_absolute_path(prog_file_abs_path)?;
-    let return_code = execute(
-        &cmdline.execution_strategy,
-        vfs.spawn(&program_id)?,
-        program,
-        options,
-    )?;
-    info!("return code: {:?}", return_code);
-    info!("time: {} micro seconds", main_time.elapsed().as_micros());
+    info!("Invoking programs on in order {:?}.", cmdline.program_sources);
```
dominic-mulligan-arm, line 359: sp: "on in order" --> "in order"

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
+1 reached, merging.

PR #409, opened 2022-04-19
update rustls, ring, and webpki to the latest versions

dreemkiller:
Updated rustls, webpki, and ring to the latest released versions.
This should have been fairly simple, since the rustls API didn't change. However, the behavior of the existing functions changed quite a bit, and required quite a lot of work to figure out.

As is, the changes we now require in rustls are very minimal: a convenience function. After this is done, I will investigate removing the need for a fork of rustls entirely.

The changes we now require in webpki are limited to support for unrecognized extensions. This would be a generally useful feature. I am going to see how difficult it would be to upstream this change. After that, we may no longer require a fork of webpki.

The changes for ring involve platform-specific support for random number generation. The platforms that we have added are not of general interest and therefore are unlikely to be upstreamed as-is. We have discussed changing the way it is done to make adding platform-specific support easier. If we were to do that, it could be upstreamed (but it's unclear how likely that would be to succeed).

### DIFF
--- a/session-manager/Cargo.toml
+++ b/session-manager/Cargo.toml
@@ -22,7 +22,7 @@ std = [
 [dependencies]
 err-derive = "0.2"
 policy-utils = { path = "../policy-utils" }
-ring = "0.16.11"
+ring = "0.16.20"
 rustls = "0.20.4"
 rustls-pemfile = "0.3.0"
 veracruz-utils = { path = "../veracruz-utils" }
--- a/session-manager/Cargo.toml
+++ b/session-manager/Cargo.toml
@@ -23,6 +23,8 @@ std = [
 err-derive = "0.2"
 policy-utils = { path = "../policy-utils" }
 ring = "0.16.11"
-rustls = "0.16"
-webpki = "=0.21.2"
-webpki-roots = "0.19.0"
+rustls = "0.20.4"
+rustls-pemfile = "0.3.0"
+veracruz-utils = { path = "../veracruz-utils" }
+webpki = "0.22"
+webpki-roots = "0.22"
--- a/session-manager/src/session.rs
+++ b/session-manager/src/session.rs
@@ -19,7 +19,7 @@ use std::{
     vec::Vec,
 };
 
-use rustls::{Certificate, ServerSession, Session as TLSSession};
+use rustls::{Certificate, ServerConnection};
 
 ////////////////////////////////////////////////////////////////////////////////
 // Sessions.
@@ -34,7 +34,7 @@ pub type Principal = Identity<Certificate>;
 /// their identifying information.
 pub struct Session {
     /// The TLS server session.
-    tls_session: ServerSession,
+    tls_connection: ServerConnection,
     /// The list of principals, their identities, and roles in the Veracruz
     /// computation.
     principals: Vec<Principal>,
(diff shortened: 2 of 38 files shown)


session-manager/Cargo.toml
```diff
@@ -22,7 +22,9 @@ std = [
...
 policy-utils = { path = "../policy-utils" }
-ring = "0.16.11"
-rustls = "0.16"
-webpki = "=0.21.2"
-webpki-roots = "0.19.0"
+ring = "0.16.20"
+rustls = "0.20.4"
+rustls-pemfile = "0.3.0"
+veracruz-utils = { path = "../veracruz-utils" }
+webpki = "=0.22"
```
dominic-mulligan-arm, line 29: Does `webpki`'s version need to be pinned?  If so, can you add a comment in the `Cargo.toml` file indicating why?

session-manager/src/session.rs
```diff
@@ -43,27 +43,30 @@ pub struct Session {
 impl Session {
     /// Creates a new session from a server configuration and a list of
     /// principals.
-    pub fn new(config: rustls::ServerConfig, principals: &Vec<Principal>) -> Self {
-        let tls_session = ServerSession::new(&std::sync::Arc::new(config));
+    pub fn new(config: rustls::ServerConfig, principals: &Vec<Principal>) -> Result<Self, SessionManagerError> {
+        let mut tls_connection = ServerConnection::new(std::sync::Arc::new(config))?;
+        tls_connection.set_buffer_limit(Some(512 * 1024));
```
dominic-mulligan-arm, line 48: Are these numbers arbitrary, or do they correspond to something?

session-manager/src/session.rs
```diff
@@ -43,27 +43,30 @@ pub struct Session {
...
 
-        Session {
-            tls_session: tls_session,
+        Ok(Session {
+            tls_connection: tls_connection,
             principals: principals.to_vec(),
```
dominic-mulligan-arm, line 52: Is it not better to take `principals` as a `Vec<Principal>` and avoid allocating here?

veracruz-server-test/Cargo.toml
```diff
@@ -41,10 +41,11 @@ lazy_static = "1.4"
...
-rustls = "0.16"
+ring = "0.16.20"
+rustls = "0.20.4"
+rustls-pemfile = "0.3.0"
 transport-protocol = { path = "../transport-protocol" }
 veracruz-server = { path = "../veracruz-server" }
 veracruz-utils = { path = "../veracruz-utils", optional = true }
-webpki = "=0.21.2"
-webpki-roots = "0.19.0"
+webpki = "=0.22"
```
dominic-mulligan-arm, line 50: Same comment here regarding version pinning.

veracruz-server/Cargo.toml
```diff
@@ -62,13 +62,13 @@ nix = { version = "0.15", optional = true }
...
+rustls = "0.20.4"
 serde = { version = "1.0.115", default-features = false, features = ["derive"] }
 serde_json = "1.0"
 signal-hook = { version = "0.3.13", optional = true }
 structopt = { version = "0.3", optional = true, features = ["wrap_help"] }
 tempfile = { version = "3.2.0", optional = true }
 transport-protocol = { path = "../transport-protocol" }
 veracruz-utils = { path = "../veracruz-utils", optional = true }
-webpki = "=0.21.2"
+webpki = "=0.22"
```
dominic-mulligan-arm, line 74: ...and here.

veracruz-utils/Cargo.toml
```diff
@@ -25,10 +25,10 @@ std = [
...
 err-derive = "0.2"
-ring = "0.16.11"
+ring = "0.16.20"
 # The cargo patch mechanism does NOT work when we add function into a macro_rules!
-rustls = { version = "0.16", optional = true }
+rustls = { version = "0.20.4" }
 serde = { version = "1.0.115", default-features = false, optional = true }
 serde_json = { version = "1.0", default-features = false }
-webpki = "=0.21.2"
+webpki = "=0.22"
```
dominic-mulligan-arm, line 33: ...here too.

dominic-mulligan-arm (changes requested):
Looks good to me, just a few minor points/questions.

mathias-arm (changes requested):
The changes to `Cargo.lock` files have many different regressions:
- duplicated versions of `quote`, `syn`, `proc-macro2`, `bindgen`, `shlex`, `which`, `env_logger`, etc.
- `libc` is sometimes also duplicated but also inconsistent among workspaces.
- `rust-optee-trustzone-sdk` and `opte-utee-*` are re-appearing because the changes in `nitro-enclave-attestation-document` did not build on the `veracruz-v2` branch (see `mathias_update_dependencies` branch)

I created a rebase of your branch in https://github.com/mathias-arm/veracruz/tree/update_rustls_0.20.4 that filtered out the changes to `Cargo.lock` files and added a separate commit for `Cargo.lock` files and using the changes in `mathias_update_dependencies` branch of `nitro-enclave-attestation-document`. This results is much smaller and saner result (see https://github.com/mathias-arm/veracruz/commit/bb0678fe92478948910044fa31815615b3b1bc28).

session-manager/src/session.rs
dreemkiller: What specific syntax are you talking about here? I'm not familiar with that pattern.

session-manager/src/session.rs
dreemkiller: removed it as it's no longer necessary

dreemkiller:
I think I've resolved the lock file problems.

mathias-arm:
> I think I've resolved the lock file problems.

You are missing changes to the `Cargo.lock` files for the IceCap workspaces.

The version of [nitro-enclave-attestation-document](https://github.com/veracruz-project/nitro-enclave-attestation-document) you are using revert the un-pin to `serde_with` that was in the `veracruz-v2` branch. I create a PR (https://github.com/veracruz-project/nitro-enclave-attestation-document/pull/5) to rebase your change on top of `veracruz-v2` and merge that into `main`.

dreemkiller:
I have added the icecap lock file updates

mathias-arm (approved):

mathias-arm:
I think we need to update `third-party/nitro-enclave-attestation-document` to use the commit after https://github.com/veracruz-project/nitro-enclave-attestation-document/pull/5 and changing the target branch to be `main` in `.gitmodules`.

PR #411, opened 2022-04-20
Move content of `veracruz-docker-image` repository directly in `veracruz`

mathias-arm:
The `veracruz-docker-image` repository contained artifacts that we did not necessarily want to add to the main `veracruz` repository (mainly to avoid bloating it). Later developments have since significantly reduced the size, leaving us with unnecessary procedural overhead of needing pull requests in both repositories.

Unfortunately `git` does deal very well with deleting a submodule and adding a directory with the same path, so you will need to remove the `docker` directory before you can checkout/pull. You also might want to save `aws-nitro-enclaves-cli` to avoid a long rebuild.

```
mv docker/aws-nitro-enclaves-cli .
rm -rf docker
git checkout pr/docker-merge
mv aws-nitro-enclaves-cli docker
```


egrimley-arm (approved):

dominic-mulligan-arm (approved):
LGTM.  Thanks @mathias-arm this should smooth development out a lot!

dominic-mulligan-arm:
+1+1 = +1 reached, merging.

PR #421, opened 2022-04-26
Some housekeeping: .gitignore, test targets, ...

egrimley-arm:

PR #427, opened 2022-04-27
Some housekeeping: .gitignore, test targets, ... (#421) without PROFILE=dev

egrimley-arm:


egrimley-arm:
This was just an experiment.

PR #428, opened 2022-04-28
Add "test-server" to "icecap" test and remove PROFILE=dev.

egrimley-arm:
There was a typo:

    test-client veracruz-test veracruz-test

instead of:

    test-client test-server veracruz-test

But it turns out that `test-server` is very slow with `PROFILE=dev`, so this is removed, though it would be good to understand why it is so slow and perhaps put it back one day.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,7 +59,7 @@ jobs:
       - name: Running IceCap test script
         id: icecap-build
         run: |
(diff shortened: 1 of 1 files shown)


dominic-mulligan-arm (approved):

dominic-mulligan-arm:
Looks good to me, though wait for one of @mathias-arm or @geky to approve, given the subject matter, and so that they are aware of the change.

dominic-mulligan-arm:
Hmm, too late, I didn't notice this was based on top of another another PR.

PR #430, opened 2022-04-28
Bump versions of library dependencies to fix security vulnerabilities identified by dependabot

dominic-mulligan-arm:
Fixed all problematic libraries that we rely on *directly*, and which dependabot identified as having issues.  All other issues seem to be caused by transitive dependencies which I'll fix in another PR.


dominic-mulligan-arm:
Closing this as it's been superseded by other changes.

PR #433, opened 2022-04-28
WIP: Test to remove our need for a rustls fork

dreemkiller:
Our fork of rustls has a function added to the CipherSuite enum to convert from a string to the enum.

This is a test to see how dirty it would be to do it in our code instead, obviating the need for our fork of rustls.


egrimley-arm:
Since `ring` and `rustls` have now gone away, is this PR obsolete and can it be closed?

PR #434, opened 2022-04-28
Missing Cargo.lock updates

mathias-arm:
It seems some of the recently merged changes were missing the `Cargo.lock` changes.

### DIFF
--- a/workspaces/applications/Cargo.lock
+++ b/workspaces/applications/Cargo.lock
@@ -15,6 +15,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
+name = "aesctr-native"
+version = "0.3.0"
+dependencies = [
+ "anyhow",
+ "postcard",
+ "serde",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.18"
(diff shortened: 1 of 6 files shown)


egrimley-arm:
I tried `make icecap && make linux && make nitro` on `main` and arrived at a similar set of changes. The main differences were:

* `"log"` added to list of dependencies in some files
* `unicode-segmentation` removed from some files
* `aws-nitro-enclaves-nsm-api`: 0.2.0 -> 0.2.1
* `cbindgen`: 0.15.0 -> 0.21.0
* `heck`: 0.3.3 -> 0.4.0
* `libc`: 0.2.119 -> 0.2.124
* `nsm-lib`: 0.2.0 -> 0.2.1
* `which`: 4.2.4 -> 4.2.5

That's quite a lot of differences and more than I expected.

When I built starting with this commit I got only two changes:

* `"log"` added in `host/Cargo.lock`
* `"log"` added in `icecap-runtime/Cargo.lock`

Let's merge this anyway before anything else gets merged!

egrimley-arm (approved):

mathias-arm:
In my PR I rolled-back `which` for `workspaces/linux-host/Cargo.lock` from 4.2.5 to 4.2.4 because this triggers the `libc` update. I would like the `libc` update to be coordinated with across platforms because IceCap uses a fork of `libc` and we need to rebase it (it should not be too hard.

The `aws-nitro-enclaves-nsm-api` /  `nsm-lib` updates should not happen after #409 where we pin to `v0.2.0` tag.

PR #437, opened 2022-05-03
Use a single function in veracruz_utils for SHA256.

egrimley-arm:


dreemkiller (dismissed):

egrimley-arm:
I've just rebased and added `Cargo.lock` files.

dreemkiller (approved):

PR #438, opened 2022-05-03
attestation_core.c: Use psa_hash_compute instead of mbedtls functions.

egrimley-arm:
Because of an `if (0 & ...)` the hash computation is never executed, but I have tested it in isolation.

Also delete the file `sha256.c`, which was already unused.


dreemkiller:
Any idea why this was in there in the first place? I would think the PSA Attestation code would be heavily biased towards the PSA Crypto API instead of the MbedTLS API.
I think @mathias-arm should be able to answer that.

mathias-arm (approved):
The changes look good.

The rest of `attestation_core.c` is a little bit confusing with a lot of code #if'ed out or commented out.

Issue #280, opened 2021-11-29
Specify maximum amount of memory available per session 


gbryant-arm:
**Requested feature**
A policy field to specify the maximum amount of memory available per session/enclave, enforced by the entity in charge of spawning the runtime manager.

**Motivation**
A malicious client could send arbitrary large files to the runtime manager, or a program could allocate large amounts of memory, which could result in runtime failure and threaten other computations running on the same machine in other execution engines. In extreme cases, this could be used as a side channel to exchange information between programs running on the same machine (this should be the responsibility of the TEE though).
We already have means to specify the maximum amount of memory an enclave may use, for example Nitro enclaves are spawn with 256MB memory.
However, it's not clear to the participants how much memory is available for the computation, and it's not homogeneously configurable across TEE backends.

PR #439, opened 2022-05-03
Specify maximum amount of memory available (Nitro)

gbryant-arm:
Add maximum isolate memory to policy (Nitro only for now).
Addresses https://github.com/veracruz-project/veracruz/issues/280

### DIFF
--- a/policy-utils/src/policy.rs
+++ b/policy-utils/src/policy.rs
@@ -97,7 +97,7 @@ pub struct Policy {
     enable_clock: bool,
     /// The maximum amount of memory in MiB available to the isolate. Only
     /// enforced in Nitro for now.
-    max_memory: u32,
+    max_memory_mib: u32,
     /// Hash of the JSON representation if the Policy was parsed from a file.
     #[serde(skip)]
     policy_hash: Option<String>,
@@ -122,7 +122,7 @@ impl Policy {
         execution_strategy: ExecutionStrategy,
         file_hashes: Vec<FileHash>,
         enable_clock: bool,
-        max_memory: u32,
+        max_memory_mib: u32,
     ) -> Result<Self, PolicyError> {
         let policy = Self {
             identities,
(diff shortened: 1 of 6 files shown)


test-collateral/generate-policy/src/main.rs
```diff
@@ -169,6 +172,7 @@ impl Arguments {
             enclave_debug_mode: false,
             execution_strategy: String::new(),
             enable_clock: false,
+            max_memory: 0,
```
dreemkiller, line 175: Would we rather set a rational default value here?
Also, (this is a question up for discussion, not a demand): Do we want to put the units in the field name:
```
max_memory_mb
```
or something or other?

policy-utils/src/policy.rs
```diff
@@ -120,6 +122,7 @@ impl Policy {
         execution_strategy: ExecutionStrategy,
         file_hashes: Vec<FileHash>,
         enable_clock: bool,
+        max_memory: u32,
```
dreemkiller, line 125: Maybe instead of `max_memory` make it `max_memory_mb` so users/policy readers/writers aren't left wondering

gbryant-arm: Good idea

test-collateral/generate-policy/src/main.rs
gbryant-arm: > Would we rather set a rational default value here?

Hmm like a fraction?

dreemkiller: ha ha!

Like "256" - which is what we currently use.

dreemkiller (dismissed):

test-collateral/generate-policy/src/main.rs
gbryant-arm: Haha got it :+1:

dreemkiller (approved):

dominic-mulligan-arm:
Seems to have been a timeout of one of the IceCap tests.  I've rerun the failing CI build target.

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
+1+1 = +1 reached, merging.

Issue #429, opened 2022-04-28
Make IceCap server tests more reliable


egrimley-arm:
Problems we have seen include:

* A single test taking more than 30 minutes on GitHub with `PROFILE=dev` and passing (apparently).
* The entire `server-test` suite finishing in about 3 minutes on `localci` without `PROFILE=dev` and passing (apparently).
* Huge variation in how long a test run takes both on GitHub and on `localci`.

The command to run just the IceCap server tests is:

`make -C workspaces/icecap-host test-server`

dominic-mulligan-arm:
I also saw this yesterday: on `localci` the `icecap` `server-test` seemed to hang indefinitely with `PROFILE=dev` on the `main` branch.  There were a few diagnostic about allocating pages produced on the command line, but then it just seemed to loop forever, though I don't think I waited long enough to see whether 30 minutes would suffice for it to do anything...

egrimley-arm:
Though this test has sometimes succeeded in about 5 minutes,
`test_phase1_new_session` timed out after 2400 s here:
https://github.com/veracruz-project/veracruz/runs/6211854461

egrimley-arm:
Here we seem to have the same problem after #440 was applied:
https://github.com/veracruz-project/veracruz/runs/6311243609

mathias-arm:
> Here we seem to have the same problem after #440 was applied: https://github.com/veracruz-project/veracruz/runs/6311243609

It's a different issue. The changes in the draft PR make IceCap crash.

PR #440, opened 2022-05-04
Update IceCap RealmOS and make tests more reliable

mathias-arm:
- Include latest IceCap RealmOS updates
- Work-around for runtime-manager/console-server race-condition (#429)
- Fix for `pkill` status code in `quickstart` GitHub Actions job

### DIFF
--- a/scripts/icecap-git-reset.sh
+++ b/scripts/icecap-git-reset.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+set -e
+
+cd $(dirname $(dirname $(realpath $0)))
+
+git submodule update --init --recursive
+
+(cd icecap/icecap ; git stash ; git fetch ; git checkout realmos ; git branch -M backup-$(date +%Y-%m-%d-%H%M%S) ; git checkout -B realmos origin/realmos)
+(cd workspaces/icecap-runtime/deps/seL4 ; git stash ; git fetch ; git checkout realmos ; git branch -M backup-$(date +%Y-%m-%d-%H%M%S) ; git checkout -B realmos origin/realmos)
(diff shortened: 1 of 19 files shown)


scripts/icecap-git-reset.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
```
egrimley-arm, line 2: Could we have a comment explaining what this script is for?

egrimley-arm (commented):
You might as well update `icecap/icecap` to `85a5c6900c4ce162bca8bee86f9a0e2af47e65ab` (including `strncmp`) while you're at it.

gbryant-arm (dismissed):
LGTM

scripts/icecap-git-reset.sh
dominic-mulligan-arm: (It also needs the standard copyright and licensing header at the top of the file.)

workspaces/icecap-runtime/src/virtio-console-server/src/main.rs
```diff
@@ -129,6 +129,11 @@ pub unsafe extern "C" fn virtio_virt_to_phys(vaddr: usize) -> usize {
 
 // entry point
 fn main(config: Config) -> Fallible<()> {
+    // FIXME: Yield 10 times to let the runtime start before the console server
```
dominic-mulligan-arm, line 132: Is this FIXME reflected in the issue tracker so it's not forgotten?

dominic-mulligan-arm (changes requested):

mathias-arm:
#443 tracks a better fix for the for runtime-manager/console-server race-condition

egrimley-arm (approved):

PR #442, opened 2022-05-05
veracruz_client.rs: Delete contents of check_runtime_hash.

egrimley-arm:
To see if any test fails.


gbryant-arm:
I usually do that when debugging locally. Saves a whole lot of time. Compiling binaries in debug mode should bypass the check but doesn't always work in Nitro.

dominic-mulligan-arm:
Is this PR needed?

egrimley-arm:
> Is this PR needed?

No, it was just an experiment. Closing!

PR #446, opened 2022-05-06
Update to Rust 1.56.1 and libc 0.2.124

mathias-arm:
Replaces #444

- Update `icecap-runtime` to use Rust `1.56.1` instead of `nightly-2021-08-25` (which does not support 2021 edition).
- Update `libc` to version `0.2.124`
- Aligning versions for `nix`, `base64`, `x509-parser`, `getrandom` crates to avoid duplicate or superfluous dependencies.


dreemkiller (approved):

geky (approved):

mathias-arm:
+1+1 = +1 reached, merging.

PR #448, opened 2022-05-06
Additional clean-ups removing diesel from proxy-attestation-server

mathias-arm:
Follow-up to #373


dreemkiller (approved):

mathias-arm:
+1 reached, merging.

PR #453, opened 2022-05-10
Fix Clippy lints: veracruz-server and veracruz-client

Mo-Fatah:
```
Developer Certificate of Origin
Version 1.1

Copyright (C) 2004, 2006 The Linux Foundation and its contributors.

Everyone is permitted to copy and distribute verbatim copies of this
license document, but changing it is not allowed.


Developer's Certificate of Origin 1.1

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
    have the right to submit it under the open source license
    indicated in the file; or

(b) The contribution is based upon previous work that, to the best
    of my knowledge, is covered under an appropriate open source
    license and I have the right under that license to submit that
    work with modifications, whether created in whole or in part
    by me, under the same open source license (unless I am
    permitted to submit under a different license), as indicated
    in the file; or

(c) The contribution was provided directly to me by some other
    person who certified (a), (b) or (c) and I have not modified
    it.

(d) I understand and agree that this project and the contribution
    are public and that a record of the contribution (including all
    personal information I submit with it, including my sign-off) is
    maintained indefinitely and may be redistributed consistent with
    this project or the open source license(s) involved.
```

## Major changes :
- Fix clippy warnings for `veracruz-server` and `veracruz-client`
- Almost all of the fixes relates to styling, useless extra code, slight performance improvement.

### Unhandled lints
There are two clippy warnings that I left. 
- The first one [here](https://github.com/veracruz-project/veracruz/blob/1252ad094f0b87fdf026747bd2545583b3fb0778/veracruz-server/src/veracruz_server_linux.rs#L463), clippy warns that this is a very complex type and it needs to be refactored into `type` definitions. If we followed clippy this will lead to change the [trait function](https://github.com/veracruz-project/veracruz/blob/1252ad094f0b87fdf026747bd2545583b3fb0778/veracruz-server/src/veracruz_server.rs#L222) which will require multiple changes inside `veracruz-server`
- The second one [here](https://github.com/veracruz-project/veracruz/blob/1252ad094f0b87fdf026747bd2545583b3fb0778/veracruz-server/src/veracruz_server_linux.rs#L13), clippy complains that naming the module `veracruz_server_linux` inside `veracruz_server_linux.rs` is a bad styling. Changing this will require multiple changes inside `veracruz-server` also. This is also generated inside `veracruz_server_nitro` .

Unless you think we should change the remaining warnings, I am thinking of project-wide suppressing for those two rules `clippy::module_inception` and `clippy::type_complexty`.

## Minor changes
I updated the clippy targets inside `linux-host/Makefile` and `icecap-host` to be synced with the changes in the `build` targets.

### DIFF
--- a/veracruz-client/src/veracruz_client.rs
+++ b/veracruz-client/src/veracruz_client.rs
@@ -105,7 +105,7 @@ impl VeracruzClient {
                 expected: parsed_cert.subject_pki.subject_public_key.data.to_vec(),
                 received: public_key.to_vec(),
             })
-        } else if let None = parsed_cert.validity.time_to_expiration() {
+        } else if parsed_cert.validity.time_to_expiration().is_none() {
             Err(VeracruzClientError::CertificateExpireError(
                 client_cert_filename.as_ref().to_string_lossy().to_string(),
             ))
@@ -122,7 +122,7 @@ impl VeracruzClient {
         client_key_filename: P2,
         policy_json: &str,
     ) -> Result<VeracruzClient, VeracruzClientError> {
-        let policy = Policy::from_json(&policy_json)?;
+        let policy = Policy::from_json(policy_json)?;
         let policy_hash = policy
             .policy_hash()
             .expect("policy did not hash json?")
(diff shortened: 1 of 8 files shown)


dominic-mulligan-arm:
I agree that `module_inception` and `type_complexity` lints are almost always something that can be ignored.  Thanks for taking a look at this, by the way!

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
+1 reached, merging.

PR #455, opened 2022-05-10
Update Wasmtime to version 35

ShaleXIONG:
Update Wasmtime to version 35.
- Change the code structure, using the `Linker` structure in Wasmtime for linking all external functions.
- Remove the global state in Wasmtime end, as the newer engine carries a state through all external calls. The old global fyle system handler becomes the internal state in the execution engine. 

Minor:
- Use `strum` to generate conversion between string and WASI function names.

Note:
The newest version of Wasmtime is 36, yet it requires rust-toolchain 1.60.

### DIFF
--- a/execution-engine/src/fs.rs
+++ b/execution-engine/src/fs.rs
@@ -1696,10 +1696,7 @@ impl FileSystem {
             ctime: Timestamp::from_nanos(0),
         };
         let node = InodeEntry {
-            //current: inode,
-            //parent: inode,
             file_stat,
-            //path: PathBuf::from(""),
             data: InodeImpl::File(Vec::new()),
         };
         self.lock_inode_table()?.insert(inode, node)?;
--- a/execution-engine/src/fs.rs
+++ b/execution-engine/src/fs.rs
@@ -27,13 +27,13 @@ use std::{
     collections::HashMap,
     convert::{AsRef, TryFrom},
     fmt::Debug,
-    // TODO: wait for icecap to support direct conversion between bytes and os_str, bypassing
-    // potential utf-8 encoding check
     path::{Component, Path, PathBuf},
     string::String,
     sync::{Arc, Mutex, MutexGuard},
     vec::Vec,
 };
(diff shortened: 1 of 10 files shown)


mathias-arm (dismissed):
This looks good.

I added a commit to update `Cargo.lock` files.

execution-engine/src/wasi/wasmtime.rs
```diff
@@ -70,84 +73,106 @@ macro_rules! convert_wasi_arg {
...
             address: address as usize,
             length: length as usize,
         }))
     }
 
     fn get_size(&self) -> FileSystemResult<u32> {
-        let memory = match self
+        // NOTE: manually and temporarily change the mutability.
+        let memory = match unsafe { (self as *const Self as *mut Self).as_mut() }
+            .unwrap()
```
dominic-mulligan-arm, line 168: When can this fail?

dominic-mulligan-arm (changes requested):
Great work @ShaleXIONG.  Just a few small changes/questions.

execution-engine/src/fs.rs
```diff
@@ -27,13 +27,13 @@ use std::{
...
     convert::{AsRef, TryFrom},
     fmt::Debug,
-    // TODO: wait for icecap to support direct conversion between bytes and os_str, bypassing
-    // potential utf-8 encoding check
     path::{Component, Path, PathBuf},
     string::String,
     sync::{Arc, Mutex, MutexGuard},
     vec::Vec,
 };
+// TODO: wait for icecap to support direct conversion between bytes and os_str, bypassing
```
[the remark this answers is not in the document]
ShaleXIONG: This is still a valid comment, but I just move it to the correct place.

execution-engine/src/wasi/wasmtime.rs
ShaleXIONG: It only fails if the raw pointer is null. It never happens here.

dominic-mulligan-arm:
+1 reached, merging.

PR #457, opened 2022-05-11
Rework on `veracruz-server-test`

ShaleXIONG:
Rework on the `veracruz-server-test`.
- Introduce a new `TestEvent` structure. It represents the clients' messages.
- Introduce a test structure, `TestExecutor`. It holds all relevant states, e.g. `client_connection` etc. It is responsible for spawning the proxy attestation server, and Veracruz runtime and server on the initialisation of this structure. The actual test is driven by a list of `TestEvent`, via the `execute` function. The function also takes care of clear-up the test environment and times out if the test runs too long. For usability, `TestExecutor` provide `test_template`, e.g.
```Rust
let events = vec![
     TestEvent::CheckHash,
     TestEvent::write_program(READ_FILE_WASM),
     TestEvent::write_data(STRING_1_DATA),
     TestEvent::execute(READ_FILE_WASM),
     TestEvent::read_result("/output/test/test.txt"),
     TestEvent::read_result("/output/hello-world-1.dat"),
     TestEvent::ShutDown,
];

TestExecutor::test_template(POLICY, CLIENT_CERT, CLIENT_KEY, events, Duration::from_secs(TIME_OUT_SECS)).unwrap();
```
- Rework on several busy-waiting`try_recv` patterns. The new test code uses `recv`, which blocks until a new message arrives.
- Add some `info!` and `error!` for readability.
- Fix an infinite looping problem. When a client wants to write to a `Rustls` connection buffer, i.e. `connection. writer().write(...)`, before the handshaking happens (it is done automatically via the first `wrhite_tls` call), while the content is longer than the default 64kb internal buffer.
- Remove several old test cases.
- Renames tests into three categories: `basic_***`, `intergration_***` and `performance_***`. The first category should be enough for most situations.


mathias-arm (dismissed):
The tests look much cleaner now.

dominic-mulligan-arm (approved):
Looks much better, the declarative descriptions of tests are really nice.

dominic-mulligan-arm (approved):

mathias-arm (approved):

PR #458, opened 2022-05-12
WIP: Add ML native module (inference)

gbryant-arm:
Add a ML native module.
To make things easier we are focusing on inference for now, as the API for training is a little different.
The ML framework currently used is `darknet` to ease integration with the video object detection use case and be able to measure the performance increase compared to the WASM equivalent. We might change that in the future.

The native module will take a model (weighted graph) and an array (input to the model) as input and output a prediction (result of the inference). The parameters are passed via `/services/mlinf.dat`.


gbryant-arm:
Superseded by https://github.com/veracruz-project/veracruz/pull/596 and https://github.com/veracruz-project/veracruz/pull/587

PR #460, opened 2022-05-12
Replace rustls with mbedtls in veracruz-client.

egrimley-arm:
Our forks of `rust-mbedtls` and `rust-psa-crypto` now use the same config.

### DIFF
--- a/veracruz-client/src/veracruz_client.rs
+++ b/veracruz-client/src/veracruz_client.rs
@@ -11,26 +11,104 @@
 
 use crate::error::VeracruzClientError;
 use log::{error, info};
+use mbedtls::alloc::List;
 use policy_utils::{parsers::enforce_leading_backslash, policy::Policy, Platform};
-use rustls::{Certificate, ClientConnection, PrivateKey};
-use rustls_pemfile;
 use std::{
     convert::TryFrom,
     io::{Read, Write},
     path::Path,
-    str::from_utf8,
+    sync::{Arc, Mutex},
 };
 use veracruz_utils::VERACRUZ_RUNTIME_HASH_EXTENSION_ID;
 use webpki;
 
-#[derive(Debug)]
+/// VeracruzClient struct. The remote_session_id is shared between
+/// VeracruzClient and InsecureConnection so that it is available from
+/// VeracruzClient methods and can also be updated by the
+/// InsecureConnection methods invoked by mbedtls. Although we do not
+/// expect multiple threads to be involved, since the compiler can not
+/// check this, it is safer to use a Mutex.
 pub struct VeracruzClient {
-    tls_connection: ClientConnection,
-    remote_session_id: Option<u32>,
+    tls_context: mbedtls::ssl::Context<InsecureConnection>,
+    remote_session_id: Arc<Mutex<Option<u32>>>,
     policy: Policy,
     policy_hash: String,
-    package_id: u32,
-    client_cert: String,
+}
+
+/// This is the structure given to mbedtls and used for reading and
+/// writing cyphertext, using the standard Read and Write traits.
+struct InsecureConnection {
+    read_buffer: Vec<u8>,
+    veracruz_server_url: String,
+    remote_session_id: Arc<Mutex<Option<u32>>>,
+}
+
+impl Read for InsecureConnection {
+    fn read(&mut self, data: &mut [u8]) -> Result<usize, std::io::Error> {
+        // Return as much data from the read_buffer as fits.
+        let n = std::cmp::min(data.len(), self.read_buffer.len());
+        data[0..n].clone_from_slice(&self.read_buffer[0..n]);
+        self.read_buffer = self.read_buffer[n..].to_vec();
+        Ok(n)
+    }
+}
+
+impl Write for InsecureConnection {
+    fn write(&mut self, data: &[u8]) -> Result<usize, std::io::Error> {
+        // To convert any error to a std::io error:
+        let err = |t| std::io::Error::new(std::io::ErrorKind::Other, t);
+
+        // Send all the data to the server.
+        let string_data = base64::encode(&data);
+        let combined_string = format!(
+            "{:} {:}",
+            self.remote_session_id
+                .lock()
+                .map_err(|_| err("lock failed"))?
+                .unwrap_or(0),
+            string_data
+        );
+        let dest_url = format!("http://{:}/runtime_manager", self.veracruz_server_url,);
+        // Spawn a separate thread so that we can use reqwest::blocking.
+        let body = std::thread::spawn(move || {
+            let client_build = reqwest::blocking::ClientBuilder::new()
+                .build()
+                .map_err(|_| err("reqwest new"))?;
+            let ret = client_build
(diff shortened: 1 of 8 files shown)


mathias-arm (dismissed):

veracruz-client/src/veracruz_client.rs
```diff
@@ -150,66 +220,56 @@ impl VeracruzClient {
...
-                "-----BEGIN CERTIFICATE-----",
-                "-----BEGIN CERTIFICATE-----\n",
-            )
-            .replace("-----END CERTIFICATE-----", "\n-----END CERTIFICATE-----");
+        ctx.establish(conn, None)?;
 
         Ok(VeracruzClient {
-            tls_connection: connection,
-            remote_session_id: None,
+            tls_connection: ctx,
```
dreemkiller, line 269: Would it make sense to rename `tls_connection` to `tls_context`?

veracruz-client/src/veracruz_client.rs
```diff
@@ -11,26 +11,96 @@
...
-#[derive(Debug)]
+/// VeracruzClient struct. The remote_session_id is shared between
+/// VeracruzClient and InsecureConnection so that it is available from
+/// VeracruzClient methods and can also be updated by the
+/// InsecureConnection methods invoked by mbedtls.
 pub struct VeracruzClient {
-    tls_connection: ClientConnection,
-    remote_session_id: Option<u32>,
+    tls_connection: mbedtls::ssl::Context<InsecureConnection>,
+    remote_session_id: Arc<Mutex<Option<u32>>>,
```
dreemkiller, line 31: Why is the `Arc<Mutex` now required for this?

veracruz-client/src/veracruz_client.rs
```diff
@@ -11,26 +11,96 @@
...
+                .map_err(|_| {
+                    std::io::Error::new(std::io::ErrorKind::Other, "reqwest send failed")
+                })?;
+            if ret.status() != reqwest::StatusCode::OK {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    "reqwest bad status",
+                ));
+            }
+            Ok(ret.text().unwrap())
```
dominic-mulligan-arm, line 82: When can this `unwrap` fail?

veracruz-client/src/veracruz_client.rs
```diff
@@ -11,26 +11,96 @@
...
+        })
+        .join()
+        .unwrap()?;
```
dominic-mulligan-arm, line 85: Similarly here.  We seem to be calling `unwrap` but then also propagating another error with `?`.

veracruz-client/src/veracruz_client.rs
```diff
@@ -375,195 +463,57 @@ impl VeracruzClient {
...
-        let mut outgoing_data_vec = Vec::new();
-        let mut start_index: usize = 0;
-        while start_index < data.len() {
-            let bytes_read = self.tls_connection.writer().write(&data[start_index..])?;
-            start_index += bytes_read;
-            let outgoing_data = Vec::new();
-            let outgoing_data_option = self.process(outgoing_data)?; // intentionally sending no data to process
-            if let Some(outgoing_data) = outgoing_data_option {
-                outgoing_data_vec.push(outgoing_data)
+        let cert = certs.iter().nth(0).unwrap().unwrap().iter().nth(0).unwrap();
```
dreemkiller, line 470: We should avoid `unwrap` in non-test code

veracruz-client/src/veracruz_client.rs
```diff
@@ -11,26 +11,96 @@
...
+        .unwrap()?;
+        // We received a response ...
+        let body_items = body.split_whitespace().collect::<Vec<&str>>();
+        if !body_items.is_empty() {
+            // If it was not empty, update the remote_session_id ...
+            let received_session_id = body_items[0].parse::<u32>().unwrap();
+            *self.remote_session_id.lock().unwrap() = Some(received_session_id);
+            // And append response data to the read_buffer.
+            for item in body_items.iter().skip(1) {
+                let this_body_data = base64::decode(item).unwrap();
```
dominic-mulligan-arm, line 94: (...and here.)

veracruz-client/src/veracruz_client.rs
```diff
@@ -11,26 +11,96 @@
...
+            }
+            Ok(ret.text().unwrap())
+        })
+        .join()
+        .unwrap()?;
+        // We received a response ...
+        let body_items = body.split_whitespace().collect::<Vec<&str>>();
+        if !body_items.is_empty() {
+            // If it was not empty, update the remote_session_id ...
+            let received_session_id = body_items[0].parse::<u32>().unwrap();
```
dominic-mulligan-arm, line 90: (...and here.)

veracruz-client/src/veracruz_client.rs
dominic-mulligan-arm, line 470: When can all of these `unwrap` calls fail?

dominic-mulligan-arm (changes requested):

dreemkiller (changes requested):

veracruz-client/src/veracruz_client.rs
egrimley-arm: I don't know! The documentation is dreadful.

veracruz-client/src/veracruz_client.rs
egrimley-arm: The comment at the start of the function is supposed to explain that. The function is being called via a C library so if I return an error, it just disappears, as far as I can tell.

veracruz-client/src/veracruz_client.rs
egrimley-arm: Ditto. A call to `write` doesn't correspond very naturally to a particular call into the C library, I think.

veracruz-client/src/veracruz_client.rs
egrimley-arm: Probably. I'll do that.

veracruz-client/src/veracruz_client.rs
egrimley-arm: What's the best alternative?

veracruz-client/src/veracruz_client.rs
egrimley-arm: The two functions that update the `remote_session_id` aren't really going to be in separate threads, I think, but because one is called via C there is no way for the Rust compiler to do any kind of memory check, so I have to choose between `unsafe` and `Arc<Mutex>`. Since it's not performance-critical, I went for the latter.

dominic-mulligan-arm (approved):
This looks good to me now, but let's wait to make sure @dreemkiller is also happy with the changes, too.

dominic-mulligan-arm:
(Sorry, @mathias-arm not @dreemkiller.  I was thinking of the wrong PR.)

dreemkiller (approved):

dominic-mulligan-arm:
+1+1 = +1 reached, merging.

PR #465, opened 2022-05-25
Improved error msg when a certificate is not found

federicobozzini:
When generating policies using `vc-pgen` if a certificate is not found this error was generated:

> Could not open certificate file.

This PR adds the certificate path to the error message. The result is something like:

> Could not open certificate file "example/example-program-cert.pem".

### DIFF
--- a/test-collateral/generate-policy/src/main.rs
+++ b/test-collateral/generate-policy/src/main.rs
@@ -632,7 +632,8 @@ fn serialize_identities(arguments: &Arguments) -> Vec<Identity<String>> {
 
             values.push(Identity::new(certificates, id as u32, file_permissions));
         } else {
-            abort_with("Could not open certificate file.");
+            let error_msg = format!("Could not open certificate file {:?}.", cert);
+            abort_with(error_msg);
         }
     }
     values


dreemkiller (approved):

dreemkiller:
The change looks good, but it can't go in until it's been signed.
Don't forget to set up your signing key and sign the commit. Then it can be approved again.
Here's some instructions: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

test-collateral/generate-policy/src/main.rs
```diff
@@ -632,7 +632,8 @@ fn serialize_identities(arguments: &Arguments) -> Vec<Identity<String>> {
 
             values.push(Identity::new(certificates, id as u32, file_permissions));
         } else {
-            abort_with("Could not open certificate file.");
+            let error_msg = format!("Could not open certificate file {:?}.", cert);
+            abort_with(error_msg);
```
dominic-mulligan-arm, line 636: You can write this just as

```rust
abort_with(format!("Could not open certificate file {:?}.", cert));
```

federicobozzini:
Thanks @dreemkiller . I've added a signature to my commit.

test-collateral/generate-policy/src/main.rs
federicobozzini: I wanted to be more verbose but I'm happy to change it if you prefer.

dominic-mulligan-arm (approved):

dominic-mulligan-arm:
+1 reached, merging.

PR #467, opened 2022-05-25
Make veracruz_server_linux a bit more reliable.

egrimley-arm:
* Pick port number at random.
* Ignore SIGCHLD to avoid zombies.
* Kill runtime manager if there is an error.

### DIFF
--- a/veracruz-server/Cargo.toml
+++ b/veracruz-server/Cargo.toml
@@ -63,6 +63,7 @@ nix = { version = "0.20.2", optional = true }
 postcard = "0.7.2"
 policy-utils = { path = "../policy-utils", optional = true }
 psa-attestation = { path = "../psa-attestation", optional = true }
+rand = "0.8.3"
 rustls = "0.20.4"
 serde = { version = "1.0.115", default-features = false, features = ["derive"] }
 serde_json = "1.0"
--- a/veracruz-server/Cargo.toml
+++ b/veracruz-server/Cargo.toml
@@ -33,6 +33,7 @@ icecap-qemu = []
 linux = [
   "data-encoding",
   "io-utils/linux",
+  "nix",
   "policy-utils/std",
   "tempfile",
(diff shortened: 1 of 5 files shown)


veracruz-server/Cargo.toml
```diff
@@ -63,6 +64,7 @@ nix = { version = "0.20.2", optional = true }
 postcard = "0.7.2"
 policy-utils = { path = "../policy-utils", optional = true }
 psa-attestation = { path = "../psa-attestation", optional = true }
+rand = "0.7"
```
mathias-arm, line 67: Could you use `rand = "0.8.3"` (used in the rest of Veracruz) to avoid duplication?

mathias-arm (changes requested):

veracruz-server/src/veracruz_server_linux.rs
```diff
@@ -270,14 +274,34 @@ pub mod veracruz_server_linux {
...
+            unsafe {
+                signal::sigaction(
+                    signal::Signal::SIGCHLD,
+                    &signal::SigAction::new(
+                        signal::SigHandler::SigIgn,
+                        signal::SaFlags::empty(),
+                        signal::SigSet::empty(),
+                    ),
+                )
+                .unwrap();
```
dominic-mulligan-arm, line 298: When can this fail?

egrimley-arm: I've changed this to `expect`.

veracruz-server/Cargo.toml
egrimley-arm: Done.

dominic-mulligan-arm (approved):

mathias-arm (approved):

dominic-mulligan-arm:
+1+1 = +1 reached, merging.

PR #472, opened 2022-06-10
Finish replacing rustls with mbedtls.

egrimley-arm:

PR #476, opened 2022-06-15
Remove ring and webpki from Veracruz.

egrimley-arm:
This is based on #472, which should be reviewed and merged first.

### DIFF
--- a/veracruz-server-test/src/main.rs
+++ b/veracruz-server-test/src/main.rs
@@ -26,7 +26,6 @@ mod tests {
     use policy_utils::{policy::Policy, Platform};
     use proxy_attestation_server;
     use std::{
-        convert::TryFrom,
         env::{self, VarError},
         error::Error,
         fs::File,
@@ -1180,33 +1179,18 @@ mod tests {
                 .iter()
                 .nth(0)
                 .ok_or("unexpected certificate")?;
-            let ee_cert = webpki::EndEntityCert::try_from(cert.as_der())?;
-            let ues = ee_cert.unrecognized_extensions();
-
+            let extensions = cert.extensions()?;
             // check for OUR extension
-            let encoded_extension_id: [u8; 3] = [
-                VERACRUZ_RUNTIME_HASH_EXTENSION_ID[0] * 40 + VERACRUZ_RUNTIME_HASH_EXTENSION_ID[1],
-                VERACRUZ_RUNTIME_HASH_EXTENSION_ID[2],
-                VERACRUZ_RUNTIME_HASH_EXTENSION_ID[3],
-            ];
-            let data = ues
-                .get(&encoded_extension_id[..])
-                .ok_or(format!("Our certificate extension is not present."))?;
-            info!("Certificate extension found.");
-
-            let extension_data = data
-                .read_all(format!("Can't read veracruz custom extension."), |input| {
-                    Ok(input.read_bytes_to_end())
-                })?;
-
-            if compare_policy_hash(
-                extension_data.as_slice_less_safe(),
-                &self.policy,
-                &target_platform,
-            ) {
-                Ok(())
-            } else {
-                Err(format!("None of the runtime manager hashes matched.").into())
+            match veracruz_utils::find_extension(extensions, &VERACRUZ_RUNTIME_HASH_EXTENSION_ID) {
+                None => Err(format!("Our certificate extension is not present.").into()),
+                Some(data) => {
+                    info!("Certificate extension found.");
+                    if compare_policy_hash(&data, &self.policy, &target_platform) {
+                        Ok(())
+                    } else {
+                        Err(format!("None of the runtime manager hashes matched.").into())
+                    }
+                }
             }
         }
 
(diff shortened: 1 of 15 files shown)


egrimley-arm:
I have now tested this on Nitro.

ShaleXIONG (commented):
Some `Cargo.lock` in workspaces. Should we keep those files?

egrimley-arm:
> Some `Cargo.lock` in workspaces. Should we keep those files?

Presumably that's a thought not directly related to this PR?

The `Cargo.lock` files were added on 10 Mar 2022 by Mathias Brossard. If we didn't have the `Cargo.lock` files we'd need precise version numbers in `Cargo.toml` or builds would become very non-reproducible!

ShaleXIONG (approved):

ShaleXIONG:
> > Some `Cargo.lock` in workspaces. Should we keep those files?
> 
> Presumably that's a thought not directly related to this PR?
> 
> The `Cargo.lock` files were added on 10 Mar 2022 by Mathias Brossard. If we didn't have the `Cargo.lock` files we'd need precise version numbers in `Cargo.toml` or builds would become very non-reproducible!

Sorry, my bad, Github folds the difference. I thought it has been deleted from the repo.

policy-utils/src/error.rs
```diff
@@ -11,8 +11,8 @@
 
 use err_derive::Error;
 #[cfg(feature = "std")]
-use rustls::{CipherSuite, Error as TLSError};
-use std::{string::String, time::SystemTimeError};
+use std::string::String;
+use std::time::SystemTimeError;
```
dreemkiller, line 15: When feature "std" is not enabled, `std::time::SystemTimeError` will still be `use`d. Is this what you intended?

session-manager/src/session.rs
```diff
@@ -12,63 +12,130 @@
...
-
 ////////////////////////////////////////////////////////////////////////////////
 // Sessions.
 ////////////////////////////////////////////////////////////////////////////////
 
 /// A principal is an individual identified with a cryptographic certificate,
 /// and assigned a set of roles that dictate what that principal can and cannot
 /// do in a Veracruz computation.
-pub type Principal = Identity<Certificate>;
+pub type Principal = Identity<List<mbedtls::x509::Certificate>>;
```
dreemkiller, line 31: Do we want to `use mbedtls::x509::Certificate` at the top of the file to make this simpler?

For me, it can go either way. However, there has been the trend within the project (and I think it's a preference within the rust community) to use a `use` statement instead of having long lists of modules, submodules, etc. within the code.

Again, this is a question, not a demand, because for me, it can go either way

session-manager/src/session.rs
```diff
@@ -12,63 +12,130 @@
...
 }
 
+struct Buffers {
+    /// Read buffer used by mbedtls for cyphertext.
+    read_buffer: Vec<u8>,
+    /// Write buffer used by mbedtls for cyphertext.
+    write_buffer: Option<Vec<u8>>,
+}
+
+struct InsecureConnection {
```
dreemkiller, line 54: What's the purpose of the `InsecureConnection` struct?

veracruz-server-test/src/main.rs
```diff
@@ -843,6 +842,63 @@ mod tests {
...
     }
 
+    struct Buffers {
+        // Read buffer used by mbedtls for cyphertext.
+        read_buffer: Vec<u8>,
+        // Write buffer used by mbedtls for cyphertext.
+        write_buffer: Option<Vec<u8>>,
+    }
+
+    struct InsecureConnection {
```
dreemkiller, line 852: Is this a duplicate of the previous `InsecureConnection` declaration? Is it necessary to duplicate it here?

dreemkiller (commented):
With a change like this, the proof is in the pudding as they say. If the tests pass (on all platforms) then I'm good with it. I've added a few comments (questions mostly, please don't take them as "must do"s).

policy-utils/src/error.rs
egrimley-arm: This is a bit messed up. I'll fix it.

session-manager/src/session.rs
egrimley-arm: I think that's worth changing. I'll do that.

session-manager/src/session.rs
egrimley-arm: I'm unsure of the correct Rust terminology here but I'll add a comment like this:

egrimley-arm: No, I see that another file already has a better explanation, which I can adapt:
```
/// This is the structure given to mbedtls and used for reading and
/// writing cyphertext, using the standard Read and Write traits.
```

veracruz-server-test/src/main.rs
egrimley-arm: I think it's a separate thing, really.

egrimley-arm:
I've modified #472. Perhaps someone could approve that first.

veracruz-server-test/src/main.rs
dreemkiller: Would it make sense to rename it to prevent confusion?

egrimley-arm: Each place where I use `mbedtls` there's a similar set-up with `Buffers` and `InsecureConnection` for interfacing with `mbedtls`, but they're not quite identical and they're file-local.

dreemkiller (approved):

Issue #482, opened 2022-06-23
Code base structure and documentation


ShaleXIONG:
The codebase structure is complex and becomes a barrier to maintenance:
- Directories `sdk`, `test-collateral` and `*-test` contain a mix of code and artefact.
- There is no clear sign of examples and dev-tool, and the readme is not very helpful.
- more readable policy file.
- Complicated Error enum.
- Large cargo dependency in certain components.

ShaleXIONG:
#487 #489 #490 #491 #493

PR #483, opened 2022-06-27
Rework on the structure and quality of the code base 

ShaleXIONG:
Rework the structure of the code base:
- Create a new sub-crate `tests` and move the `veracruz-test` and `veracruz-server-test` to `tests/tests/intergrate_test.rs` and `tests/tests/server_test.rs`. The new `tests` crate also contains some common tests code used by the two tests. Rework the structure of the old `veracruz-test`, or the new `intergrate_test.rs`.
- Move examples and data-generators to a new `examples` directory and move the `policy-generator` from the directory `test-collateral` to `sdk`.
- Remove unused functions and rename some functions in `veracruz-client` and `transport-protocol`. Rework several functions in `veracruz-client` too.

This merge is for #482 and only solve some problems in #482.


mathias-arm (dismissed):

examples/rust-examples/sort-numbers/src/main.rs
```diff
@@ -17,7 +17,7 @@
 use std::fs;
 use anyhow;
 
-const INPUT_FILENAME: &'static str = "/input/unsorted_numbers.txt";
+const INPUT_FILENAME: &'static str = "/output/unsorted_numbers.txt";
```
dreemkiller, line 20: are we sure this is correct?

dreemkiller (commented):
Does this change the workflow for Veracruz development on any platform?

For example, currently, to run Linux server tests, I run
`make test-server` from within the `workspaces/linux-host` directory.
Does this change that? If so, we need to send out a notice to all developers so they are aware of the change before they pull latest.

mathias-arm:
> Does this change the workflow for Veracruz development on any platform?
> For example, currently, to run Linux server tests, I run `make test-server` from within the `workspaces/linux-host` 

From what I understand of the changes in [workspaces/linux-host/Makefile](https://github.com/veracruz-project/veracruz/pull/483/files#diff-7f2298792853649ab298d648e9e9076238baf0c83c1de9786c830e16532ecbc4) the makefile targets have been updated.

examples/rust-examples/sort-numbers/src/main.rs
ShaleXIONG: Yes. This program will read another wasm program's result. The test case here is to execute two wasm binaries, `random_u32_list.wasm` and `sort_number.wasm`. The former will general a list of u32, and the latter, which is this program, read the result and sort it.

The test case is `basic_pipeline` in `server_tests.rs`.

ShaleXIONG:
> Does this change the workflow for Veracruz development on any platform?
> 
> For example, currently, to run Linux server tests, I run `make test-server` from within the `workspaces/linux-host` directory. Does this change that? If so, we need to send out a notice to all developers so they are aware of the change before they pull latest.

All high-level targets in `Makefile` in `workspaces` directory remains the same.

mathias-arm (approved):

PR #485, opened 2022-06-27
Remove the two-second sleep from veracruz_server_nitro.rs.

egrimley-arm:
The sleep seems to be unnecessary: it happens after the connect (in `NitroEnclave::new` in `io-utils/src/nitro.rs`) and adding a longer sleep at the start of `nitro_main` in `runtime_manager_nitro.rs` does not make the tests fail.

Also bracket the call to `connect` with calls to `alarm::set` and `alarm::cancel` so we do not wait for ever for the enclave to start.

### DIFF
--- a/io-utils/src/nitro.rs
+++ b/io-utils/src/nitro.rs
@@ -10,6 +10,7 @@
 //! information on licensing and copyright.
 
 use err_derive::Error;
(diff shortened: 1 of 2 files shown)


dreemkiller (approved):
If this works, then it's all good

PR #486, opened 2022-06-29
WIP: Revamped the way we generate PSA Attestation tokens to make it easier…

dreemkiller:
… to maintain

The previous method was cribbed from the TF-M project, which targets micro controllers and had some constructs necessary for those platforms to save space by that made it harder to maintain.


egrimley-arm:
This PR uses `psa-crypto`, while #501 remove `psa-crypto`!

PR #487, opened 2022-07-01
Rework on the error handler in veracruz-client

ShaleXIONG:
Use anyhow as the error handler in `veracruz-client` and remove several async blocks.

### DIFF
--- a/tests/tests/integration_test.rs
+++ b/tests/tests/integration_test.rs
@@ -281,8 +281,7 @@ async fn veracruz_phase4_linear_regression_two_clients_parallel() {
             let program_data = read_local_file(prog_path).unwrap();
             info!("### program provider send binary.");
             client
-                .write_file("/program/linear-regression.wasm", &program_data)
-                .await?;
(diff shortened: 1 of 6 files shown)


tests/tests/integration_test.rs
```diff
@@ -331,8 +328,8 @@ async fn server_tls_loop<P: AsRef<str>>(policy_json: P) -> Result<()> {
 
 /// Test states.
 struct TestExecutor {
-    // The policy for the runtime.
-    policy: Policy,
+    //// The policy for the runtime.
+    //policy: Policy,
```
mathias-arm, line 332: Should you remove it?

tests/tests/integration_test.rs
```diff
@@ -372,7 +369,7 @@ impl TestExecutor {
         proxy_attestation_setup(policy.proxy_attestation_server_url().clone());
 
         Ok(TestExecutor {
-            policy,
+            //policy,
```
mathias-arm, line 372: Should you remove it?

mathias-arm (dismissed):

PR #491, opened 2022-07-14
Rework error handling in both veracruz and proxy servers, transport protocol, and session manager. 

ShaleXIONG:
Rework the error handling and codebase.
- Use `anyhow` for handling errors in `session` and `transport-protocol`.
- Remove the unused error cases in `veracruz-server` and `proxy-server`. Note that these two crates still use customed Error types as they implement `actix` error message.
- Change the code structure of `veracruz-server`.


gbryant-arm (approved):
LGTM

PR #493, opened 2022-07-21
Generate keys and certificates in CI.

ShaleXIONG:
Generate keys and certificates in CI.

Minor:
- Remove a few unused SGX- and TZ-related artefacts.
- Remove unused Error cases.
- Remove unused protocol buffer types.
- Update VC client due to the change of the protocol buffer.


egrimley-arm:
As we discussed in the meeting today, it might be good to replace the RSA keys with ECDSA keys at the same time. Replacing `openssl genrsa 2048` with `openssl ecparam -name secp521r1 -genkey` might work, for example.

egrimley-arm:
The current commit (506a477c7a25a53efed6e9c130196f1cc2434191) passed on CI but `veracruz_phase2_intersection_set_sum_three_clients` seems to fail on my machine.

egrimley-arm:
Another thing: changes to `Cargo.lock` files do not seem to have been committed.

egrimley-arm:
When I rebased this onto `main`, there were no conflicts, but some of the tests failed. Have anyone else tried that?

egrimley-arm:
The problem seems to be with the way the `CAKey` and `CACert` are generated: if I regenerate those two files on `main` with the recipe used in this PR, then the tests fail.

egrimley-arm:
I think you should probably rebase this PR onto `main` and force-push that, so that we see the CI failure here.

Then replace `secp256k1` with `prime256v1` in `workspaces/shared.mk` and push again: with any luck that will pass. (I'm testing locally now.)

egrimley-arm:
What are the changes in `vc.c`? There doesn't seem to be any obvious link with the description of this PR or the corresponding commit message.

ShaleXIONG:
> What are the changes in `vc.c`? There doesn't seem to be any obvious link with the description of this PR or the corresponding commit message.

I updated the PR message now. The `vc.c` should be changed since we change the structure of the protocol buffer.

egrimley-arm (approved):

PR #494, opened 2022-07-22
Switch from TLS 1.2 to TLS 1.3.

egrimley-arm:


dreemkiller (dismissed):

veracruz-utils/src/csr.rs
```diff
@@ -41,12 +41,13 @@ pub enum CertError {
...
 
 pub fn generate_csr(private_key_der: &[u8]) -> Result<Vec<u8>> {
-    let mut pk_private = mbedtls::pk::Pk::from_private_key(private_key_der, None).unwrap();
     let mut rng = |buffer: *mut u8, size: usize| {
         let mut slice = unsafe { std::slice::from_raw_parts_mut(buffer, size) };
         getrandom(&mut slice);
         0
     };
+    let mut pk_private =
+        mbedtls::pk::Pk::from_private_key(&mut rng, private_key_der, None).unwrap();
```
gbryant-arm, line 50: Could you introduce a new error message instead of unwrapping?

gbryant-arm (changes requested):

veracruz-utils/src/csr.rs
egrimley-arm: Done.

gbryant-arm (approved):
LGTM

PR #497, opened 2022-08-01
Update Mbed TLS to mbedtls-3.2.1.

egrimley-arm:
This is now possible because we are using `prime256v1` instead of RSA.


egrimley-arm:
I've tested 2d335595c226cd526ae18f360dedd42bf5b6b5b5 (current version of this) on Nitro.

ShaleXIONG (approved):

PR #498, opened 2022-08-02
Replace use of x509-parser with mbedtls.

egrimley-arm:


veracruz-client/src/veracruz_client.rs
```diff
@@ -172,21 +171,32 @@ impl VeracruzClient {
...
-            Err(anyhow!(VeracruzClientError::UnexpectedKey))
-        } else if parsed_cert.validity.time_to_expiration().is_none() {
-            Err(anyhow!(VeracruzClientError::UnexpectedCertificate))
-        } else {
-            Ok(())
+            return Err(anyhow!(VeracruzClientError::UnexpectedKey));
+        }
+
+        // Check validity period.
+        #[cfg(features = "std")]
```
dreemkiller, line 187: Do we need "std" checks in the Client?

veracruz-utils/src/csr.rs
```diff
@@ -67,27 +67,26 @@ pub fn generate_csr(private_key_der: &[u8]) -> Result<Vec<u8>> {
...
+        now.year(),
+        now.month(),
+        now.day(),
+        now.hour(),
+        now.minute(),
+        now.second()
     );
-    return Ok(generated_time.as_bytes().to_vec());
+    if r.is_err() || cursor.position() != 14 {
+        panic!("bad time")
```
dreemkiller, line 89: Is a panic the right thing to do here? or return an error?

veracruz-client/src/veracruz_client.rs
egrimley-arm: I found we did, though I didn't investigate why.

veracruz-utils/src/csr.rs
egrimley-arm: Panic is the right thing to do, I think. This can only fail if the year is negative, or greater than 9999, which could theoretically happen but I don't think we care about handling that nicely, or if the minutes are greater than 99, or something like that, which is an internal error.

veracruz-client/src/veracruz_client.rs
egrimley-arm: I would have thought it's best to leave it in for now. We'll probably never need the client under IceCap, but it's working now, so why break it?

dreemkiller: Good point

veracruz-utils/src/csr.rs
egrimley-arm: I think the code has "year 10000" problems in lots of places. I think the RFCs prescribe four-digit years.

dreemkiller: I think we can leave that problem for our supersmart descendants. If they aren't supersmart, then they will not be using computers by then, and they won't notice that we took this shortcut.

dreemkiller (approved):

PR #501, opened 2022-08-04
Remove rust-psa-crypto; use rust-mbedtls instead.

egrimley-arm:
- execution-engine: Use mbedtls::cipher.
- psa-attestation: C code refers to mbedtls-sys instead.
- third-party/rust-mbedtls: Various changes.
- third-party/rust-psa-crypto: Removed.
- veracruz-utils: Use mbedtls::hash.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -27,9 +27,6 @@
 	path = third-party/wasmi
 	url = https://github.com/veracruz-project/wasmi.git
 	branch = veracruz
-[submodule "third-party/rust-psa-crypto"]
-	path = third-party/rust-psa-crypto
-	url = https://github.com/veracruz-project/rust-psa-crypto.git
 [submodule "workspaces/icecap-runtime/deps/seL4"]
 	path = workspaces/icecap-runtime/deps/seL4
 	url = https://gitlab.com/arm-research/security/icecap/seL4.git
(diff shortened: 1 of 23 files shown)


egrimley-arm:
The tests passed at https://github.com/veracruz-project/veracruz/actions/runs/2799377318

But I will try undoing some of the changes that perhaps were not necessary ...

egrimley-arm:
@gbryant-arm, it would be good if you could test this on Nitro, though I don't think this PR will help with the entropy problem you're currently working on.

@ShaleXIONG , it would be good if you could test the code in `aead.rs` and `aes.rs`.

gbryant-arm (changes requested):
There still are references to `rust-psa-crypto` in some toml files under `workspace`

gbryant-arm:
`rust-mbedtls` doesn't get recompiled when making changes to `mbedtls_hardware_poll.c`:
I suggest we add the following code after https://github.com/veracruz-project/rust-mbedtls/blob/c9df29d0016de1d67ce597d6669e1ebda12dfb3d/mbedtls-sys/build/build.rs#L108:
```
// hardware poll                                                        
println!(                                                               
   "cargo:rerun-if-changed={}",                                        
   self.mbedtls_src                                                    
       .join(Path::new("..").join("build").join("mbedtls_hardware_poll.c"))
       .display()                                                      
); 
```

gbryant-arm:
Passes end to end with the tiny YOLO model :+1: 
It doesn't pass with the big model but that's orthogonal to this PR

egrimley-arm:
@gbryant-arm, I've fixed those two problems you pointed out and rebased.

gbryant-arm (approved):
Runs e2e on Nitro :heavy_check_mark:

PR #504, opened 2022-08-11
Strip Nitro binaries before generating the EIF file

gbryant-arm:
This generates much smaller EIF files with no additional compilation cost.
Stripping symbols is not a limitation here, as there is no debugger in Nitro.


dreemkiller (approved):
nice

Issue #507, opened 2022-08-12
Entropy problem on Nitro


gbryant-arm:
**Describe the bug**
On some Nitro machines, requesting entropy by whether reading `/dev/random` or sending a `getrandom` syscall results in failure: the function call hangs and does not return immediately (or ever). This might be due to entropy shortage or incorrect entropy initialisation.
Passing `GRND_RANDOM` or `GRND_NONBLOCK` to `getrandom()` doesn't seem to make any difference.
Reading from `/dev/urandom` works.

**To Reproduce**
Call libc's (musl's) `getrandom()` or read `/dev/random` inside Nitro and wait for the return... forever (doesn't happen on every Nitro machine)

**Expected behaviour**
`getrandom()` or `/dev/random` should promptly return random bytes.

**Potential solutions**

##### 1. Use the NSM library
On Nitro, it is recommended to use AWS' NSM library for RNG.
The functions of interest are `nsm_get_random()` and `aws_nitro_enclaves_library_seed_entropy()`, the latter calling `nsm_get_random()` and seeding `/dev/random` with the returned random bytes.
It's curious that `/dev/random` is not seeded automatically by the kernel whenever the `getrandom` syscall is received.
Several approaches can be considered to redirect external dependencies to the right RNG:
* Use Veracruz's `platform-services`: In Veracruz, `platform-services` exposes `getrandom()` that calls functions from the NSM lib. WASM programs calling `getrandom()` are automatically redirected to it, but linking external dependencies, especially C code, against it is trickier (cf. #511)
* Mbed TLS specific: Pass a pointer to a custom RNG when initialising mbedtls (`mbedtls_ssl_conf_rng()`)

##### 2. Read RN from `/dev/urandom` (quick workaround)

dreemkiller:
What are the properties of the instances that are experiencing the failures?
Have they been up a long time? Because originally, `nsm` was not plumbed into `/dev/random`. Is it possible that the failures are happening on instances launched before that change was made?

dreemkiller:
Here are some thoughts about `/dev/random` vs `/dev/urandom`: https://www.2uo.de/myths-about-urandom/

TL/DR: `/dev/urandom` should be fine.

gbryant-arm:
> What are the properties of the instances that are experiencing the failures? Have they been up a long time? Because originally, `nsm` was not plumbed into `/dev/random`. Is it possible that the failures are happening on instances launched before that change was made?

[...]

PR #509, opened 2022-08-15
Use mbedtls fake_random feature on Nitro.

egrimley-arm:
I have probably enabled `fake_random` for everything built for Nitro, not just for things that go into the enclave. I am not sure what the best way is to enable it just for the enclave.

Fix for https://github.com/veracruz-project/veracruz/issues/507, following up on https://github.com/veracruz-project/rust-mbedtls/pull/3

### DIFF
--- a/execution-engine/Cargo.toml
+++ b/execution-engine/Cargo.toml
@@ -12,6 +12,8 @@ icecap = [
   "policy-utils/icecap",
 ]
 nitro = [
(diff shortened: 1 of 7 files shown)


gbryant-arm:
Looks good. I'll try that e2e.
I don't how to apply that to the enclave only. Maybe some cargo magic in `workspaces/nitro-runtime/Cargo.toml`?

PR #513, opened 2022-08-19
DSL for Executing programs.

ShaleXIONG:
Features:
Add DSL for executing several programs. The DSL contains standard sequential composition, `program_path_1 ; program_path_2 ;` and conditional composition, `if conditional_file { ... } else { ... }`. The conditional composition checks if the `conditional_file` exists and either executes the `true` branch if exists, or `false` branch if there is a `false` branch. An example program is:
```
/programs/pre_process.wasm ;
if /output/after_pre_process.txt 
{ /programs/ml.wasm ; }
else 
{ /programs/produce_error_message.wasm ; }
```

Applications:
- In an ML application where the algorithm remains confidential, the program provider is able to cut the entire program into several smaller public programs and confidential programs. Then use this DSL to define the correct execution order.
- The conditional composition allows the program to check if a native module exists, and calls the native module or executes the backup wasm program.

Changes:
We use `lalrpop` to define and interpret the script. To enable the new way to execute programs, there are the following technique changes
- Add a new `pipelines` field in the policy. It defines allowed pipelines and their rights.
- Add a new protocolbuf for calling pipelines based on the `pipeline_id`.
- Add several internal functions in the `runtime-manager` for pipelines and rework the old `executing` programs interface. It is unified to `execute` pipeline.
- Add a test case in the `server tests`.

Minors:
- Rework on policy-generator. A valid policy structure is created by a `new` method call, rather than `new` followed by a separate `init` call in the previous case.


execution-engine/src/engines/common.rs
```diff
@@ -803,16 +814,14 @@ pub struct WasiWrapper {
...
-    pub(crate) environment_variables: Vec<(String, String)>,
-    /// The array of program arguments that have been passed to this program,
-    /// again from the global policy file.
-    pub(crate) program_arguments: Vec<String>,
-    /// The exit code, if program calls proc_exit.
+    /// The exit code returned by the last executing program.
     exit_code: Option<u32>,
-    /// Whether clock functions (`clock_getres()`, `clock_gettime()`) should be enabled.
+    /// The environment variables currently set, and their bindings.
+    environment_variables: Vec<(String, String)>,
```
dreemkiller, line 820: Why were these two (`environment_variables` and `program_arguments`) moved? It makes the diff hard to read.

execution-engine/src/engines/wasmi.rs
```diff
@@ -86,11 +86,7 @@ impl MemoryHandler for MemoryRef {
     type Slice = &'static [u8];
     type SliceMut = &'static mut [u8];
 
-    fn get_slice<'a>(
-        &'a self,
-        address: u32,
-        length: u32,
-    ) -> FileSystemResult<Bound<'a, Self::Slice>> {
+    fn get_slice(&self, address: u32, length: u32) -> FileSystemResult<Bound<Self::Slice>> {
```
dreemkiller, line 89: Which format for this function prototype is more "rusty"? I definitely prefer the new one, but do the rust tools have an opinion on this?

execution-engine/src/engines/wasmi.rs
```diff
@@ -106,11 +102,11 @@ impl MemoryHandler for MemoryRef {
         })))
     }
 
-    fn get_slice_mut<'a>(
-        &'a mut self,
```
dreemkiller, line 110: What change resulted in the ability for us not to use lifetimes here?

policy-utils/src/principal.rs
```diff
@@ -31,6 +32,8 @@ pub enum Principal {
     Participant(u64),
     /// Program in Veracruz, indentified by the program file name.
     Program(String),
+    /// Pipeline in Veracruz, indentified by the pipeline name.
+    Pipeline(String),
```
dreemkiller, line 36: So the `Pipeline` operates as it's own principal?

sdk/generate-policy/src/main.rs
```diff
@@ -106,31 +89,37 @@ struct Arguments {
...
-    proxy_attestation_server_ip: Option<SocketAddr>,
+    proxy_attestation_server_ip: SocketAddr,
     /// The filename of the Proxy Attestation Service certificate
     proxy_service_cert: PathBuf,
     /// The filename of the Runtime Manager CSS file for SGX measurement.  This is
     /// optional.
     css_file: Option<PathBuf>,
     /// The filename of the Runtime Manager PRCR0 file for Nitro Enclave
     /// measurement.  This is optional.
-    pcr0_file: Option<PathBuf>,
```
dreemkiller, line 120: `pcr0_file` is the correct name

tests/tests/server_test.rs
```diff
@@ -280,8 +280,7 @@ fn basic_pipeline() {
     let events = vec![
```
dreemkiller, line 280: Have any tests been added that use this new feature?

dreemkiller (changes requested):
Mostly a lot of questions with no required changes. Only one change (in my opinion) needs to be made: `pcr0_file` is the correct name, as the file contains the value of PCR0, and none of the other PCR values.

execution-engine/src/engines/wasmi.rs
ShaleXIONG: It is the `cargo fmt`. I believe they make it one line since the parameter list is not too long.

execution-engine/src/engines/wasmi.rs
ShaleXIONG: The compiler warns us to remove it. I believe it probably automatically adds it now.

tests/tests/server_test.rs
ShaleXIONG: I reworked the existing test `execute_pipeline`. The test called two programs in two distinct executing program requests, but now calls an executing pipeline request. The pipeline has been statically defined in the policy.

policy-utils/src/principal.rs
ShaleXIONG: Yes. It is to define a separate right/permission for the entire pipeline. It will ignore programs' individual rights in the pipeline.

Further design is considered, for example, an intersection of pipeline and individual rights.

execution-engine/src/engines/common.rs
ShaleXIONG: I moved back to the original order. The git diff is still confused about the change though.

dreemkiller (approved):

PR #515, opened 2022-08-23
Implement timer for `getrandom()`

### LINKED ISSUE #507: Entropy problem on Nitro

gbryant-arm:
Since #507 hasn't been reproduced yet, we propose to print an error message whenever #507 shows up again. This shall be a signal to open a ticket on AWS.
TLDR: Detect but don't fix #507


gbryant-arm:
Of course, icecap doesn't support threads... Replacing multithreading with signal handling

gbryant-arm:
icecap doesn't support signal handling either?

gbryant-arm:
Ready for review

egrimley-arm:
Why is `entropy_poll.c` modified? Why can't everything be done in `mbedtls_hardware_poll.c`?

egrimley-arm:
Previously, the contents of `mbedtls-sys/vendor` were identical to upstream `mbedtls-3.2.1`. Can we please keep it identical and just modify the wrapper rather than Mbed TLS itself?

egrimley-arm:
Also, can we please keep the git history of `rust-mbedtls` nice and simple? It could be a single commit for this PR, surely?

gbryant-arm:
> Why is `entropy_poll.c` modified? Why can't everything be done in `mbedtls_hardware_poll.c`?

I've simply greped every call to `getrandom()`. If you're certain this file is not used then I'll remove it.

> Previously, the contents of mbedtls-sys/vendor were identical to upstream mbedtls-3.2.1. Can we please keep it identical and just modify the wrapper rather than Mbed TLS itself?

I guess you're referring to `entropy_poll.c`. Cf. my answer above.

> Also, can we please keep the git history of rust-mbedtls nice and simple? It could be a single commit for this PR, surely?

Sure, let's pack the rust-mbedtls updates together

egrimley-arm:
> If you're certain this file is not used then I'll remove it.

It's easy enough to check. The configuration is not dynamically generated, so if the code isn't used now, it won't be used in future.

egrimley-arm:
We define `MBEDTLS_NO_PLATFORM_ENTROPY` so `mbedtls_platform_entropy_poll` does not get compiled on any of our platforms.

egrimley-arm:
The diff looks good, but can we simplify the changes to `rust-mbedtls` to a single commit?

egrimley-arm (approved):

mathias-arm:
The `gbryant-arm/getrandom-timer` branch in [rust-mbedtls](https://github.com/veracruz-project/rust-mbedtls) should have been merged into `veracruz` before this PR was merged. Opening veracruz-project/rust-mbedtls#6 (and closing veracruz-project/rust-mbedtls#3).

PR #516, opened 2022-09-02
Update rust-toolchain, wasmtime and openssl-src.

egrimley-arm:
This solves some Dependabot alerts. A newer `wasmtime` requires a newer Rust toolchain, and changing the Rust toolchain meant updating `icecap/sysroot/rust`.


mathias-arm (approved):
Great work!

Two non-blocking comments:
- Looking at the `Cargo.lock` files, I saw there are now in some cases two versions of the `object` crates, 0.27.1 and 0.28.4. The 0.27.1 version is only used by `backtrace` version 0.3.64, so if we able to update to 0.3.65 we will avoid that duplication.
- The update to Rust version 1.60.0 is not reflected in the Docker configuration files (we preload Rust 1.56.1).

egrimley-arm:
Thanks for those comments. I think it would be very helpful to have all the Rust version updates in a single commit, for someone to refer to next time we update it, so I've modified the PR to include that. I'll leave the other thing for a separate PR. (I suspect there may be other crates that are unnecessarily included with several versions. It might be worth checking with a script, unless you've already done that?)

mathias-arm:
You can use `cargo tree -d -e no-build` in the different workspaces to find duplicates. Some duplicates would be very difficult to fix.

egrimley-arm:
I have now also upgraded `backtrace`, as you suggested.

mathias-arm (approved):

Issue #518, opened 2022-09-20
Compilation artifacts are not reproducible due to everchanging build ID


gbryant-arm:
**Describe the bug**
Veracruz binaries (client, server, attestation, runtime manager) contain a build ID (`.note.gnu.build-id` section) in their ELF headers. This build ID is a hash determined by rustc and/or the linker.
However, for unknown reasons (timestamp somewhere? source of randomness?), the build ID changes every time the crate is cargo cleaned then rebuilt.
This results in binaries that are functionally equivalent but have different overall hashes, which gives different Docker images.
FWIW some debug sections (e.g. .debug_info, .debug_abbrev, .debug_loc, .debug_pubtypes) change between builds as well

**To Reproduce**
* Build Veracruz, e.g. `make nitro`
* Hash e.g. `veracruz-server`: `sha256sum nitro-host/target/debug/veracruz-server`
* Clean: `make clean`
* Build Veracruz again: `make nitro`
* Hash `veracruz-server`: `sha256sum nitro-host/target/debug/veracruz-server`
* The two hashes are different because the build ID is different: `objdump --section=.note.gnu.build-id --full-contents nitro-host/target/release/veracruz-server`

**Expected behaviour**
It's not clear to me how the build ID is determined, but I would expect it to stay constant between builds as long as the binaries are functionally equivalent and the code doesn't change.

**Solutions**
Strip build ID from binaries (`strip --remove-section=.note.gnu.build-id <binary>`) or fix build ID generation

dreemkiller:
I find it hard to imagine that the toolchain doesn't have a way to resolve this, as this is a big barrier to reproducible builds. 
However, a quick search doesn't show anything obvious.

gbryant-arm:
We could ignore it (stripping it or setting a default build ID) or fixing the cause: it's possible that one of our dependencies is doing something funky and interferes with the build ID somehow

egrimley-arm:
Firstly, I can't reproduce this as described above: I got the same binary after running `make clean`. However, when I tried with a fresh check-out I did get a different binary, and the only difference between the `veracruz-server` binaries seemed to be the build-id.

We can suppress generation of the build-id with `RUSTFLAGS="-C link-args=-Wl,--build-id=none"`: I tested that and it seemed to work.

However, it would be interesting to know why the build-id is changing. One of the inputs to the linker must be different each time, even though the change only affects the build-id. If we could discover what that input is perhaps we could find a better solution that using `--build-id=none`.

[...]

PR #519, opened 2022-09-22
Strip build ID from Nitro binaries

gbryant-arm:
Fix for https://github.com/veracruz-project/veracruz/issues/518


egrimley-arm:
Here's an alternative approach: #520

PR #520, opened 2022-09-22
Nitro: For reproducibility, do not include build-id in binaries.

egrimley-arm:
Add RUSTFLAGS="-C link-args=-Wl,--build-id=none" to Makefiles.

(It would be interesting to know what change in the input to the linker is causing the build-id to vary.)


dreemkiller (approved):

egrimley-arm:
This was a work-around for issue #518.

PR #525, opened 2022-09-28
Re-organization of crates

mathias-arm:
This PR moves all Veracruz crates in `crates/` directory (except for ones under `workspaces`).


mathias-arm:
Will rebase after #651 is merged.

egrimley-arm:
I think the reasons for doing this should be recorded in the PR or in the log message. (Was there something about making `grep` easier?)

If I've understood correctly, this increases the number of symbolic links from 7 to 8 (by adding `crates/third-party -> ../third-party`) but gets rid of the potential recursion because we now have `workspaces/*/crates -> ../../crates` instead of `workspaces/*/crates -> ../..`.

What disruption might be caused by moving `third-party` into the `crates/` directory instead of adding that symbolic link? I can imagine that there *might* be problems with merging or rebasing when submodules move around, but I can't remember trying it so perhaps it just works?

mathias-arm:
> I think the reasons for doing this should be recorded in the PR or in the log message. (Was there something about making `grep` easier?)

Yes, I wanted to put the Veracruz Rust code in one place. Having the different crates at the top-level with others directories (in particular `workspaces`) is making things harder to grep for things.

> What disruption might be caused by moving `third-party` into the `crates/` directory instead of adding that symbolic link? I can imagine that there _might_ be problems with merging or rebasing when submodules move around, but I can't remember trying it so perhaps it just works?

Initially I was thinking about moving `third-party` as well, but moving submodules adds a little bit of friction: when switching between branches before and after the move, git does not clean-up. Not a show-stopper though.

> On second thoughts, perhaps it's better to keep `third-party/` at the top level anyway.

I am now leaning also in this direction since it does help separating our code from `third-party`.

> But then I have to ask myself why `QCBOR` and `t_cose` are not in `third-party/`.

That could also be addressed (later?).

dreemkiller (approved):

PR #527, opened 2022-09-29
Replace "curl" crate with "reqwest".

egrimley-arm:
This makes the `Cargo.lock` files a bit longer, because `reqwest` has so many dependencies, but `reqwest` seems to be well maintained, is more popular than `curl`, and is implemented in Rust rather than a wrapper for a C library.


egrimley-arm:
@mathias-arm: This is still a draft but perhaps you'd like to take a look.

mathias-arm:
The `http` part of `io-utils` is only used by `veracruz-server`, so with some refactoring, the runtime workspaces should not require `reqwest` (et al.).

egrimley-arm:
> How about enabling `native-tls-vendored` for `reqwest`? I would possibly reduce the inflation in `Cargo.lock` files.

That doesn't seem to help, but `default-features = false` helps.

dreemkiller (approved):

PR #537, opened 2022-10-06
Update URL for IceCap repositories

mathias-arm:


egrimley-arm (approved):
It's identical to what I did!

Issue #535, opened 2022-10-06
Encryption/Decryption Native Module


ericvh:

ericvh:
using parts of this in the client side of the demo

gbryant-arm:
Done

PR #538, opened 2022-10-10
Fix padding in AES-CTR native module

gbryant-arm:
AES-CTR expects the output to be at least a full block longer than the input, or else it fails with `CipherFullBlockExpected`.
Completes https://github.com/veracruz-project/veracruz/issues/535

### DIFF
--- a/execution-engine/src/native_modules/aes.rs
+++ b/execution-engine/src/native_modules/aes.rs
@@ -100,8 +100,9 @@ impl AesCounterModeService {
             .map_err(|_| ErrNo::Canceled)?;
 
             let block_size = cypher.block_size();
-            // Mbed TLS requires the output buffer to be at least `ilen + block_size` long
-            let padded_size = (input.len() + 2 * block_size - 1) / block_size * block_size;
+            // Mbed TLS requires the output buffer to be at least `ilen + block_size` long.
+            // Cf. the documentation of `mbedtls_cipher_update()`
+            let padded_size = input.len() + block_size;
             output.resize(padded_size, 0);
 
             cypher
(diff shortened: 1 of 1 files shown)


execution-engine/src/native_modules/aes.rs
```diff
@@ -102,10 +102,7 @@ impl AesCounterModeService {
             let block_size = cypher.block_size();
             // For some reason, the length of the output buffer needs to be
             // a multiple of the block size and at least two blocks:
-            let padded_size = std::cmp::max(
-                2 * block_size,
-                (input.len() + block_size - 1) / block_size * block_size,
-            );
+            let padded_size = (input.len() + 2 * block_size - 1) / block_size * block_size;
```
dreemkiller, line 105: Understanding what this does requires the reader know Rust operator precedence pretty well, as well as how it handles division of integers. Can we make it clearer by adding parentheses and a comment about the behavior of integer division?

execution-engine/src/native_modules/aes.rs
```diff
@@ -124,10 +121,7 @@ impl AesCounterModeService {
...
```
dreemkiller, line 124: Same on this one.

execution-engine/src/native_modules/aes.rs
```diff
@@ -102,10 +102,7 @@ impl AesCounterModeService {
             let block_size = cypher.block_size();
             // For some reason, the length of the output buffer needs to be
```
dreemkiller, line 103: I know this isn't your comment, but can we find a resource describing why the output buffer needs to be padded as such and update this comment?

dreemkiller (changes requested):

execution-engine/src/native_modules/aes.rs
egrimley-arm: Or at least remove the existing comment that states something quite different!

It seems very bizarre that the output buffer has to be a multiple of TWICE the block size. Do we have any idea why that is?

I wonder whether the original comment and code was correct for some other algorithm or configuration or whether it was always just wrong.

gbryant-arm:
The output buffer only needs to be at least block_size longer than the input. No need to be a multiple of the block size

gbryant-arm:
https://github.com/fortanix/rust-mbedtls/issues/184 discusses

egrimley-arm (approved):
If you want to be careful, test this on input of length 0, 1, 2, 3, ..., 999, say. The previous version was presumably only tested on lengths up to the block size.

dreemkiller:
It should be sufficient to test it on input lengths of 1-16. If paranoid, go up to 33

gbryant-arm:
Works

dreemkiller (approved):

PR #540, opened 2022-10-11
Docker: Upgrade Nitro CLI

### LINKED ISSUE #507: Entropy problem on Nitro

gbryant-arm:
Upgrade Nitro CLI to 1.2.0.
This could help fix #507

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -6,7 +6,7 @@ USER ?= $(shell id -un)
 UID ?= $(shell id -u)
 OS_NAME := $(shell uname -s | tr A-Z a-z)
 ARCH = $(shell uname -m)
-AWS_NITRO_CLI_REVISION = v1.1.0
+AWS_NITRO_CLI_REVISION = v1.2.0
 
 ifeq ($(OS_NAME),darwin)
 LOCALIP = $(shell "(ifconfig en0 ; ifconfig en1) | grep "inet " | grep -Fv 127.0.0.1 | awk '{print $2}'")


dreemkiller (approved):

gbryant-arm:
> That looks uncontroversial! Of course the CI won't test it so it would be nice to test it manually before merging.

I can give it a try, I really need a newer container!

PR #542, opened 2022-10-13
Replace HTTP with TCP.

egrimley-arm:
Currently using TCP as if it were HTTP, so each packet is sent by opening a new connection!


egrimley-arm:
On GitHub this keeps failing on IceCap but it works on my machine. Perhaps one of the arbitrary delays in `integration_tests.rs` is to blame.

dreemkiller (approved):

PR #545, opened 2022-10-13
Replace RNG workaround with monitoring mechanism and remove security disclaimer

gbryant-arm:
We can remove the RNG workaround and the security disclaimer now that the bug is gone.
Let's keep the monitoring mechanism (timeout) just in case


egrimley-arm (approved):

egrimley-arm (changes requested):
No, don't merge this. The history of `rust-mbedtls` is all wrong, I think.

gbryant-arm:
> No, don't merge this. The history of `rust-mbedtls` is all wrong, I think.

Right, it's missing https://github.com/veracruz-project/rust-mbedtls/commit/9c8cc8302eae354b3a8ee3e2dff2a008688f36f6

egrimley-arm (approved):
No, I think this is correct. Submodules are confusing. I'll merge this so as to get this confusing nightmare behind me. (The history of `rust-mbedtls` is a bit messy.)

PR #547, opened 2022-10-18
Remove some async, base64 and sleep.

egrimley-arm:
- No async in veracruz-client.
- No async in veracruz-server.
- No base64-encoding in communication between client and server.
- veracruz_server::server::server now returns only when the server is ready and listening.
- Remove sleeps from integration_test.rs and server_test.rs.

### DIFF
--- a/veracruz-client/src/veracruz_client.rs
+++ b/veracruz-client/src/veracruz_client.rs
@@ -11,6 +11,7 @@
 
 use crate::error::VeracruzClientError;
 use anyhow::{anyhow, Result};
+use bincode;
 use log::{error, info};
 use mbedtls::{alloc::List, pk::Pk, ssl::Context, x509::Certificate};
 use policy_utils::{parsers::enforce_leading_backslash, policy::Policy, Platform};
@@ -22,8 +23,6 @@ use std::{
         Arc,
     },
 };
-use tokio::io::{AsyncReadExt, AsyncWriteExt};
-use tokio::net::TcpStream;
 use veracruz_utils::VERACRUZ_RUNTIME_HASH_EXTENSION_ID;
 
 /// VeracruzClient struct. The remote_session_id is shared between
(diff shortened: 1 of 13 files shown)


veracruz-server/src/cli.rs
```diff
@@ -47,27 +46,15 @@ fn main() {
...
     );
-    match sys.block_on(veracruz_server) {
-        Ok(_) => {}
-        Err(err) => {
-            eprintln!("{}", err);
-            process::exit(1);
-        }
+
+    loop {
+        std::thread::sleep(std::time::Duration::MAX);
```
dreemkiller, line 58: I also said this in another PR, but I think this one comes first, so here it is again:
I don't think this is necessary or desirable. If the server exits with an error above, it will panic (and execution will stop). If it exits without an error (not sure if that's possible, really), I would still like the program to stop instead of hang. `sleep` feels wrong here.

veracruz-server/src/server.rs
```diff
@@ -16,31 +16,24 @@ use crate::platforms::icecap::VeracruzServerIceCap as VeracruzServerEnclave;
...
-    if fields.len() < 2 {
+fn veracruz_server_request(
+    enclave_handler: EnclaveHandler,
+    input_data: &[u8],
+) -> Result<Vec<u8>, VeracruzServerError> {
+    if input_data.len() < 4 {
         return Err(VeracruzServerError::InvalidRequestFormatError);
     }
-    let session_id = match fields[0].parse::<u32>()? {
+    let session_id = match LittleEndian::read_u32(&input_data) {
```
dreemkiller, line 36: yadda, yadda, yadda: `LittleEndian` here

dreemkiller (changes requested):

veracruz-server/src/cli.rs
dreemkiller: Got it

veracruz-client/src/veracruz_client.rs
```diff
@@ -70,45 +71,30 @@ impl Write for InsecureConnection {
         let err = |t| std::io::Error::new(std::io::ErrorKind::Other, t);
 
         // Send all the data to the server.
-        let string_data = base64::encode(&data);
-        let combined_string = format!(
-            "{:} {:}",
+        let mut combined_data = vec![0, 0, 0, 0];
+        LittleEndian::write_u32(
```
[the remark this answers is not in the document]
egrimley-arm: Yes, `bincode` seems better. I'm replacing it.

dreemkiller (approved):

PR #548, opened 2022-10-19
Re-build Proxy Attestation Service using Veraison

dreemkiller:


docker/base/Dockerfile
```diff
@@ -15,6 +15,36 @@
 # NOTE: We try to follow the guide in https://docs.docker.com/develop/develop-images/dockerfile_best-practices/
 #       Each RUN contains a bundle of steps, which reduces the cache.
 
+from golang:1.19 AS go_builder
+
+RUN set -eux; \
```
mathias-arm, line 20: Should we specify some tags or commits for those repositories?

workspaces/shared.mk
```diff
@@ -131,11 +132,12 @@ PATH_UNLINK_FILE        := $(shell echo "2^26" | bc)
 POLL_FD_READWRITE       := $(shell echo "2^27" | bc)
 SOCK_SHUTDOWN           := $(shell echo "2^28" | bc)
 FD_EXECUTE              := $(shell echo "2^29"  | bc)
+FD_EXECUTE              := $(shell echo "2^29"  | bc)
```
mathias-arm, line 135: Duplicate

mathias-arm (commented):
This PR is missing the `Cargo.lock` changes.

workspaces/icecap-host/Makefile
```diff
@@ -23,6 +23,7 @@ WORKSPACE_DIR = $(abspath ..)
 OUT_DIR ?= $(abspath test-collateral)
 MEASUREMENT_FILE = $(abspath ../icecap-runtime/css-icecap.bin)
 MEASUREMENT_PARAMETER = --css-file $(MEASUREMENT_FILE)
+PROFILE_FLAG = release
```
mathias-arm, line 26: This statement is overridden by `common.mk`. Did you mean `PROFILE = release`?

workspaces/Makefile
```diff
@@ -32,7 +32,7 @@ nitro: shared
 	$(MAKE) -C nitro-runtime
 
 icecap: shared
-	$(MAKE) -C icecap-runtime build
+	$(MAKE) -C icecap-runtime build PROFILE=release
 	$(MAKE) -C icecap-host
```
mathias-arm, line 37: I have concerns about forcing the value of the `PROFILE` (it will take precedence over the value set by user), and also having possibly different values for `icecap-runtime` and `icecap-host`.

docker/base/Dockerfile
dreemkiller: Yes, we should.

workspaces/icecap-host/Makefile
dreemkiller: Yes, I did.

workspaces/Makefile
dreemkiller: The issue I am trying to solve is that anything other than PROFILE=release does not work (and I wasted 1.5 days of work figuring this out). How would you propose we solve this?

dreemkiller:
Regarding the lack of `Cargo.lock` files: I thought it was our policy to add those last? Supposing that was the case, I omitted them until I can get all other issues in the PR resolved, and then I was going to submit them.

Also, there are a number of changes in the `Cargo.lock` files that are unrelated to this change, as there have been a number of merges that have occurred recently that don't contain them.

workspaces/Makefile
mathias-arm: We should fix the `dev` profile settings in `workspace/icecap-runtime/Cargo.toml` so that it works.

dreemkiller: done

docker/base/Dockerfile
```diff
@@ -15,6 +15,36 @@
...
+    echo "iteration 0"; \
+    git clone --branch aws_nitro https://github.com/dreemkiller/services.git --branch v0.2 --depth 1; \
+    git clone https://github.com/veracruz-project/proxy_attestation_server.git --branch v0.1 --depth 1; \
+    cd proxy_attestation_server; \
+    go build -o ./vts github.com/veraison/services/vts/cmd/vts-service; \
```
mathias-arm, line 25: I should have been clearer that the comment applied to the repositories involved in the `go build` steps as well.

workspaces/icecap-host/Cargo.toml
```diff
@@ -30,3 +29,8 @@ resolver = "2"
 codegen-units = 1
 lto = true
 opt-level = 3
+
+[profile.dev]
+codegen-units = 1
+lto = true
```
mathias-arm, line 35: There are more differences between `dev` and `release` than `opt-level` and `codegen-units`. The defaults are (source: https://doc.rust-lang.org/cargo/reference/profiles.html):

```
[profile.dev]
opt-level = 0
debug = true
split-debuginfo = '...'  # Platform-specific.
debug-assertions = true
overflow-checks = true
lto = false
panic = 'unwind'
incremental = true
codegen-units = 256
rpath = false
```

```
[profile.release]
opt-level = 3
debug = false
split-debuginfo = '...'  # Platform-specific.
debug-assertions = false
overflow-checks = false
lto = false
panic = 'unwind'
incremental = false
codegen-units = 16
rpath = false
```

Is setting the `lto` to `true` needed to fix the issue?

mathias-arm (commented):
A lot of positive changes. I am interested to see the `Cargo.lock` diffs you mentioned (in particular the ones not related to your changes).

workspaces/icecap-host/Cargo.toml
dreemkiller: Well, these changes went through CI fine. So whatever needed to be changed was changed. Since this icecap issue has been longstanding, I don't think this it the PR to root-cause it.

dreemkiller:
The `Cargo.lock` files have been pushed.

The changes are legion.

workspaces/icecap-host/Cargo.toml
mathias-arm: That's because CI calls `make -C /work/veracruz/workspaces icecap-tests PROFILE=release`
(https://github.com/veracruz-project/veracruz/blob/main/.github/workflows/main.yml#L61)

docker/base/Dockerfile
dreemkiller: Those are currently not pulling from github, but from the `services` pull above them (there's an override in the `go.mod` file as not all of the necessary changes have been upstreamed). Once we get them upstreamed, the `git clone` for `services` will no longer be necessary, and the `go.mod` file will specify the exact revision needed.

workspaces/icecap-host/Cargo.toml
dreemkiller: I'm reverting this, as this is turning into too much trouble and this PR is not the place to figure this out

dreemkiller:
Is this ready to merge? Or do you have further questions?

dreemkiller:
Is this ready to merge or are there further questions?

mathias-arm (dismissed):
Some less desirable updates to `Cargo.lock` files in this PR, but that is not due to your changes.

mathias-arm (dismissed):

mathias-arm (approved):

PR #549, opened 2022-10-19
Remove async from integration_test.rs.

egrimley-arm:
We also remove the timeouts, but I think these were mostly not working anyway.

### DIFF
--- a/tests/tests/integration_test.rs
+++ b/tests/tests/integration_test.rs
@@ -53,49 +53,17 @@ const INTERSECTION_SET_SUM_ADVERTISEMENT_DATA: &'static str =
 const STRING_1_DATA: &'static str = "hello-world-1.dat";
 const STRING_2_DATA: &'static str = "hello-world-2.dat";
 
-const TIME_OUT_SECS: u64 = 1200;
-
 use anyhow::{anyhow, Result};
-use either::{Left, Right};
 use env_logger;
 use log::{error, info};
 use policy_utils::policy::Policy;
-use std::{
-    env::{self, VarError},
-    future::Future,
-    path::Path,
-    time::Duration,
-};
-use tokio::time::Instant;
+use std::{path::Path, thread, time::Instant};
 use veracruz_client::{self, VeracruzClient};
 use veracruz_server;
 
-/// A wrapper to force tests to panic after a timeout.
-///
-/// Note this is overrideable with the VERACRUZ_TEST_TIMEOUT environment
-/// variable, which provides a timeout in seconds
-pub async fn timeout<F: Future>(timeout: Duration, f: F) -> <F as Future>::Output {
-    let timeout = match env::var("VERACRUZ_TEST_TIMEOUT")
-        .map_err(Left)
-        .and_then(|timeout| timeout.parse::<u64>().map_err(Right))
-    {
-        Ok(val) => Duration::from_secs(val),
-        Err(Left(VarError::NotPresent)) => timeout,
-        Err(err) => panic!("Couldn't parse VERACRUZ_TEST_TIMEOUT: {:?}", err),
-    };
-
-    match tokio::time::timeout(timeout, f).await {
-        Ok(r) => r,
-        Err(_) => panic!(
-            "timeout after {:?}, specify VERACRUZ_TEST_TIMEOUT to override",
-            timeout
-        ),
-    }
-}
-
 /// A test of veracruz using network communication using a single session
-#[tokio::test]
-async fn veracruz_phase1_get_random_one_client() {
+#[test]
+fn veracruz_phase1_get_random_one_client() {
     TestExecutor::test_template(
         SINGLE_CLIENT_POLICY,
         &vec![(CLIENT_CERT, CLIENT_KEY)],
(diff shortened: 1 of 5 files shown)


tests/tests/integration_test.rs
```diff
@@ -247,78 +216,78 @@ async fn veracruz_phase3_string_edit_distance_four_clients() {
...
+        let data_filename = data_dir(LINEAR_REGRESSION_DATA);
+        info!("### data provider read input.");
+        let data = read_local_file(&data_filename).unwrap();
+        info!("### data provider send input.");
+        client.write_file("/input/linear-regression.dat", &data)?;
+        Result::<()>::Ok(())
+    });
+
+    client1_handle.join().unwrap().unwrap();
+    client2_handle.join().unwrap().unwrap();
```
dreemkiller, line 269: Per our discussion yesterday: it looks like the data client and the program provider client are parallel. However, since they are both `join`ed before the result retriever starts, they are not in parallel with the result retriever. I'd be interested in seeing what happens if the result retriever were parallel, but this is not a requirement before getting this PR through.

tests/tests/integration_test.rs
```diff
@@ -366,76 +333,44 @@ impl TestExecutor {
 
     /// Execute this test. Clients collectively execute as a block, driven by the `events`, in
     /// parallel with the server.
-    async fn execute<P: AsRef<Path>, Q: AsRef<Path>>(
+    fn execute<P: AsRef<Path>, Q: AsRef<Path>>(
         self,
         client_cert_key_pairs: &[(P, Q)],
         events: Vec<(usize, TestEvent)>,
-        timeout: Duration,
+        _timeout: Duration,
```
dreemkiller, line 340: If we're not using `_timeout`, can we remove it?

tests/tests/server_test.rs
```diff
@@ -956,9 +956,6 @@ fn init_veracruz_server_and_tls_session<T: AsRef<str>>(
     let mut veracruz_server =
         VeracruzServerEnclave::new(policy_json.as_ref()).map_err(|e| anyhow!("{:?}", e))?;
 
-    // wait for the client to start
-    thread::sleep(Duration::from_millis(100));
-
```
dreemkiller, line 961: nice. Kill all the `sleep`s

veracruz-client/src/veracruz_client.rs
```diff
@@ -70,45 +71,30 @@ impl Write for InsecureConnection {
         let err = |t| std::io::Error::new(std::io::ErrorKind::Other, t);
 
         // Send all the data to the server.
-        let string_data = base64::encode(&data);
-        let combined_string = format!(
-            "{:} {:}",
+        let mut combined_data = vec![0, 0, 0, 0];
+        LittleEndian::write_u32(
```
dreemkiller, line 75: I don't really love this. `LittleEndian` is very arch specific. Is there another way to do this?
We've used rust `enums` with `bincode` in other places. I'm not sure if that's appropriate here, though.

veracruz-server/src/cli.rs
```diff
@@ -47,27 +46,15 @@ fn main() {
...
-
     // create Veracruz Server instance
-    let veracruz_server = match veracruz_server::server::server(&policy_json) {
-        Ok(veracruz_server) => veracruz_server,
-        Err(err) => {
-            eprintln!("{}", err);
-            process::exit(1);
-        }
-    };
+    veracruz_server::server::server(&policy_json).unwrap();
```
dreemkiller, line 50: Have you run anything using the `server-cli` with this PR? I haven't run it in a while, so no judgement if you haven't. Just curious.

veracruz-server/src/cli.rs
```diff
@@ -47,27 +46,15 @@ fn main() {
...
     );
-    match sys.block_on(veracruz_server) {
-        Ok(_) => {}
-        Err(err) => {
-            eprintln!("{}", err);
-            process::exit(1);
-        }
+
+    loop {
+        std::thread::sleep(std::time::Duration::MAX);
```
dreemkiller, line 58: Is this necessary or desirable? I would think if the server returns, we the server has exited. It will `panic` on any error, and not `panic` on... maybe never. In any case, if the server does exit, I would like my program to exit and not to just hang.

veracruz-server/src/server.rs
```diff
@@ -16,31 +16,24 @@ use crate::platforms::icecap::VeracruzServerIceCap as VeracruzServerEnclave;
...
-    if fields.len() < 2 {
+fn veracruz_server_request(
+    enclave_handler: EnclaveHandler,
+    input_data: &[u8],
+) -> Result<Vec<u8>, VeracruzServerError> {
+    if input_data.len() < 4 {
         return Err(VeracruzServerError::InvalidRequestFormatError);
     }
-    let session_id = match fields[0].parse::<u32>()? {
+    let session_id = match LittleEndian::read_u32(&input_data) {
```
dreemkiller, line 36: Same question on the `LittleEndian` here. Just not sure if it's the right thing to do

dreemkiller (changes requested):

tests/tests/integration_test.rs
egrimley-arm: I've just tried it: if the result retriever runs first then it fails: `Failed to send TLS data.  Error produced: NoEnt.`

veracruz-client/src/veracruz_client.rs
egrimley-arm: Done in other PR.

veracruz-server/src/cli.rs
egrimley-arm: This gets checked by `quickstart` test in CI.

veracruz-server/src/cli.rs
egrimley-arm: I've answered this on #547.

veracruz-server/src/server.rs
egrimley-arm: I've replaced with `bincode`.

dreemkiller (approved):

PR #552, opened 2022-10-25
Make veracruz-client use a persistent connection.

egrimley-arm:
The `session_id` is no longer used in `veracruz-client`.

This is probably just the first step, changing how the client communicates with the server. The next step will probably be to change how the server communicates with the enclave.

This change seems to make transferring a 100 MB file into the enclave 3x faster, and transferring a 100 MB file out of the enclave 140x faster. On Linux with `PROFILE=release`. If I've measured it correctly. (The new speeds are 65 MB/s in and 50 MB/s out.)

The changes to `Cargo.lock` files are caused by previous PRs. This change does not affect the `Cargo.lock` files.

### DIFF
--- a/veracruz-server/src/server.rs
+++ b/veracruz-server/src/server.rs
@@ -16,80 +16,60 @@ use crate::platforms::icecap::VeracruzServerIceCap as VeracruzServerEnclave;
 use crate::platforms::linux::veracruz_server_linux::VeracruzServerLinux as VeracruzServerEnclave;
 #[cfg(feature = "nitro")]
 use crate::platforms::nitro::veracruz_server_nitro::VeracruzServerNitro as VeracruzServerEnclave;
-use bincode;
 use policy_utils::policy::Policy;
 use std::io::{Read, Write};
-use std::net::{Shutdown, TcpListener, TcpStream};
+use std::net::{TcpListener, TcpStream};
 use std::sync::{Arc, Mutex};
 use std::thread;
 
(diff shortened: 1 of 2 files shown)


veracruz-server/src/server.rs
```diff
@@ -16,80 +16,58 @@ use crate::platforms::icecap::VeracruzServerIceCap as VeracruzServerEnclave;
...
-
-        enclave.tls_data(session_id, received_data.to_vec())?
+        enclave.new_tls_session()?
     };
 
-    // Shutdown the enclave
-    if !active_flag {
-        let mut enclave_handler_locked = enclave_handler.lock()?;
+    loop {
+        // This buffer size gave close to optimal performance for
```
gbryant-arm, line 41: Interesting! Have you tried different sizes? This should probably be moved to the top as a const

gbryant-arm (changes requested):

veracruz-server/src/server.rs
egrimley-arm: I tried it with some different sizes for the buffer, but I didn't do a lot of measurements or a thorough analysis: with 10000 it seemed significantly slower, and with 65536 it seemed only a little bit faster, and with 1000000 it seemed to be perhaps significantly slower. A smaller buffer seems safer in case the performance is limited by available cache so I went for 32768 rather than 65536.

gbryant-arm (approved):

PR #553, opened 2022-10-26
WIP: Add Darknet inference native module and associated example

gbryant-arm:
This native module calls Darknet to perform inference on an image and returns a list of predictions.
Input files are copied from the VFS onto the kernel FS, in a dedicated directory in which the module operates. This directory is removed after execution and the output is copied to the VFS directly.

Future enhancements (optional):
* Make detection thresholds configurable

### DIFF
--- a/execution-engine/src/native_modules/darknet_inference.rs
+++ b/execution-engine/src/native_modules/darknet_inference.rs
@@ -1,0 +1,202 @@
+//! A native module for ML inference on Darknet.
+//! Takes an input image, feeds it to the model and outputs a list of detected
+//! objects.
+//!
+//! ## Authors
+//!
+//! The Veracruz Development Team.
+//!
+//! ## Licensing and copyright notice
+//!
+//! See the `LICENSE_MIT.markdown` file in the Veracruz root directory for
+//! information on licensing and copyright.
+
+use crate::fs::{FileSystem, FileSystemResult};
+use crate::native_modules::common::Service;
+use darknet::{BBox, Detection, Image, Network};
+use lazy_static::*;
+use serde::Deserialize;
+use std::cmp::Ordering;
+use std::env::{current_dir, set_current_dir};
+use std::fmt::Write as _;
+use std::fs::{create_dir, create_dir_all, read_to_string, remove_dir_all, File};
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use wasi_types::ErrNo;
+
+lazy_static! {
+    // Current directory before moving to the module's directory
+    static ref WORK_DIR: PathBuf = current_dir().unwrap_or(PathBuf::from(""));
+
+    // The native module executes in a dedicated directory to mitigate data
+    // leaks
+    static ref MODULE_DIR: PathBuf = WORK_DIR.clone().join("darknet_inference_service");
+}
+
+/// Module's API.
+#[derive(Deserialize, Debug)]
+pub(crate) struct DarknetInferenceService {
+    /// Path to the input to be fed to the network.
+    input_path: PathBuf,
+    /// Path to the model's configuration.
+    cfg_path: PathBuf,
+    /// Path to the actual model (weights).
+    model_path: PathBuf,
+    /// Path to the labels file containing all the objects that can be detected.
+    labels_path: PathBuf,
+    /// Path to the output file containing the result of the prediction.
+    output_path: PathBuf,
+}
+
+impl Service for DarknetInferenceService {
+    /// Return the name of this service
+    fn name(&self) -> &str {
+        "Darknet Inference Service"
+    }
+
+    /// Triggers the service. The details of the service can be found in
+    /// function `infer`.
+    /// Here is the enter point. It also erases the state unconditionally
+    /// afterwards.
+    fn serve(&mut self, fs: &mut FileSystem, _input: &[u8]) -> FileSystemResult<()> {
+        // when reaching here, the `input` bytes are already parsed.
+
+        // Prepare module's directory
+        create_dir(MODULE_DIR.as_path()).map_err(|_| ErrNo::Access)?;
+        set_current_dir(MODULE_DIR.as_path())?;
+
+        let result = self.infer(fs);
+        // NOTE: erase all the states and files passed to the module and created
+        // during execution
+        self.reset();
+        result
+    }
+
+    /// For the purpose of demonstration, we always return true. In reality,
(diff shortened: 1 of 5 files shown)


execution-engine/src/native_modules/darknet_inference.rs
```diff
@@ -0,0 +1,202 @@
...
+//! information on licensing and copyright.
+
+use crate::fs::{FileSystem, FileSystemResult};
+use crate::native_modules::common::Service;
+use darknet::{BBox, Detection, Image, Network};
+use lazy_static::*;
+use serde::Deserialize;
+use std::cmp::Ordering;
+use std::env::{current_dir, set_current_dir};
+use std::fmt::Write as _;
```
dreemkiller, line 21: I've never seen this. What does it do?

execution-engine/src/native_modules/darknet_inference.rs
```diff
@@ -0,0 +1,202 @@
...
+use std::path::{Path, PathBuf};
+use wasi_types::ErrNo;
+
+lazy_static! {
+    // Current directory before moving to the module's directory
+    static ref WORK_DIR: PathBuf = current_dir().unwrap_or(PathBuf::from(""));
+
+    // The native module executes in a dedicated directory to mitigate data
+    // leaks
+    static ref MODULE_DIR: PathBuf = WORK_DIR.clone().join("darknet_inference_service");
```
dreemkiller, line 33: As I understand it, `MODULE_DIR` will be set to `WORK_DIR`. and `WORK_DIR` is being set to `current_dir()` or "". In what scenarios is `MODULE_DIR` different from the `CURRENT_DIR` and thus actually executes the module in a dedicated directory?

execution-engine/src/native_modules/darknet_inference.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    /// Triggers the service. The details of the service can be found in
+    /// function `infer`.
+    /// Here is the enter point. It also erases the state unconditionally
+    /// afterwards.
+    fn serve(&mut self, fs: &mut FileSystem, _input: &[u8]) -> FileSystemResult<()> {
+        // when reaching here, the `input` bytes are already parsed.
+
+        // Prepare module's directory
+        create_dir(MODULE_DIR.as_path()).map_err(|_| ErrNo::Access)?;
```
dreemkiller, line 65: `create_dir` can fail for reasons other than lack of access, right?

execution-engine/src/native_modules/darknet_inference.rs
```diff
@@ -0,0 +1,202 @@
...
+        self.reset();
+        result
+    }
+
+    /// For the purpose of demonstration, we always return true. In reality,
+    /// this function may check validity of the `input`, and even buffer the
+    /// result for further uses.
+    fn try_parse(&mut self, input: &[u8]) -> FileSystemResult<bool> {
+        let deserialized_input: DarknetInferenceService =
+            match postcard::from_bytes(&input).map_err(|_| ErrNo::Canceled) {
```
dreemkiller, line 80: `from_bytes` can fail for other reasons than `Canceled, right?

execution-engine/src/native_modules/darknet_inference.rs
```diff
@@ -0,0 +1,202 @@
...
+        // Load network and labels
+        let mut net =
+            Network::load(cfg_path, Some(model_path), false).map_err(|_| ErrNo::Canceled)?;
+        let object_labels = read_to_string(labels_path)?
+            .lines()
+            .map(ToOwned::to_owned)
+            .collect::<Vec<_>>();
+
+        // Run inference
+        let image = Image::open(input_path).map_err(|_| ErrNo::Canceled)?;
```
dreemkiller, line 136: Not always `Canceled`, right?

execution-engine/src/native_modules/darknet_inference.rs
```diff
@@ -0,0 +1,202 @@
...
+        let detections = net.predict(&image, 0.25, 0.5, 0.45, true);
```
dreemkiller, line 137: Where do these magic numbers come from?

dreemkiller (changes requested):

execution-engine/src/native_modules/darknet_inference.rs
gbryant-arm: It's useful when a trait requires an import but the name of the module is already taken (by std::io::Write here)

execution-engine/src/native_modules/darknet_inference.rs
gbryant-arm: `MODULE_DIR` will always be different from `WORK_DIR` since it has a suffix.
But if `current_dir()` fails, `MODULE_DIR` won't be an absolute path and subsequent fs operations will fail. That's not great but we can't return errors in `lazy_static`

execution-engine/src/native_modules/darknet_inference.rs
gbryant-arm: Yes. I could return `Canceled`, since that seems to be the default error for native modules (cf. AES modules)

execution-engine/src/native_modules/darknet_inference.rs
gbryant-arm: Cf. answer above. WASI errors are not exhaustive, we should pick a default for generic errors

execution-engine/src/native_modules/darknet_inference.rs
gbryant-arm: Cf. answer above. WASI errors are not exhaustive, we should pick a default for generic errors

execution-engine/src/native_modules/darknet_inference.rs
gbryant-arm: These are parameters (thresholds) to filter detections (after inference IIUC).
They should be passed as arguments to the module

gbryant-arm: Alternatively we could process detections on the client side: unless the labels list is confidential, there's no real benefit in doing that in the enclave. That would also simplify the module's API and make it more generic

gbryant-arm:
Waiting for a solution to build C++ to be merged (#556)

gbryant-arm:
Waiting for the new CI images to be pushed

gbryant-arm:
Reopened in [a new PR](https://github.com/veracruz-project/veracruz/pull/571)

PR #554, opened 2022-10-26
Freestanding execution engine: Add read/write access to services

gbryant-arm:
Programs running in the freestanding execution engine can't call services


egrimley-arm (dismissed):
I don't understand the whole permissions system and file system layout to know whether this is right or not, and I don't see anything corresponding to this in `veracruz/execution-engine/`, but I suppose it's all right.

gbryant-arm:
The FEE doesn't have enforce policies, so you have to manually configure the VFS to specify what files the program can access.
Whereas the actual execution engine configures access rights based on the policy.

egrimley-arm (approved):

PR #556, opened 2022-10-28
WIP: Add C++ support to build system

gbryant-arm:
Currently the build system doesn't allow the Nitro and IceCap runtimes to link against C++ code. This is an obstacle on the critical path towards building complex native modules (https://github.com/veracruz-project/veracruz/pull/553).
We propose to:
* Add a C++ compiler to build the IceCap runtime, which is simply missing
* Replace `musl` with [`musl-cross-make`](https://github.com/richfelker/musl-cross-make) for the Nitro runtime, to overcome musl's limitations

### DIFF
--- a/docker/nitro/Dockerfile
+++ b/docker/nitro/Dockerfile
@@ -59,6 +59,12 @@ ENV \
     AARCH64_UNKNOWN_LINUX_MUSL_OPENSSL_INCLUDE_DIR=/work/veracruz-nitro-root-enclave/musl/include \
     X86_64_UNKNOWN_LINUX_MUSL_OPENSSL_INCLUDE_DIR=/work/veracruz-nitro-root-enclave/musl/include
 
+RUN wget https://musl.cc/${ARCH}-linux-musl-native.tgz && \
+    tar zxvf ${ARCH}-linux-musl-native.tgz && \
+    cp ${ARCH}-linux-musl-native/bin/${ARCH}-linux-musl-gcc /bin && \
+    cp ${ARCH}-linux-musl-native/bin/${ARCH}-linux-musl-g++ /bin && \
+    rm -rf ${ARCH}-linux-musl-native ${ARCH}-linux-musl-native.tgz
+
 RUN mkdir -p /var/log/nitro_enclaves ; \
     touch /var/log/nitro_enclaves/nitro_enclaves.log ; \
     chmod -R a+rw /var/log/nitro_enclaves ; \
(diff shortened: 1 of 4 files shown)


docker/nitro/Dockerfile
```diff
@@ -59,6 +59,12 @@ ENV \
     AARCH64_UNKNOWN_LINUX_MUSL_OPENSSL_INCLUDE_DIR=/work/veracruz-nitro-root-enclave/musl/include \
     X86_64_UNKNOWN_LINUX_MUSL_OPENSSL_INCLUDE_DIR=/work/veracruz-nitro-root-enclave/musl/include
 
+RUN wget https://musl.cc/${ARCH}-linux-musl-native.tgz && \
+    tar zxvf ${ARCH}-linux-musl-native.tgz && \
+    cp ${ARCH}-linux-musl-native/bin/${ARCH}-linux-musl-gcc /bin && \
+    cp ${ARCH}-linux-musl-native/bin/${ARCH}-linux-musl-g++ /bin && \
```
mathias-arm, line 65: Rather than copying the binaries, you can make a symbolic link, or better yet extend `$PATH`.

dreemkiller (dismissed):

docker/nitro/Dockerfile
```diff
@@ -51,14 +51,21 @@ RUN wget https://github.com/openssl/openssl/archive/refs/tags/openssl-3.0.1.tar.
     make -j $(nproc) && \
     make install_sw install_ssldirs && \
     cd .. && \
-    rm -rf openssl-openssl-3.0.1 openssl-3.0.1.tar.gz
+    rm openssl-openssl-3.0.1 openssl-3.0.1.tar.gz
```
mathias-arm, line 54: This change seems suspicious since `openssl-openssl-3.0.1` is a directory.
Side question should we update to version 3.0.6? There is at least one DoS fix in [3.0.2](https://github.com/openssl/openssl/blob/master/CHANGES.md#changes-between-301-and-302-15-mar-2022) that might be relevant.

gbryant-arm: Good catch

gbryant-arm: How about we bump openssl in a separate PR?

dreemkiller: That seems like the correct thing to do

gbryant-arm: Done in #557

mathias-arm (approved):

PR #559, opened 2022-11-08
CI check on `Cargo.lock`

ShaleXIONG:
Add check on if any `Cargo.lock` is updated in the CI/CD run. 

Each CI process outputs if `Cargo.lock` is updated and where they are. A new final step reads all outputs and triggers failure if there are any update to `Cargo.lock`. Such check is done by the `verify-changed-files` action. Yet it is our fork due to a `git` with `--local` flag issue. This bug will be reported to `verify-changed-files`.

Minor:
- update the checkout action from v2 to v3.


mathias-arm (dismissed):

egrimley-arm:
Is it possible to use a script that invokes `git` instead of this buggy `verify-changed-files` action?

ShaleXIONG:
> Is it possible to use a script that invokes `git` instead of this buggy `verify-changed-files` action?

I will prefer to not reinventing the wheel. In fact, the action itself is very simple anyway, ref to [the script here](https://github.com/tj-actions/verify-changed-files/blob/main/entrypoint.sh). It is a combination of git-diff, grep and pattern match.

egrimley-arm:
> I will prefer to not reinvent the wheel. In fact, the action itself is very simple anyway, ref to [the script here](https://github.com/tj-actions/verify-changed-files/blob/main/entrypoint.sh). It is a combination of git-diff, grep and pattern match.

OK, perhaps you're right in this case, but reinventing the wheel is the right thing to do if the alternative is importing something like the Node.js/NPM `is-even` module. I suppose it's mostly about guessing which approach is most likely to require future maintenance. (That script uses some `git diff` options that I'd never heard of.)

mathias-arm:
> I will prefer to not reinvent the wheel. In fact, the action itself is very simple anyway, ref to [the script here](https://github.com/tj-actions/verify-changed-files/blob/main/entrypoint.sh). It is a combination of git-diff, grep and pattern match.

If we are not using the upstream anyways, why not import the script into our repository and run it from there? That sounds simpler than adding a new (forked) repository.

ShaleXIONG:
> > I will prefer to not reinvent the wheel. In fact, the action itself is very simple anyway, ref to [the script here](https://github.com/tj-actions/verify-changed-files/blob/main/entrypoint.sh). It is a combination of git-diff, grep and pattern match.
> 
> If we are not using the upstream anyways, why not import the script into our repository and run it from there? That sounds simpler than adding a new (forked) repository.

Sorry to come back a bit late. Do we allow copy-paste others' code directly into our repo? I feel a bit uncomfortable doing so. The code is simple to read and understand the script but there is some subtlety in the code.

egrimley-arm:
> Sorry to come back a bit late. Do we allow copy-paste others' code directly into our repo? I feel a bit uncomfortable doing so. The code is simple to read and understand the script but there is some subtlety in the code.

Here, I think, is a precedent in our repo: https://github.com/veracruz-project/veracruz/blob/2682973574a1424352f8905a6b74b470a36f4145/psa-attestation/c_src/attest_token.c#L1-L9

`verify-changed-files` seems to use the same MIT licence, but check that carefully. Then, I would suggest, add a suitable header with a copyright line mentioning the original authors and a link to where you took it from, like in `attest_token.c` (which @dreemkiller worked on, so I'll mention him and give him a chance to object). No need to add an Arm or Veracruz copyright line until we make changes, which it would be neater to do as a separate commit even if you're making those changes in the same PR that imports the file.

mathias-arm:
Unless I am missing something, it feels like we do not need most of the features the external dependency brings. How about something like this:

```
  linux:
    runs-on: ubuntu-latest
    needs: [check]
    outputs:
      output: ${{ steps.check-diff.outputs.cargo-lock-changed }}
    container:
      image: ghcr.io/veracruz-project/veracruz/ci@sha256:b5eb834cd58c69e6acc19b1b19c6345a020ad2ad090cdab3bc9f7d5d48fb00ba
      volumes:
        - ${{ github.workspace }}:/work/veracruz
    steps:
      - name: Check out the Veracruz repository
        uses: actions/checkout@v3
        with:
          submodules: recursive
      - name: Running linux test script
        id: linux-build-and-test
        run: |
            make -C /work/veracruz/workspaces linux-tests
      - name: Run step only when any of the above files change.
        id: check-diff
        run: |
          cd /work/veracruz
          if git diff */workspaces/*/Cargo.lock | grep -q '^[+]' ; then \
            echo "::warning::Cargo.lock files modified" ; \
            echo "cargo-lock-changed=true" >> $GITHUB_OUTPUT ; \
          else \
            echo "cargo-lock-changed=false" >> $GITHUB_OUTPUT ; \
          fi
      - name: Upload Cargo.lock files
        if: steps.check-diff.outputs.cargo-lock-changed == 'true'
        uses: actions/upload-artifact@v3
        with:
            name: linux.zip
            path: **/workspaces/**/Cargo.lock
```

ShaleXIONG:
> Unless I am missing something, it feels like we do not need most of the features the external dependency brings. How about something like this:
> 
> ```
>   linux:
>     runs-on: ubuntu-latest
>     needs: [check]
>     outputs:
>       output: ${{ steps.check-diff.outputs.cargo-lock-changed }}
>     container:
>       image: ghcr.io/veracruz-project/veracruz/ci@sha256:b5eb834cd58c69e6acc19b1b19c6345a020ad2ad090cdab3bc9f7d5d48fb00ba
>       volumes:
>         - ${{ github.workspace }}:/work/veracruz
>     steps:
>       - name: Check out the Veracruz repository
>         uses: actions/checkout@v3
>         with:
>           submodules: recursive
>       - name: Running linux test script
>         id: linux-build-and-test
>         run: |
>             make -C /work/veracruz/workspaces linux-tests
>       - name: Run step only when any of the above files change.
>         id: check-diff
>         run: |
>           cd /work/veracruz
>           if git diff */workspaces/*/Cargo.lock | grep -q '^[+]' ; then \
>             echo "::warning::Cargo.lock files modified" ; \
>             echo "cargo-lock-changed=true" >> $GITHUB_OUTPUT ; \
>           else \
>             echo "cargo-lock-changed=false" >> $GITHUB_OUTPUT ; \
>           fi
>       - name: Upload Cargo.lock files
>         if: steps.check-diff.outputs.cargo-lock-changed == 'true'
>         uses: actions/upload-artifact@v3
>         with:
>             name: linux.zip
>             path: **/workspaces/**/Cargo.lock
> ```

I have updated the script, can you have another look? I tried a CI where two `cargo.lock` files were updated, https://github.com/veracruz-project/veracruz/actions/runs/3840216059, can you also check if it makes sense?

.github/workflows/main.yml
```diff
@@ -108,3 +184,50 @@ jobs:
...
+        if: steps.check-diff.outputs.cargo-lock != ''
+        uses: actions/upload-artifact@v3
+        with: 
+          name: quickstart 
+          path: workspaces/**/Cargo.lock
+
+  cargo-lock-check:
+    needs: [linux, nitro, icecap, quickstart]
+    runs-on: ubuntu-latest
+    container:
```
mathias-arm, line 209: We probably don't need to be using our image for this workflow

.github/workflows/main.yml
```diff
@@ -29,67 +29,143 @@ jobs:
...
+        run: |
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
       - name: Running Nitro test script
         id: nitro-build
         run: |
             make -C /work/veracruz/workspaces nitro
+      - name: Check modification to Cargo.lock 
+        id: check-diff
+        run: |
+          file_changed=$(git diff --diff-filter=ACMUXTRD --name-only -- '**Cargo.lock' | tr '\n' ' ' | xargs)
```
mathias-arm, line 95: I don't think we need both `tr` and `xargs`

.github/workflows/main.yml
```diff
@@ -29,67 +29,143 @@ jobs:
...
+          echo "cargo-lock=$file_changed" >> $GITHUB_OUTPUT
+      - name: Output list of changed Cargo.lock
```
mathias-arm, line 97: This step could be done as part as the previous step with something like:
```
if [ -n "$file_changed" ] ; then \
   echo "::warning::Cargo.lock files modified"  ; \
   echo "::warning::Cargo.lock change list: $file_changed"  ; \
fi
```

mathias-arm (dismissed):
I added some minor comments intended to make the changes smaller and simpler (the file is already quite large and complex), merging should not be stopped because of them.

.github/workflows/main.yml
ShaleXIONG: The `GITHUB` output dislikes the existence of line-break. The last `xargs` removes starting and trailing white space.

.github/workflows/main.yml
ShaleXIONG: Removed all the container fields.

mathias-arm (approved):

.github/workflows/main.yml
mathias-arm: I am just surprised it makes any difference, because I did a few tests:
```
$ echo >> workspaces/applications/Cargo.lock
$ echo >> workspaces/data-generators/Cargo.lock
$ foo=$(git diff --diff-filter=ACMUXTRD --name-only -- '**Cargo.lock' | tr '\n' ' ' | xargs) ; echo $foo
workspaces/applications/Cargo.lock workspaces/data-generators/Cargo.lock
$ foo=$(git diff --diff-filter=ACMUXTRD --name-only -- '**Cargo.lock' | tr '\n' ' ') ; echo $foo
workspaces/applications/Cargo.lock workspaces/data-generators/Cargo.lock
$ foo=$(git diff --diff-filter=ACMUXTRD --name-only -- '**Cargo.lock' | xargs) ; echo $foo
workspaces/applications/Cargo.lock workspaces/data-generators/Cargo.lock
```
I think the trailing whitespace is being trimmed in the assignment. I will resolve the conversation.

PR #562, opened 2022-11-10
Add `psmisc` to base image

gbryant-arm:
For compatibility reasons (GLIBC), Veracruz must be deployed in a Docker container provided by us.
The issue is that my deployment scripts depend on `killall`, which is part of `psmisc`, which isn't part of the base image.

### DIFF
--- a/docker/base/Dockerfile
+++ b/docker/base/Dockerfile
@@ -94,6 +94,7 @@ RUN apt-get update && \
         pkg-config \
         protobuf-compiler \
         procps \
+        psmisc \
         python3-dev \
         python3-pip \
         python3-protobuf \


dreemkiller (approved):

dreemkiller (changes requested):
Need to push a generated CI docker image, push it to ghcr.io/veracruz-project/veracruz/ci, probably as version v22.11a, and update .github/workflows/main.yaml

dreemkiller:
Per our discussion this morning, the consensus is that we don't need to push CI images and update the github/workflows/main.yaml file unless it's anticipated to be a breaking change.

dreemkiller (approved):

Issue #565, opened 2022-11-14
Linux Runtime Manager panics printing to stdout


gbryant-arm:
**Describe the bug**
On Linux, the Runtime Manager panics printing to stdout:
```
thread 'main' panicked at 'failed printing to stdout: Broken pipe (os error 32)', library/std/src/io/stdio.rs:1164:9
```
~~Incidentally, this results in Veracruz-Server's CPU usage spiking to 100%~~ (fixed in #567) 
This happens because VC-Server is using `stdout` to sync with RM, then destroys the pipe after sync, resulting in subsequent writes to `stdout` to fail.

**To Reproduce**
Add a `println!` to any dependency of the RM, e.g. in `execution-engine/src/fs.rs`.
Recompile and run Veracruz end-to-end.

**Expected behaviour**
The Runtime Manager shouldn't fail, let alone panic.

**Possible solutions**
1. As a workaround, `eprintln!` should be used for debugging instead of `println!`, however we don't have much control over what native modules do. For example, TensorFlow Lite, used by the TensorFlow inference native module, prints all sorts of stuff to `stdout`
2. Use a named pipe for the sync channel to avoid interfering with standard streams
3. Redirect `stdout` "back to itself" (or `/dev/null` in release mode?) after the pipe is destroyed on the VC-Server side. Is this possible on IceCap?

egrimley-arm:
Why does code in the Runtime Manager need to print to `stdout`? What should happen with data written to `stdout`, on Linux or other targets?

gbryant-arm:
That's mainly for debugging purposes. Admittedly we should use `eprintln!` or `info!` for that

gbryant-arm:
Hmm actually some native modules might use `stdout`. `tflite-rs` does that, and that's an issue

egrimley-arm:
Would it be a solution to get the Runtime Manager to redirect `stdout` to a file before it starts running the application?

gbryant-arm:
Native modules could write to any file, `stdout` is only one of them.
That could be a short-term solution though, until we figure out how to sandbox native modules. The other short-term solution is to fix this bug.

egrimley-arm:
I've reproduced "Veracruz-Server's CPU usage spiking to 100%" and I'm going to look at that first. It's probably going into a loop repeatedly reading EOF and failing to recognise it and give up.

gbryant-arm:
[...]

PR #567, opened 2022-11-14
One bug fix for issue #565.

egrimley-arm:
* In io-utils, use write_all/read_exact instead of write/read.

### DIFF
--- a/io-utils/src/fd.rs
+++ b/io-utils/src/fd.rs
@@ -24,22 +24,11 @@ where
     {
         let mut buff = [0u8; 9];
         LittleEndian::write_u64(&mut buff, len as u64);
-
-        let mut sent_bytes = 0;
-
-        while sent_bytes < 9 {
-            sent_bytes += fd.write(&buff[sent_bytes..9])?;
-        }
+        fd.write_all(&buff)?;
     }
 
     // 2. Send the data proper.
-    {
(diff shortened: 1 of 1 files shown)


egrimley-arm:
@gbryant-arm Can this be merged now?

gbryant-arm (changes requested):
The broken pipe bug on Linux is fixed by remapping `stdout` to `/dev/null`. Would that be easier for VC-Server and RM to use a channel other than `stdout` for syncing, say a named pipe? This way we wouldn't interfere with the standard channels, which will make it easier in the future to redirect them to `/dev/null` depending on whether we are in debug mode

egrimley-arm:
It would almost certainly not be "easier" ... and I don't know exactly what's possible on Nitro, but presumably the enclave is not sharing a file system with the host? Shall I just remove the `/dev/null` part of this PR so we can merge the rest and close this PR?

egrimley-arm:
Perhaps the correct solution is to have the runtime manager connect to the Veracruz server rather than the other way round? Or is that not allowed? This PR is probably not the right place to discuss!

egrimley-arm:
I've removed the `/dev/null` stuff.

gbryant-arm:
> I've removed the `/dev/null` stuff.

Sounds good. I'm updating issue #565 to reflect that

PR #568, opened 2022-11-15
Add TensorFlow Lite inference module and MNIST example

gbryant-arm:
This native module takes a model and an input tensor, invokes TF Lite then returns the result of inference to the caller.
As experimented in #553, input files are copied from the VFS onto the kernel FS, in a dedicated directory in which the module operates. This directory is removed after execution and the output is copied to the VFS directly.

Future enhancements:
* Add another more involved example as a separate PR
* Support several input and output tensors
* Allow to further configure the TF interpreter. It's currently impossible to configure the number of threads due to #565


execution-engine/src/native_modules/tflite_inference.rs
```diff
@@ -0,0 +1,183 @@
...
+        let result = self.infer(fs);
+        // NOTE: erase all the states and files passed to the module and created
+        // during execution
+        self.reset();
+        result
+    }
+
+    /// For the purpose of demonstration, we always return true. In reality,
+    /// this function may check validity of the `input`, and even buffer the
+    /// result for further uses.
```
dreemkiller, line 75: Since this is no longer just a demo, should we change this?

gbryant-arm: Sure. Should we perform any check on the input? We can't really access any sensitive file since the module attempts to read them from the VFS with the program's permissions, before copying them to the FS.

gbryant-arm: I can't see any trivial way to inject malformed input here. Let's just amend the comment

gbryant-arm: Should the number of threads be configurable by the caller? Can you think of any attacks leveraging that?

dreemkiller (commented):
Why are there `Cargo.lock` file when there are no code updates?

gbryant-arm:
> Why are there `Cargo.lock` file when there are no code updates?

Blunder. Fixed

gbryant-arm:
Superseded by #587 and https://github.com/veracruz-project/tflite-nm

PR #571, opened 2022-11-17
Add Darknet inference native module and associated example

gbryant-arm:
Same as https://github.com/veracruz-project/veracruz/pull/553

### DIFF
--- a/examples/rust-examples/darknet-inference-native/src/main.rs
+++ b/examples/rust-examples/darknet-inference-native/src/main.rs
@@ -1,0 +1,64 @@
+//! An example program to call the Darknet inference module.
+//!
+//! ## Context
+//!
+//! It calls the module mounted at path `/services/darknet_inference.dat`, via
+//! the postcard encoding of the interface,
+//! ```
+//! pub struct DarknetInferenceInput {
+//!     input_path: PathBuf,
+//!     cfg_path: PathBuf,
+//!     model_path: PathBuf,
+//!     labels_path: PathBuf,
+//!     output_path: PathBuf,
+//! }
+//! ```
+//!
+//! ## Authors
+//!
+//! The Veracruz Development Team.
+//!
+//! ## Copyright
+//!
+//! See the file `LICENSE_MIT.markdown` in the Veracruz root directory for licensing
+//! and copyright information.
+
+use serde::Serialize;
+use std::{
+    fs::{read_to_string, write},
+    path::PathBuf,
+};
+
+/// The interface with the Darknet inference service
+#[derive(Serialize, Debug)]
+pub struct DarknetInferenceInput {
+    input_path: PathBuf,
+    cfg_path: PathBuf,
+    model_path: PathBuf,
+    labels_path: PathBuf,
+    output_path: PathBuf,
+}
+
+/// Example to invoke the Darknet inference service.
+/// Pass an image, a YOLO model with its configuration and a labels file to the
+/// service.
+/// The prediction is written to `output_path`
+fn main() -> anyhow::Result<()> {
+    let darknet_inference_input = DarknetInferenceInput {
+        input_path: PathBuf::from("input/image.jpg"),
+        cfg_path: PathBuf::from("input/yolov3-tiny.cfg"),
+        model_path: PathBuf::from("input/yolov3-tiny.weights"),
+        labels_path: PathBuf::from("input/coco.names"),
+        output_path: PathBuf::from("output/prediction.dat"),
+    };
+    println!("service input {:x?}", darknet_inference_input);
+
+    let darknet_inference_input_bytes = postcard::to_allocvec(&darknet_inference_input)?;
+    println!("calling Darknet Inference service...");
+    write("/services/darknet_inference.dat", darknet_inference_input_bytes)?;
+    let result = read_to_string(darknet_inference_input.output_path)?;
+    println!("prediction:\n{:x?}", result);
+    println!("service return");
+
+    Ok(())
+}
--- a/examples/rust-examples/darknet-inference-native/src/main.rs
+++ b/examples/rust-examples/darknet-inference-native/src/main.rs
@@ -2,17 +2,8 @@
 //!
 //! ## Context
 //!
-//! It calls the module mounted at path `/services/darknet_inference.dat`, via
-//! the postcard encoding of the interface,
-//! ```
-//! pub struct DarknetInferenceInput {
-//!     input_path: PathBuf,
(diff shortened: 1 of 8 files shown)


gbryant-arm:
Looks like `free()` isn't supported on icecap?
```
error[E0425]: cannot find function `free` in crate `libc`
   --> /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/darknet-0.3.4/src/network.rs:186:19
    |
186 |             libc::free(ptr as *mut c_void);
    |                   ^^^^ not found in `libc`
    |
help: consider importing one of these items
```

egrimley-arm:
> Looks like `free()` isn't supported on icecap?

It seems to be used in the C code. Perhaps it just isn't (yet) available via the `libc` crate. So it might be easy to add.

egrimley-arm:
This PR adds `darknet`, which depends on `darknet-sys`, as a dependency of `execution-engine`; `darknet-sys` has a load of C code as a git submodule, I think, and builds it with the `cmake` crate used by `build.rs`, in the usual way. Or is it C++ code? One of the READMEs refers to C, but this line in the `build.rs` seems to refer to C++: https://github.com/alianse777/darknet-sys-rust/blob/master/build.rs#L178

If the C/C++ code has a lot of awkward and unavoidable requirements, getting this to build for IceCap may be difficult. So the options seem to be:

* Make it work on IceCap, provided we can achieve that in hours rather than days.

* Remove IceCap from CI.

* Make this `darknet` feature optional so that it gets built and tested by CI on the other platforms, but not on IceCap. It seems to me that it's a well-defined separate module so it ought to be possible to make it optional and it's quite likely that native-code modules like this will be platform-specific in the future so we'll probably in any case need to implement some kind of mechanism to make make them optional, so I think this option sounds quite attractive, but what do other people think?

gbryant-arm:
The `darknet` crate indirectly depends on https://github.com/AlexeyAB/darknet/tree/04b0b15ee42c3b54ae457eb639d9b666a9f5ca8d, which is a mix of C (core) and C++ (examples).
The third solution, optionally building darknet, sounds the most attractive indeed. In the future native modules won't be a dependency of the execution engine and will be built separately.

gbryant-arm:
Whoops, forgot to use the feature

gbryant-arm:
Ready for review

examples/rust-examples/darknet-inference-native/src/main.rs
```diff
@@ -0,0 +1,84 @@
...
+
+use serde::Serialize;
+use std::{
+    fs::{read_to_string, write},
+    path::PathBuf,
+};
+
+/// The interface with the Darknet inference service.
+#[derive(Serialize, Debug)]
+pub struct DarknetInferenceInput {
```
egrimley-arm, line 25: Should this structure be defined in some place where it can be shared?

execution-engine/src/fs.rs
```diff
@@ -36,6 +38,7 @@ use std::{
 };
 // TODO: wait for icecap to support direct conversion between bytes and os_str, bypassing
 // potential utf-8 encoding check
+use log::error;
```
egrimley-arm, line 41: Did `rustfmt` suggest moving `use log::error;` to here? If not, perhaps leave it where it was. At this location it seems to separate a comment from the code that the comment refers to! If we're keeping these `use` declarations in alphabetical order then this is the wrong place anyway.

execution-engine/src/fs.rs
```diff
@@ -1951,12 +1971,9 @@ impl FileSystem {
     }
 
     /// Check if a `file_name` exists.
-    /// Note: this function *has* side effect ! 
+    /// Note: this function *has* side effect !
```
egrimley-arm, line 1974: (Space before exclamation mark is very French.)

execution-engine/src/native_modules/darknet_inference.rs
```diff
@@ -0,0 +1,238 @@
...
+    static ref WORK_DIR: PathBuf = current_dir().unwrap_or(PathBuf::from(""));
+
+    // The native module executes in a dedicated directory to mitigate data
+    // leaks
+    static ref MODULE_DIR: PathBuf = WORK_DIR.clone().join("darknet_inference_service");
+}
+
+/// Module's API.
+#[derive(Deserialize, Debug)]
+pub(crate) struct DarknetInferenceService {
```
egrimley-arm, line 38: This is a repeat of the stuff in `rust-examples`? I think we should have this in a shared file somewhere (but I don't know where that would be). If for some reason that's not possible or practicable there should be a comment explaining where the other copy is so that people know that they have to update both if it has to be changed.

execution-engine/src/native_modules/darknet_inference.rs
```diff
@@ -0,0 +1,238 @@
...
+        // NOTE: erase all the states and files passed to the module and created
+        // during execution
+        self.reset();
+        result
+    }
+
+    /// An attacker may inject malformed paths but that should be caught by the
+    /// VFS when attempting to access the corresponding files.
+    /// The input image is resized or letterboxed (depending on the `letterbox`
+	/// parameter) before being fed to the model, which guarantees dimensions
```
egrimley-arm, line 94: Why are these lines indented?

egrimley-arm (changes requested):

execution-engine/src/fs.rs
gbryant-arm: It is. @ShaleXIONG are you a French spy? Speculatively marking this as resolved but waiting for a background check on @ShaleXIONG

execution-engine/src/native_modules/darknet_inference.rs
gbryant-arm: There is a lot of repeat, though it sounds like a bad idea to link the examples against Veracruz, they should be self-contained / easy to build.
Adding a comment to refer to the structure expected by the native module sounds like a good idea :+1:

execution-engine/src/fs.rs
gbryant-arm: Hmm looks like `rustfmt` did that but when I reorder the line alphabetically and run `rustfmt` again it doesn't reorder it

egrimley-arm (changes requested):
Can you include the changes to `workspaces/nitro-runtime/Cargo.lock`? It looks good apart from that.

gbryant-arm:
This should be ported to a separate dynamic native module

gbryant-arm:
Superseded by #596 and https://github.com/veracruz-project/darknet-inference-nm

PR #572, opened 2022-11-22
Tidying: rustfmt and prevent warnings.

egrimley-arm:


egrimley-arm:
I was wondering why GitHub isn't showing the submodule diffs, but of course in this case the submodules are on GitLab!

ShaleXIONG (approved):

PR #573, opened 2022-11-23
Connect from Runtime Manager to Veracruz Server instead of other way.

### LINKED ISSUE #565: Linux Runtime Manager panics printing to stdout

egrimley-arm:
Only on Linux, for now. This probably fixed #565.

For reviewing this you might want to look at the separate commits as there is a commit that changes indentation.

### DIFF
--- a/runtime-manager/src/platforms/linux.rs
+++ b/runtime-manager/src/platforms/linux.rs
@@ -30,7 +30,7 @@ use log::{error, info};
 use psa_attestation::{
     psa_initial_attest_get_token, psa_initial_attest_load_key, psa_initial_attest_remove_key,
 };
-use std::net::TcpListener;
+use std::net::TcpStream;
 use veracruz_utils::{
     runtime_manager_message::{RuntimeManagerRequest, RuntimeManagerResponse, Status},
     sha256::sha256,
@@ -39,9 +39,6 @@ use veracruz_utils::{
 ////////////////////////////////////////////////////////////////////////////////
 // Constants.
 ////////////////////////////////////////////////////////////////////////////////
-
-/// Incoming address to listen on.  Note that `0.0.0.0` implies all addresses.
-const INCOMING_ADDRESS: &'static str = "0.0.0.0";
 
 /// **TOTALLY INSECURE** root private key to use for Linux PSA attestation.
 ///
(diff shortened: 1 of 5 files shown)


dreemkiller (commented):
So, this works, then?

egrimley-arm:
> What are the changes in nitro for, then?

There's a separate commit that removes some unused functions.

egrimley-arm:
This could be three separate tiny PRs, I suppose.

dreemkiller (approved):

veracruz-server/src/common.rs
```diff
@@ -104,21 +103,12 @@ pub trait VeracruzServer {
...
-    fn close_tls_session(&mut self, session_id: u32) -> VeracruzServerResult<()>;
-
     // The first bool indicates if the enclave is active, and the second vec contains the response
     fn tls_data(
         &mut self,
         session_id: u32,
         input: Vec<u8>,
     ) -> VeracruzServerResult<(bool, Option<Vec<Vec<u8>>>)>;
-
-    fn shutdown_isolate(&mut self) -> Result<(), Box<dyn Error>>;
```
gbryant-arm, line 123: Why was this moved outside of the `VeracruzServer` trait? Shouldn't `shutdown_isolate()` be implemented for every target?

gbryant-arm (commented):
Looks good. Is the reason for not implementing this for IceCap and Nitro too, making the PR smaller? Are you hopeful that the same thing can be implemented on those targets?

runtime-manager/src/platforms/linux.rs
```diff
@@ -201,29 +197,19 @@ pub fn linux_main() -> Result<()> {
...
-        error!("Did not receive any port to listen on.  Exiting...");
+        error!("No address given. Exiting...");
         return Err(anyhow!(RuntimeManagerError::CommandLineArguments));
     };
 
     let measurement = if let Some(measurement) = matches.value_of("runtime_manager_measurement") {
-        info!(
-            "Received {} as Runtime Manager enclave measurement.",
-            measurement
-        );
```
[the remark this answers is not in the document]
egrimley-arm: It just seemed unnecessary, and I wasn't adding anything for the other command-line option.

veracruz-server/src/common.rs
egrimley-arm: It's implemented, but it's not in the API any more because the shutdown is initated by dropping the struct: https://github.com/veracruz-project/veracruz/blob/2682973574a1424352f8905a6b74b470a36f4145/veracruz-server/src/server.rs#L61

PR #591, opened 2023-01-20
WIP: Modification to use the external `nitro-enclave` crate

dreemkiller:
The nitro-specific code from `io-utils` has been moved to a new crate external to this repo (as it may be generally useful). This change uses that new crate.


mathias-arm (approved):
Looks good to me

PR #593, opened 2023-02-24
Updates for Dependabot

dreemkiller:
Need to get bumpalo to 3.11.1
Need to get tokio to 1.18.4

### DIFF
--- a/workspaces/icecap-host/Cargo.lock
+++ b/workspaces/icecap-host/Cargo.lock
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "actix-rt"
-version = "2.7.0"
+version = "2.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7ea16c295198e958ef31930a6ef37d0fb64e9ca3b6116e6b93a8bdae96ee1000"
+checksum = "15265b6b8e2347670eb363c47fc8c75208b4a4994b27192f345fcbe707804f3e"
 dependencies = [
  "actix-macros",
  "futures-core",
@@ -184,9 +184,9 @@ checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
 
 [[package]]
 name = "bumpalo"
-version = "3.9.1"
+version = "3.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a4a45a46ab1f2412e53d3a0ade76ffad2025804294569aae387231a0cd6e0899"
+checksum = "572f695136211188308f16ad2ca5c851a712c464060ae6974944458eb83880ba"
 
 [[package]]
 name = "byteorder"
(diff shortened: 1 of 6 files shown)


egrimley-arm (commented):
The `Cargo.toml` changes look good, but I didn't get all the `Cargo.lock` changes when I tried building it myself: I didn't get the `host/Cargo.lock` and `*-runtime/Cargo.lock` changes. Is that expected, or were those changes included accidentally, left over from a previous version of the `Cargo.toml` changes?

mathias-arm:
I used the following command:

```
for i in workspaces/icecap-host workspaces/linux-host workspaces/nitro-host ; do \
   (cd $i ; cargo update -p bumpalo --precise 3.11.1 ; \
               cargo update -p actix-rt --precise 2.8.0 ;
               cargo update -p schannel --precise 0.1.21) ; \
done
```

`actix-rt` version 2.8 raises the tokio requirement to 1.18.4 (probably for the same reasons) and the `schannel` upgrade from 0.1.20 to 0.1.21 uses `windows-sys` version 0.42 instead of 0.36, which allows deduplication of `windows-targets` (and its depedencies. I did not need any to update `reqwest`.

The command above is basically running the build with the following changes and then backing them out:
```
diff --git tests/Cargo.toml tests/Cargo.toml
index 9587280e..b70e3942 100644
--- tests/Cargo.toml
+++ tests/Cargo.toml
@@ -32,7 +32,7 @@ nitro = [
 ]

 [dependencies]
-actix-rt = "2"
+actix-rt = "2.8"
 anyhow = "1"
 either = "1.6.1"
 env_logger = "0.9.0"
@@ -44,6 +44,7 @@ nix = "0.25.0"
 postcard = "0.7.2"
 policy-utils = { path = "../policy-utils", optional = true }
 reqwest = { version = "0.11", default-features = false, features = ["blocking"]}
+schannel = "0.1.21"
 veracruz-client = { path = "../veracruz-client" }
 veracruz-server = { path = "../veracruz-server" }
diff --git veracruz-server/Cargo.toml veracruz-server/Cargo.toml
index 35ed24be..e702889d 100644
--- veracruz-server/Cargo.toml
+++ veracruz-server/Cargo.toml
@@ -49,6 +49,7 @@ nitro = [
 anyhow = "1"
 base64 = "0.13.0"
 bincode = { version = "1.2.1", default-features = false }
+bumpalo = "3.11.1"
 data-encoding = { version = "2.3.2", optional = true }
 env_logger = { version = "0.9.0", optional = true }
 err-derive = "0.2"
```

As @egrimley-arm mentioned. With the changes in the `Cargo.toml` files of this PR, I don't get the `*-runtime/Cargo.lock` changes either.

dreemkiller:
What I see when I attempt to update `actix-rt` to `2.8.0` is that it depends tokio, which upgrades to `1.25.0`, which depends `windows-sys 0.42.0` and on mio,  which depends on `windows-sys 0.45.0`. So I don't see that the upgrade to `actix-rt` `2.8.0` is especially helpful in regards to `windows-sys`. Am I missing something?

mathias-arm:
Updating `schannel` to version 0.1.21 (`cargo update -p schannel --precise 0.1.21`) bumps its dependency on `windows-sys` from 0.36.1 to to 0.42.0 avoiding having 3 versions of `windows-sys`.
Having both `windows-sys 0.42.0` and `windows-sys 0.45.0` seems unavoidable as you mentioned but that duplication is not as bad because they share the same dependencies (when `windows-sys 0.36.1` uses distinct versions for its dependencies).

mathias-arm (approved):

PR #597, opened 2023-03-22
Add VOD deployment job to GitHub workflows

gbryant-arm:
Our CI doesn't include any tests for the freestanding execution engine (FEE), nor any deployments of relatively complex examples.
In this PR we add a job to GitHub workflows that deploys the video object detection example as a standalone binary, in wasmtime, in the FEE and end-to-end.

The changes add about 5min to the `linux` job (need to build Veracruz-Linux in addition to running the tests) and the new `vod-full-deployment` job takes about 4min and is executed after `linux`. This doesn't affect the total execution time since the `icecap` job takes more than 30min, which is more than `linux` and `vod-full-deployment` combined.

Some ideas to optimise this workflow further (**won't be implemented in this PR**):
* Build VOD (native and WASM binaries) in parallel to other jobs then share binaries with deployment job
  + Pro: Saves ~1min
  + Con: Adds another concurrent job to the workflow
* Add deployment steps to Veracruz build jobs as opposed to having a separate job for deployment
  + Pro: Saves ~2min (container and artifact sharing overhead)
  + Con: Need to rerun the entire job if deployment fails

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -96,14 +96,7 @@ jobs:
         with: 
           name: linux_deployment_artifacts
           path: |
-            artifacts/proxy_cleanup.sh
-            artifacts/ca-cert.conf
-            artifacts/cert.conf
-            artifacts/freestanding-execution-engine
-            artifacts/generate-policy
-            artifacts/veracruz-client
-            artifacts/veracruz-server
-            artifacts/runtime_manager_enclave
+            artifacts/*
 
   vod-full-deployment:
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,6 +40,10 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
+      - name: Build Veracruz-Linux
+        id: linux-build
+        run: |
+           make -C /work/veracruz/workspaces linux
       - name: Running linux test script
         id: linux-build-and-test
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,6 +65,143 @@ jobs:
         with: 
           name: linux 
           path: workspaces/**/Cargo.lock
+      - name: Prepare deployment artifacts
+        run: |
+          # Strip binaries
+          strip \
+          workspaces/host/target/debug/freestanding-execution-engine \
+          workspaces/host/target/debug/generate-policy \
+          workspaces/linux-host/target/debug/veracruz-client \
+          workspaces/linux-host/target/debug/veracruz-server \
+          workspaces/linux-runtime/target/debug/runtime_manager_enclave
+          # Copy artifacts to new directory
+          mkdir -p artifacts
+          cp -a \
+          proxy_cleanup.sh \
+          workspaces/ca-cert.conf \
+          workspaces/cert.conf \
+          workspaces/host/target/debug/freestanding-execution-engine \
+          workspaces/host/target/debug/generate-policy \
+          workspaces/linux-host/target/debug/veracruz-client \
+          workspaces/linux-host/target/debug/veracruz-server \
+          workspaces/linux-runtime/target/debug/runtime_manager_enclave \
+          artifacts/
+      - name: Upload deployment artifacts
+        id: upload-deployment-artifacts
+        uses: actions/upload-artifact@v3
+        with: 
+          name: linux_deployment_artifacts
+          path: |
+            artifacts/proxy_cleanup.sh
+            artifacts/ca-cert.conf
+            artifacts/cert.conf
+            artifacts/freestanding-execution-engine
+            artifacts/generate-policy
+            artifacts/veracruz-client
+            artifacts/veracruz-server
+            artifacts/runtime_manager_enclave
+
+  vod-full-deployment:
+    runs-on: ubuntu-latest
+    needs: [linux]
+    container:
+      image: ghcr.io/veracruz-project/veracruz/ci@sha256:38add618b42f59a89fdd4c46831dcf9c9e73b159bc74c87251c8263a2ec6238a
+      volumes:
+        - ${{ github.workspace }}:/work/video-object-detection
+    steps:
+      - name: Check out the VOD repository
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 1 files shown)


gbryant-arm:
Switching to "ready for review" just to trigger the CI and see what happens

.github/workflows/main.yml
```diff
@@ -40,6 +40,10 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
+      - name: Build Veracruz-Linux
+        id: linux-build
+        run: |
+           make -C /work/veracruz/workspaces linux
```
mathias-arm, line 46: Would it make sense to add `PROFILE=release`?

gbryant-arm: Hmm wouldn't that significantly slow down the build?

egrimley-arm:
Someone who wants to test the CI on their own machine has to copy-paste quite a lot of shell code from main.yml into a terminal. I wonder if it would be better to move the shell code into a script that one could more easily run manually.

Arguably that's a new feature that's distinct from what you're adding here so could be a separate PR, though. And there are various ways one could do it: for example, a single script that runs a different CI job depending on the argument supplied to it?

gbryant-arm:
@egrimley-arm The way things are implemented, the VOD readme is parsed by the CI script while still being palatable to someone who doesn't care about the CI and just wants to build and run VOD. Finding the right balance between sharing code in the README and having CI-specific code is hard to find, and I reckon the balance we've reached here is not bad.
Moving CI code to a separate script is possible but would intricate things further.
Out of curiosity, why and how would we run the CI outside of github?

egrimley-arm:
> Out of curiosity, why and how would we run the CI outside of github?

I tend to do that when investigating why CI failed, and I also like to run an abbreviated version of the CI tests, rather than follow entirely separate instructions, to check whether a change I've made breaks anything.

.github/workflows/main.yml
```diff
@@ -65,6 +69,143 @@ jobs:
...
+          workspaces/linux-host/target/debug/veracruz-server \
+          workspaces/linux-runtime/target/debug/runtime_manager_enclave \
+          artifacts/
+      - name: Upload deployment artifacts
+        id: upload-deployment-artifacts
+        uses: actions/upload-artifact@v3
+        with: 
+          name: linux_deployment_artifacts
+          path: |
+            artifacts/proxy_cleanup.sh
```
egrimley-arm, line 99: Apparently you can use a wildcard instead of repeating the list of 8 files, so you could just write `artifacts/*` here.

.github/workflows/main.yml
```diff
@@ -65,6 +69,143 @@ jobs:
...
+      - name: Check out the VOD repository
+        uses: actions/checkout@v3
+        with:
+          repository: 'veracruz-project/video-object-detection'
+          ref: 'main'
+          submodules: recursive
+          set-safe-directory: true
+      - name: Build
+        run: |
+          # grab every bash code block for this step, remove line continuation,
```
egrimley-arm, line 125: This 9-line pattern is now repeated 6 times in the file. I wouldn't insist on it, but it might we worth moving it into one place, which would presumably have to be a new file, unfortunately, so it's not a big win.

.github/workflows/main.yml
egrimley-arm: I've just tried the `make linux` and `make linux-tests` with and without `PROFILE=release` (on both commands) on my machine:
without: 4:21 + 8:27
with: 7:55 + 2:17
Is that plausible, or was something else interfering, I wonder?
If it is roughly representative, then `PROFILE=release` is worth having even without the `vod-full-deployment`.
Perhaps I'll set it to run a few more times while I'm having lunch ...

egrimley-arm: Top tip: Don't try to use `time` in a `/bin/sh` script!

gbryant-arm: Ooh interesting! Can you reproduce that experiment a few times?
Maybe we could also limit the build to the artifacts we need: `make -C /work/veracruz/workspaces linux` builds the Rust examples and data generators, which we don't need

egrimley-arm: We decided to stick with a debug build for the CI because the GitHub machine doesn't have lots of cores for parallel building, but since I have the data from my machine I might as well post it here. I've ordered the times in each line so the third one is the median of the five.
```
debug build real: 3m24.437s, 3m25.781s, 3m26.476s, 3m28.335s, 3m29.153s
debug build user: 26m0.750s, 26m10.264s, 26m18.469s, 26m3.376s, 26m6.292s
debug build sys: 2m24.606s, 2m24.846s, 2m24.939s, 2m25.214s, 2m27.750s

debug tests real: 6m13.795s, 6m20.578s, 6m22.794s, 6m24.346s, 6m25.020s
debug tests user: 4m21.314s, 4m21.385s, 4m21.557s, 4m22.661s, 4m24.407s
debug tests sys: 0m26.566s, 0m26.674s, 0m26.951s, 0m27.113s, 0m27.270s

release build real: 7m46.685s, 7m46.804s, 7m48.685s, 7m49.407s, 7m49.688s
release build user: 37m19.445s, 37m3.408s, 37m7.533s, 37m8.536s, 37m9.781s
release build sys: 2m1.333s, 2m1.609s, 2m1.728s, 2m2.033s, 2m3.264s

release tests real: 2m12.240s, 2m12.978s, 2m8.444s, 2m8.673s, 2m8.675s
release tests user: 3m36.855s, 3m36.911s, 3m37.511s, 3m38.965s, 3m39.464s
release tests sys: 0m9.687s, 0m9.763s, 0m9.815s, 0m9.870s, 0m9.946s
```

gbryant-arm: Interesting! How come debug tests is slower than release tests?

mathias-arm: Debug builds are faster because they do less optimizations and no LTO. The resulting binaries are also slower, often much slower.

gbryant-arm: @mathias-arm Never mind, I misread @egrimley-arm 's table

ShaleXIONG (approved):

.github/workflows/main.yml
gbryant-arm: Let's refactor this in another PR

PR #600, opened 2023-04-11
Add "std" feature to "serde" dependency of virtio-console-server.

egrimley-arm:
We can see that this is a real requirement by trying to build virtio-console-server on its own. We have survived without it so far because virtio-console-server was only ever built in the same workspace together with runtime-manager, which depends on veracruz-utils, which depends on mbedtls, which has the "default" feature, that includes "std", that specifies "serde/std".

It is better to make the requirement explicit rather than depend on the intricacies of Cargo.


dreemkiller (approved):

egrimley-arm:
(It turns out that instead of adding `, "std"` to `Cargo.toml` I could have removed `#![no_std]` from `src/main.rs`: a different way of making it possible to build `virtio-console-server` on its own.)

Issue #601, opened 2023-04-12
`compute_file_hash()` is misleading


gbryant-arm:
**Describe the bug**
The comments and error messages in `compute_file_hash()` indicate that this function is specific to WASM binaries, which isn't the case. Counter example: It's used to compute the Linux enclave hash.

**Expected behaviour**
The comments and error messages in `compute_file_hash()` should be changed to make it more generic and avoid printing misleading error messages.

PR #602, opened 2023-04-12
Change comments and error messages for compute_file_hash per issue #601

dreemkiller:
This is a fix for #601


gbryant-arm (approved):
Nice

PR #606, opened 2023-04-25
Update spin from 0.9.* to 0.9.8 in workspaces/*/Cargo.lock.

egrimley-arm:
This is an extension of dependabot's PR #599, which updated just 0.9.3 to 0.9.8.

There are still uses of spin 0.5.2, because of rust-mbedtls.

### DIFF
--- a/workspaces/applications/Cargo.lock
+++ b/workspaces/applications/Cargo.lock
@@ -891,9 +891,9 @@ dependencies = [
 
 [[package]]
 name = "spin"
-version = "0.9.2"
+version = "0.9.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "511254be0c5bcf062b019a6c89c01a664aa359ded62f78aa72c6fc137c0590e5"
+checksum = "6980e8d7511241f8acf4aebddbb1ff938df5eebe98691418c4468d0b72a96a67"
 dependencies = [
  "lock_api",
 ]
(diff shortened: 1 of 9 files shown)


mathias-arm (approved):
The `mbedtls` depends on `spin = ^0.4.0` according to crates.io. It's strange that we end up with `0.5.2` and `0.9.x`.

egrimley-arm:
> The `mbedtls` depends on `spin = ^0.4.0` according to crates.io. It's strange that we end up with `0.5.2` and `0.9.x`.

This explains `0.5.2`: https://github.com/veracruz-project/rust-mbedtls/commit/c51cd3a99fb933a5094a1966727ade162b2cf154

mathias-arm:
> > The `mbedtls` depends on `spin = ^0.4.0` according to crates.io. It's strange that we end up with `0.5.2` and `0.9.x`.
> 
> This explains `0.5.2`: [veracruz-project/rust-mbedtls@c51cd3a](https://github.com/veracruz-project/rust-mbedtls/commit/c51cd3a99fb933a5094a1966727ade162b2cf154)

Should we bump it to 0.9.8?

egrimley-arm:
> Should we bump it to 0.9.8?

I'll try it to see if it works, anyway ...

egrimley-arm:
It seems that the highest released version of `lazy_static` uses version `0.5.0` of `spin`, and several of our crates use `lazy_static`, directly or indirectly.

PR #607, opened 2023-05-05
Implement WASM bypass

gbryant-arm:
On some systems where WASM is not supported, it might be beneficial to be able to invoke native binaries straight away without needing a call from a WASM program.
This PR implements just that. We add a new type of native modules, provisioned, which is provisioned like a WASM program, but is executed just like a dynamic native module, i.e. in a sandbox environment. Programs provisioned without a `*.wasm` extension are interpreted as provisioned native modules.
Note the program must have read access to itself so that it can be copied to the kernel's filesystem upon invocation.
Also note there is no support for providing an execution configuration to a provisioned native module. It's not hard to implement but would bloat policies a bit.

~~NB: This PR sits on top of https://github.com/veracruz-project/veracruz/pull/587 which will have to be merged beforehand~~

### DIFF
--- a/execution-engine/src/fs.rs
+++ b/execution-engine/src/fs.rs
@@ -462,7 +462,7 @@ impl InodeTable {
                 let path = path.unwrap();
                 let service = Arc::new(Mutex::new(Box::new(native_module.clone())));
                 let new_inode = self.new_inode()?;
-                let path = strip_root_slash(path);
+                let path = strip_root_slash_path(path);
                 // Call the existing function to create general files.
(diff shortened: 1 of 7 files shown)


egrimley-arm (approved):
This looks all right to me.
There's no test, but perhaps there's a good reason for that: how would we test native code?
I didn't know about "raw identifiers"!

gbryant-arm:
>  There's no test, but perhaps there's a good reason for that: how would we test native code?

Interesting question actually! VOD is part of the CI but VOD in its native form is only run as a standalone binary outside Veracruz. We'd benefit from testing the native module infra (native module manager, sandboxer, policies)

PR #609, opened 2023-05-09
Add documentation for native modules

gbryant-arm:
Document https://github.com/veracruz-project/veracruz/pull/587 and https://github.com/veracruz-project/veracruz/pull/607

### DIFF
--- a/COMPONENTS.markdown
+++ b/COMPONENTS.markdown
@@ -26,6 +26,9 @@ runtime which drives the other components.
 - platform_services: provides an abstraction layer over important services that
 each isolate implementation provides.  At the moment, this consists of a single
 service: random number generation.
+- Native Module Manager: manages the secure execution of native modules, i.e.
+non-WASM programs that can be invoked by a WASM program via the VFS, or by a
+participant directly. See `Native modules` below for more details.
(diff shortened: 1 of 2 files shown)


COMPONENTS.markdown
```diff
@@ -57,6 +60,40 @@ service exposes to client code.
...
+## Native modules
+In addition to WebAssembly, Veracruz can execute native code, which has the
+benefit of increasing performance (native [VOD](https://github.com/veracruz-project/video-object-detection)
+executes ~35% faster than WebAssembly) and simplifying the build of
+complex software packages, at the cost of reducing portability.
+Native code execution is achieved by the Native Module Manager described above.
+Here is a summary of the several ways to execute native code in Veracruz:
+
+- Static native module:
+  + Part of the runtime
```
ShaleXIONG, line 72: The sub-bullets here are difficult to read. It will be better to write a paragraph for each bullet.

gbryant-arm: Replaced the bullet points with a table

dreemkiller (approved):

COMPONENTS.markdown
gbryant-arm: I replaced the bullet points with a table

PR #613, opened 2023-05-11
Veracruz-Linux: Resolve symlinks to source locations in debug info at build time

gbryant-arm:
rustc doesn't seem to resolve symlinks to source locations in debug info by default, and since we're using symlinks for our workspaces (https://github.com/veracruz-project/veracruz/tree/main/workspaces/linux-host/crates and https://github.com/veracruz-project/veracruz/tree/main/workspaces/linux-runtime/crates), this causes debugging complications down the line (cf. https://github.com/vadimcn/codelldb/discussions/931).
Several workarounds exist at the debugger level but I propose to resolve source locations by remapping them at build time, as it seems to be the lowest hanging fruit and generates very little friction overall.
This PR does that, resulting in improved debugging in VSCode using the CodeLLDB, one of the few if not only solution to interactively debug Rust code in VSCode

### DIFF
--- a/workspaces/linux-host/Makefile
+++ b/workspaces/linux-host/Makefile
@@ -35,10 +35,13 @@ TEST_PARAMETERS = VERACRUZ_POLICY_DIR=$(OUT_DIR) \
 
 CLIPPY_OPTIONS = --no-deps -A clippy::type_complexity -A clippy::module_inception -D warnings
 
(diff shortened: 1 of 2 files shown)


egrimley-arm (approved):
This looks all right, but I'm curious to know whether the file paths in the DWARF info are absolute or relative and whether this PR changes that.

PR #614, opened 2023-05-23
Update build system to reduce need for Makefile

mathias-arm:
Update workspaces so `cargo build` works without needing features or environment variables. This allows out-of-the-box IDE support (like Visual Studio Code with `rust-analyzer`).


gbryant-arm (changes requested):
Nice one!
Looks good at first glance, just trying to understand how this works.
Mutexing this PR till I can build Nitro and successfully open the repo in VSCode

workspaces/applications/Makefile
```diff
@@ -15,8 +15,7 @@ WORKSPACE_DIR = $(abspath ..)
 include $(WORKSPACE_DIR)/common.mk
 
 all:
-	rustup target add wasm32-wasi
-	cargo build --target wasm32-wasi $(PROFILE_FLAG) $(V_FLAG)
+	cargo build $(PROFILE_FLAG) $(V_FLAG)
```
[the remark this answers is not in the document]
mathias-arm: Good idea

workspaces/icecap-host/src/icecap-host/src/lib.rs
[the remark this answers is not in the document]
mathias-arm: It's an empty package that is part of the workspace. Its only function is to toggle some features for the crates where those features were typically enabled on the cargo command-line.

workspaces/nitro-host/Makefile
```diff
@@ -68,22 +61,16 @@ doc:
...
 	$(CC) cargo clippy $(PROFILE_FLAG) $(V_FLAG) \
 		-p veracruz-client \
 		-p veracruz-server -p io-utils -p transport-protocol \
-		-p psa-attestation -p veracruz-utils  \
-		--features veracruz-client/nitro \
-		--features veracruz-server/nitro \
-		--features io-utils/nitro \
-		--features psa-attestation/nitro \
-		--features veracruz-utils/nitro \
+		-p psa-attestation -p veracruz-utils \
```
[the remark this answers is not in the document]
mathias-arm: I leave the `-p` parameters for the `cargo clippy` commands.

gbryant-arm:
> It only works if you open one workspace at a time.

You mean `workspaces/linux*` or `workspaces/nitro*` for instance?

mathias-arm:
> You mean `workspaces/linux*` or `workspaces/nitro*` for instance?

Yes, you open Visual Studio Code in one of the workspaces (`applications`, `linux-host`, etc.) so that it has the same "view" as the `cargo build` command, which means `rust-analyzer` gets the right context
- The right version of Rust is used thanks to `rust-toolchain.toml` 
- The right target is used thanks to `.cargo/config.toml`
- The right top-level project is used: having the workspace `Cargo.toml` at the top-level for VSCode means that it gets picked up and, with my tweak, the right features get enabled. If I open VSCode at the Veracruz top-level directory, it keeps starting rust-analyzer on each crate independently with default features (many crates actually don't compile if you don't enable non-default features).

workspaces/linux-host/src/linux-host/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+authors = ["The Veracruz Development Team"]
+description = "A dummy package"
+edition = "2018"
+name = "linux-host"
+version = "0.3.0"
+
+[dependencies]
+tests = { path = "../../crates/tests", features = [ "linux" ] }
```
egrimley-arm, line 9: This dependency on `tests` gives a warning, I think, though removing it leads to an error. Not sure what's best!

```
warning: linux-host v0.3.0 (/work/veracruz/workspaces/linux-host/src/linux-host) ignoring invalid dependency `tests` which is missing a lib target
```

mathias-arm: I forgot to commit a file (empty `tests/src/lib.rs`). That should be fixed now.

icecap/README.md
```diff
@@ -104,7 +104,7 @@ with some indications of how they depend on each other.
 
 ## How Veracruz IceCap binaries start
 
-There are two Veracruz IceCap binaries: `runtime_manager_enclave.elf`
+There are two Veracruz IceCap binaries: `runtimeimanager-enclave.elf`
```
egrimley-arm, line 107: `runtimeimanager-enclave` is presumably a typo.

tests/src/lib.rs
```diff
@@ -0,0 +1 @@
+// Intentionally empty
```
egrimley-arm, line 1: Perhaps you want to add a comment like that in `workspaces/icecap-host/src/icecap-host/src/lib.rs`, too.

(I'm commenting here because I can't see how to add a comment on an empty file!)

workspaces/icecap-runtime/Makefile
```diff
@@ -422,3 +422,4 @@ build: elfloader
 
 .PHONY: clean
 	rm -rf css-icecap.bin $(build_dir)
```
egrimley-arm, line 424: I changed this on `main` recently. You'll probably have to rebase.

tests/src/lib.rs
egrimley-arm: Also `workspaces/linux-host/src/linux-host/src/lib.rs`.

egrimley-arm (commented):
This looks good to me, but Guilhem wanted to run some test with debug, I think.

gbryant-arm:
> Yes, you open Visual Studio Code in one of the workspaces (applications, linux-host, etc.) so that it has the same "view" as the cargo build command, which means rust-analyzer gets the right context

rust-analyzer works on my end, neat.

Debugging doesn't work anymore though: https://github.com/veracruz-project/veracruz/pull/613 was resolving the symlinks to the codebase, and if we revert it, only symbols under a specific workspace (e.g. `linux-host`) will get resolved, which is fine if you debug a single binary but becomes problematic if you debug several binaries belonging to different workspaces, say veracruz-server and runtime manager.
I'm not sure how to solve this nicely. We could remap the same binaries on a workspace basis, but that might prove impractical. Other than that there are some workarounds at the debugger level but aren't exactly satisfactory.

On another note, should we document how to get Veracruz up and running with VS Code? I can help with the debugging part

egrimley-arm:
> Debugging doesn't work anymore though

Do we merge anyway, though? Or should this be fixed before merging?

workspaces/icecap-runtime/Makefile
mathias-arm: Fixed in rebase

mathias-arm:
> Debugging doesn't work anymore though: #613 was resolving the symlinks to the codebase, and if we revert it, only symbols under a specific workspace (e.g. `linux-host`) will get resolved, which is fine if you debug a single binary but becomes problematic if you debug several binaries belonging to different workspaces, say veracruz-server and runtime manager. I'm not sure how to solve this nicely. We could remap the same binaries on a workspace basis, but that might prove impractical. Other than that there are some workarounds at the debugger level but aren't exactly satisfactory.

I don't think I reverted #613. I think the target environment is to have a Visual Studio Code window attached on the development docker container for each workspace you need to debug on.

> On another note, should we document how to get Veracruz up and running with VS Code? I can help with the debugging part

Something similar in spirit to `CLI_QUICKSTART.markdown` but started from Visual Studio Code?

gbryant-arm:
> I don't think I reverted https://github.com/veracruz-project/veracruz/pull/613

I meant we may have to revert it if we're going for this solution.

> I think the target environment is to have a Visual Studio Code window attached on the development docker container for each workspace you need to debug on.

Okay, I guess we could first debug say veracruz-server, and if no bug is found we debug runtime-manager in another workspace.
That's a bit tedious but could work. Otherwise we can fall back to some workarounds described in https://github.com/vadimcn/codelldb/discussions/931

> Something similar in spirit to CLI_QUICKSTART.markdown but started from Visual Studio Code?

Sounds good :+1:

gbryant-arm:
Hmm actually it might be tricky to debug veracruz-server and runtime-manager, because veracruz-server spawns runtime-manager, which probably means they have to run in the same debugger instance, hence in the same workspace

mathias-arm:
> Hmm actually it might be tricky to debug veracruz-server and runtime-manager, because veracruz-server spawns runtime-manager, which probably means they have to run in the same debugger instance, hence in the same workspace

You can attach the debugger to a running runtime.

dreemkiller (approved):

mathias-arm:
I think part of the debugging issue is that the runtime is compiled into the server (https://github.com/veracruz-project/veracruz/blob/main/veracruz-server/src/platforms/linux.rs#L44) and saved in a temporary directory. If it was executed from its `target` path I think there would be less issues.

gbryant-arm (approved):
LGTM.
Regarding debugging we'll figure something out later. Workarounds exist

Issue #541, opened 2022-10-12
EIF image rebuilt every other time


gbryant-arm:
**Describe the bug**
The EIF image is rebuilt every other time

**To Reproduce**
* Run `PROFILE=dev make nitro` several times and witness the weirdness

**Expected behaviour**
The EIF image should be rebuilt whenever `runtime_manager_enclave` changes

egrimley-arm:
This was fixed by #617.

PR #615, opened 2023-05-24
Preserve file modification date when stripping binaries

gbryant-arm:
Fix for https://github.com/veracruz-project/veracruz/issues/541

### DIFF
--- a/workspaces/nitro-host/Makefile
+++ b/workspaces/nitro-host/Makefile
@@ -41,7 +41,7 @@ build:
 		--features veracruz-client/nitro \
 		--features veracruz-server/nitro \
 		--features cli
-	strip target/$(PROFILE_PATH)/veracruz-client target/$(PROFILE_PATH)/veracruz-server
+	strip -p target/$(PROFILE_PATH)/veracruz-client target/$(PROFILE_PATH)/veracruz-server
 
 .PHONY: $(MEASUREMENT_FILE)
 $(MEASUREMENT_FILE):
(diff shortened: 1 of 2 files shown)


egrimley-arm:
The change looks safe enough but ... why should the EIF be rebuilt every time (is it because of a hardwired certificate expiry date?), why was it being rebuilt every other time, why does this change fix the problem, and is it safe to modify the output of `cargo build` in place rather than create the stripped file elsewhere, which is what happens in `workspaces/icecap-runtime/Makefile` using a `-o` argument with `strip`?

gbryant-arm:
> why should the EIF be rebuilt every time (is it because of a hardwired certificate expiry date?)

I don't know how often it should be built. At least as often as the RM binary is built.
Before this PR, the EIF would be built everytime the RM binary is built, but also every other time the RM binary was not built IIUC, which is confusing and wastes CPU time.

> why was it being rebuilt every other time

It was built every other time because `strip` was called regardless of whether `cargo build` was successful in building RM (there doesn't seem to be any way to know whether `cargo build` actually rebuilt the binary, besides grepping the output or looking at the binary), which changed the modification date of the RM binary, which led `make` to believe on the next run that the RM binary had been updated.

> why does this change fix the problem

Now `strip` preserves modification time, which is one solution among others

> and is it safe to modify the output of cargo build in place rather than create the stripped file elsewhere, which is what happens in workspaces/icecap-runtime/Makefile using a -o argument with strip?

In the past we suffered from stale binaries originating from superfluous copies (cf. https://github.com/veracruz-project/veracruz/pull/339). It's a real pain to debug so I'd argue against that.
How about calling `strip` only if the binary was updated?

egrimley-arm:
The alternating behaviour can be demonstrated with this simplified Makefile:
```
eif: binary
	cp binary eif

binary: binary-phony

binary-phony:
	touch binary
```

egrimley-arm:
I've made #617 which fixes a related bug, I think, though I haven't really tested it: it's time to go home now!

gbryant-arm:
Already fixed in https://github.com/veracruz-project/veracruz/pull/617

PR #617, opened 2023-05-25
workspaces/nitro-*/Makefile: Use strip -p and recursive make.

egrimley-arm:
The changes in nitro-host/Makefile are unimportant.

In nitro-runtime/Makefile, when building "nitro", we first invoke "cargo build" and "strip -p" and then use a recursive make invocation to see whether we need to run "nitro-cli". Note the use of "&:", which might not matter here, but apparently one should use it for multiple targets built by a single recipe.

The bug we are fixing here is that without the recursive make invocation a change to the runtime manager's source does not get propagated through to the EIF file by a single invocation of "make". This is likely to be a general problem whenever something depends on the output of "cargo build".


gbryant-arm:
I'm confused about the bug and can't reproduce it right now.
Are you saying that `make` must be run several times to propagate changes to the RM through to the EIF?

egrimley-arm:
> I'm confused about the bug and can't reproduce it right now. Are you saying that `make` must be run several times to propagate changes to the RM through to the EIF?

Yes. Sometimes you have to run `make` twice.

egrimley-arm:
To demonstrate the problem:

* `cd veracruz/workspaces`
* `make nitro`
* `touch ../runtime-manager/src/main.rs`
* `make nitro`
* output should include `Enclave Image successfully created` near the end
* `touch ../runtime-manager/src/main.rs`
* `make nitro`
* output should include `Enclave Image successfully created` near the end

If `Enclave Image successfully created` is missing in either case, it's not working.

egrimley-arm:
I used `touch ../runtime-manager/src/main.rs` for simplicity. You may also replace that command with something that makes a significant change to the source, like `perl -i ...` something or other or change one of the source files with an editor.

egrimley-arm:
It doesn't matter whether updating the timestamp on a source file causes a rebuild, but a functional change must result in a rebuild. I think `cargo` doesn't distinguish between the two cases, so we can use `touch` to simulate a significant change, but it would be safer to test by making a real significant change somewhere.

egrimley-arm:
`perl -i -pe 's/(nitro_main) /$1 x /;' ../runtime-manager/src/main.rs` instead of the `touch` perhaps?

gbryant-arm:
Okay, finally managed to reproduce it. Dunno what happened.
And forget what I said about cargo not generating a new binary if the code doesn't functionally change, that's complete BS :no_good: 
LGTM!
Just one question: why did you remove `test-collateral` from the PHONY target?

egrimley-arm:
> LGTM! Just one question: why did you remove `test-collateral` from the PHONY target?

It's in `shared.mk` anyway.

gbryant-arm (approved):

PR #618, opened 2023-05-31
Avoid unnecessary rebuilds when building for Linux

egrimley-arm:
To demonstrate the problem that this PR is intended to fix:

* `cd veracruz/workspaces`
* `make clean`
* `make linux`
* `make linux`

There will be several invocations of `cargo build`, but if any significant work happens after the second `make linux`, that's a bug.

---

* `transport-protocol/build.rs`: Put generated file in tmp dir and rename.

This is to avoid unnecesssary rebuilds when several other crates have a path dependency on transport-protocol. Without this change, one of the crates getting rebuilt causes the timestamp on the generated file (`transport_protocol.rs`) to get updated, which makes the other crates get rebuilt by `cargo build` next time.

A neater solution might be to put the generated file under `OUT_DIR`, but currently that seems to be difficult.
(See https://github.com/stepancheg/rust-protobuf/issues/673.)

* `examples/data-generators/`: Fix dependencies on directories.

Also various minor simplifications and tidying.

A dependency on a directory causes unnecessary rebuilds as make looks at the modification time of the directory, which gets updated as files are created in the directory, so the directory ends up being newer than some of the target files, forcing a rebuild. An order-only prerequisite is used to avoid this problem.

List of changes:

- Use an "order-only prerequisite" to create `$(TARGET_DIR)`.
- Remove all uses of `.SECONDEXPANSION` and `notdir`.
- Remove all "static pattern rules".
- Use "grouped targets" where appropriate.
- Update sources for external data.
- Add `moving-average-*/apple_prices.csv`.
- Remove trailing `/` from value of `TARGET_DIR`.
- Use `quick-clean: clean`.
- Remove trailing spaces.


egrimley-arm:
I was only looking for problems in the Linux build but I expect these changes also affect builds for other targets.

The percentage change in the build time from a fresh checkout is probably rather small, but rebuilds after changing certain components might be changed more significantly.

mathias-arm:
Why not use the `|` (https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html):
```
$(TARGET_DIR_DATS): %.dat : $$(notdir $$@) | $(TARGET_DIR)
```

egrimley-arm:
Unless you've got several recipes that need to create the same directory, doesn't that just make the `Makefile` more complicated?

egrimley-arm:
There are now quite a lot of new changes in `examples/data-generators/*/Makefile`.

mathias-arm (approved):

PR #619, opened 2023-06-02
Created separate crates for each of the platform's runtime managers

### LINKED ISSUE #81: Isolate platform specific changes to their own crates

dreemkiller:
Split runtime-manager into platform-specific top level crates (linux-runtime-manager, nitro-runtime-manager, icecap-runtime-manager) so new platforms are easier to add and to make code maintenance easier (using features for this in one crate was a mistake). This partially resolves #81


icecap-runtime-manager/src/main.rs
```diff
@@ -0,0 +1,201 @@
...
+            event: event,
+            virtio_console_server_ring_buffer_badge: virtio_console_server_ring_buffer_badge,
+            active: true,
+        }
+    }
+
+    fn run(&mut self) -> Fallible<()> {
+        let icecap_runtime = icecap_runtime::IcecapRuntime{};
+
+        println!("icecap_runtime_manager::linux_main accept run looping");
```
mathias-arm, line 94: I think this should use `debug_println!` and I think that `linux` bit is inaccurate.

mathias-arm (changes requested):
This PR adds a lot of `println!` calls that should be converted to `log::info` (or `log::debug`, etc.) calls.

mathias-arm (dismissed):

PR #620, opened 2023-06-05
Improve IceCap Makefile, particularly PHONY targets.

egrimley-arm:
* `workspaces/icecap-runtime/Makefile`: Remove most of the PHONY targets.

If a target with a recipe depends on a phony target, then the recipe gets run again and again and again, so don't do that.

Instead we create empty files in `t/` that record when a target was last built.

This prevents some things from getting unnecessarily built twice during a normal build after a fresh checkout.

Also add the `clean` target.

### DIFF
--- a/workspaces/icecap-runtime/Makefile
+++ b/workspaces/icecap-runtime/Makefile
@@ -120,40 +120,27 @@ $(now):
 # seL4
 ###########################################################################
 
-.PHONY: sel4-configure
-sel4-configure: $(sel4_build_dir) $(sel4_dts_path)
-		cmake -G Ninja \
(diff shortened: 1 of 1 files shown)


mathias-arm:
One of the reason things get re-built is that for the different parts of the IceCap build it's difficult to have `make` determine when they need to be re-compiled, and they also are bad at being idem-potent. So we have to make a choice between re-building everything all the time, or risk that some seL4 part that rarely changes will not be rebuilt when it changes unless you run `make clean`. I am in favor of your proposed change, but I think we should have more input.

mathias-arm (approved):

PR #621, opened 2023-06-07
Switch back to building parse-capDL with stack

mathias-arm:
The dependencies for stack2cabal are not fully pinned, so it seems a dependency update broke the build.


egrimley-arm (approved):
I've not tested this, but it looks good to me.

PR #626, opened 2023-06-20
Slightly improve the Makefiles, especially icecap-host/Makefile.

egrimley-arm:


workspaces/linux-host/Makefile
```diff
@@ -69,6 +69,7 @@ test-client: test-dependencies
 veracruz-test: test-dependencies build-tests
 	RUST_LOG="integration_test=info,error" $(CARGO_TEST) --test integration_test -- --test-threads=1 --nocapture
 
+.PHONY: $(RUNTIME_ENCLAVE_BINARY_PATH)
```
mathias-arm, line 72: That is not consistent with the changes in IceCap. We should probably align on having a `.PHONY` target `<something>-runtime` rather that the runtime path as the dependency.

mathias-arm (changes requested):

workspaces/icecap-host/Makefile
```diff
@@ -65,10 +65,12 @@ build-veracruz-server-test: rustup-plat $(VERACRUZ_ICECAP_QEMU_IMAGE)
 		) target/release/veracruz-server-test
 
 .PHONY: $(MEASUREMENT_FILE)
```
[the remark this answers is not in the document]
egrimley-arm: `$(MEASUREMENT_FILE)` is a target in `shared.mk`, but if it's going to be phony then it might as well be a simple name rather than a variable expression. I'll try changing it to `measurement-file`.

workspaces/linux-host/Makefile
egrimley-arm: Yes, I'll try that.

mathias-arm (approved):

mathias-arm (approved):

PR #627, opened 2023-06-20
Updates after runtime split

mathias-arm:
This pull request contain multiple updates:
- Update and align on libc 0.2.144 (including IceCap)
- Update `postcard`
- Align the `Cargo.lock` files in different workspaces to the same version.
- Use aws-nitro-enclaves-nsm-api published crate rather than git tag
- Use `rustls` for `reqwest` (instead of `openssl`) because it works better for cross-compilation.


nitro-runtime-manager/src/nitro_runtime.rs
```diff
@@ -39,39 +37,36 @@ impl PlatformRuntime for NitroRuntime {
...
+        }
+
+        let request = nsm_api::api::Request::Attestation{
+            user_data: Some(serde_bytes::ByteBuf::from(csr_hash)),
+            nonce: Some(serde_bytes::ByteBuf::from(challenge.clone())),
+            public_key: None
         };
-        att_doc.insert(0, 0xd2); // the golang implementation of cose needs this. Still need to investigate why
+        // generate the attestation document
+        let response = nsm_api::driver::nsm_process_request(nsm_fd, request);
```
dreemkiller, line 54: should we instead do, above:
```
use nsm_api::driver::nsm_process_request;
```
so we can have:
```
let response = nsm_process_request(nsm_fd, request);
```
here?

nitro-runtime-manager/src/nitro_runtime.rs
```diff
@@ -39,39 +37,36 @@ impl PlatformRuntime for NitroRuntime {
...
-                nsm_api::api::ErrorCode::Success => (),
-                _ => return Err(anyhow!(RuntimeManagerError::NsmErrorCode(status))),
-            }
-            unsafe {
-                buffer.set_len(buffer_len as usize);
-            }
-            buffer.clone()
+
+        let csr_hash = sha256(&csr);
+        let nsm_fd = nsm_api::driver::nsm_init();
```
dreemkiller, line 42: should we instead do, above:
```
use nsm_api::driver::nsm_init;
```
so we can have:
```
let nsm_fd = nsm_init();
```
here?

nitro-runtime-manager/src/nitro_runtime.rs
```diff
@@ -39,39 +37,36 @@ impl PlatformRuntime for NitroRuntime {
...
+        if nsm_fd < 0 {
+            let e = nsm_api::api::ErrorCode::InternalError;
+            return Err(anyhow!(RuntimeManagerError::NsmErrorCode(e)));
+        }
+
+        let request = nsm_api::api::Request::Attestation{
```
dreemkiller, line 48: I'd like to do the same here, but `Attestation` is probably a bit too generic for that

mathias-arm: I think I will import `nsm_api::api::{Request, Response}`, this way it will be `Request::Attestation` and it will not clash with `Response::Attestation`

workspaces/nitro-host/Makefile
```diff
@@ -10,10 +10,12 @@
 # licensing and copyright information.
 
 .PHONY: all build clean clean-cargo-lock clippy default doc fmt \
-	test-client test-dependencies test-server veracruz-test
+	test-client test-dependencies test-server veracruz-test build-tests
```
egrimley-arm, line 13: Nit: "tests" is missing here?

mathias-arm: I will fix that in another PR, there are some other minor changes I have in mind.

PR #628, opened 2023-06-21
Split veracruz-server into platform-specific crates

### LINKED ISSUE #81: Isolate platform specific changes to their own crates

dreemkiller:
Create a `linux-veracruz-server` crate, an `icecap-veracruz-server` crate, and a `nitro-veracruz-server` crate.
Modify the `veracruz-server` crate to contain the platform-independent code, as well as provide a trait that the platform-specific crates can implement.

This will completely resolve #81


gbryant-arm (approved):
LGTM

PR #632, opened 2023-06-22
Tidy up all the Rust files with rustfmt --edition=2018.

egrimley-arm:
We should probably do this at some point, but it doesn't have to be now.

I was surprised at how many files were changed (28) despite our `githooks/pre-commit`.


mathias-arm (approved):
Looks good. I rebased it and re-ran `fmt` just in case.

mathias-arm (approved):

PR #635, opened 2023-06-27
Test that native module encryption and decryption works. Fix #502.

### LINKED ISSUE #502: Test that native module encryption and decryption works
Currently there does not seem to be anything to test that the encryption and decryption implemented in `native_modules/aes.rs` and `native_modules/aead.rs` works, for example that encryption followed by decryption gives the original data.

(I added something like `output[1] = 3;` and nothing broke. Was I just lucky?)

egrimley-arm:


gbryant-arm:
Come to think of it, it doesn't make any difference to remove the pipeline, since you explicitly request the program execution in the tests. But I've just noticed pipelines are not supported in the CLI client

PR #645, opened 2023-07-25
Prepare 23.07 docker images

mathias-arm:


.gitignore
```diff
@@ -45,3 +45,22 @@ sdk/target
 */.idea/
 */*/.idea/
 build/**
+
+execution-engine/Cargo.lock
```
egrimley-arm, line 49: Any reason not to use `*/Cargo.lock` here?

docker/Makefile
```diff
@@ -20,9 +20,16 @@ DOCKER_GROUP_ID=$(shell stat --format "%g" $(shell readlink -f /var/run/docker.s
 NE_GID := $(shell getent group ne | awk -F: '{printf $$3}')
 endif
 
+ifeq ($(ARCH),x86_64)
+DOCKER_ARCH = amd64/
```
egrimley-arm, line 24: The trailing `/` is a bit unexpected. Is it necessary?

docker/Makefile
```diff
@@ -50,6 +57,10 @@ base: base/Dockerfile
 all-base: base linux-base nitro-base icecap-base
 	echo 'All base docker images re-built from scratch'
 
+# Make sure the cargo registry directory exists to avoid permission issues
+$(HOME)/.cargo/registry:
+	mkdir -p $(HOME)/.cargo/registry
```
egrimley-arm, line 62: You can use `$@` instead of repeating the target.

transport-protocol/.gitignore
```diff
@@ -1 +1,2 @@
+src/transport_protocol.rs
```
egrimley-arm, line 1: I think this file is no longer created so it shouldn't be necessary to ignore it.

.gitignore
mathias-arm: I wanted to be sure that I would not ignore more files than I needed to but your suggestion is more future-proof.

docker/Makefile
mathias-arm: Yes. The Dockerfile does not support the case where the `/` is not in the variable.

egrimley-arm (approved):

PR #647, opened 2023-07-27
Update CI to use v23.07 Docker image

mathias-arm:


.github/workflows/main.yml
```diff
@@ -27,15 +27,15 @@ jobs:
         id: cosign-verify
         run: |
             COSIGN_EXPERIMENTAL=true cosign verify \
-              ghcr.io/veracruz-project/veracruz/ci@sha256:f9c1d079ce0fc404ec8a9c5f6c70a52fc5e52dc4866847d1b66260006207f67b
```
egrimley-arm, line 30: Why did this one used to be different from the others?

egrimley-arm: Obviously it would be nice not to repeat that hash in six places. I have no idea whether something like this would work: #648

egrimley-arm: It looks like it doesn't work, at least not like that.

PR #648, opened 2023-07-27
WIP: Can we use a variable here?

egrimley-arm:
See #647


mathias-arm:
I think I tried the same thing (or the variant where the whole image name was a variable) and it did not work.

egrimley-arm:
I give up for now!

PR #653, opened 2023-08-15
Revert "Fix docker images name"

egrimley-arm:
This reverts commit f5af8054ddc8e72b1758d2f47f900cac69cf4711.

The docker image names should perhaps be changed, but that commit changes only some of them and breaks things: I tested `make localci-build` after doing `docker rmi ...` to get rid of the existing images that might otherwise get used.


egrimley-arm:
I'll make a similar PR for `main` if this is approved.

mathias-arm (approved):

PR #654, opened 2023-08-15
Revert "Fix docker images name"

egrimley-arm:
This reverts commit f5af8054ddc8e72b1758d2f47f900cac69cf4711.

This is a PR on `main` corresponding to #653 on `veracruz-v2`.

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -105,7 +105,7 @@ localci-exec:
 	docker exec -i -t $(CONTAINER)-localci-$(USER)-$(VERSION) /bin/bash || true
 
 localci-base: ci/Dockerfile.local ci-base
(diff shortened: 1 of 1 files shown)


mathias-arm:
Simply reverting does not address the issues from #650. I think `docker/README.md` also needs to be updated, because some issues stem from the difference between `make linux-base` and `make linux-build` (replace `linux` with `nitro`, `ci` or `localci`).

egrimley-arm:
There seem to be a lot of things wrong in `docker/README.md` (I don't think "squash" is an experimental feature any more, there is no submodule, ...), but I can't see any that relate to this PR in particular. I'll probably make a separate PR tomorrow, or two PRs as those things should probably be fixed on `veracruz-v2`.

mathias-arm:
Closing in favor of #656 which contains this change and I think is the best solution.

PR #656, opened 2023-08-16
Update docker/README.md

egrimley-arm:
This is a version for `main` of #655 for `veracruz-v2`.

Implemented on top of #654, which is expected to be merged first, so "Files changed" shows more than one PR. It probably makes more sense to look at the top commit.

PR #657, opened 2023-08-17
Add $USER to all Docker image names.

egrimley-arm:
This is another attempt to do what #650 was probably trying to do.

Implemented on top of #656, which is expected to be merged first, so "Files changed" shows more than one PR. It probably makes more sense to look at the top commit.


mathias-arm:
I don't think we need the intermediate images to be renamed only the "leaf" one. I am worried also about the consequences on the `ci` image build.

egrimley-arm:
> I don't think we need the intermediate images to be renamed only the "leaf" one.

That might be what we already have (after #654 is applied).

> I am worried also about the consequences on the `ci` image build.

I'm worried about that, too, and generally how there a lot of things to test here.

third-party/rust-mbedtls
mathias-arm, line 1: [9232612](https://github.com/veracruz-project/veracruz/pull/657/commits/92326124a717e700a3f1539b89c0aa927adaf813) includes this change which is surprising given the commit name. If I understand it changes from https://github.com/veracruz-project/rust-mbedtls/commit/7deda7121d54e97693c1e800bb717add116b80ef to https://github.com/veracruz-project/rust-mbedtls/commit/843291002b226a38821acda887bc05548dec2f2f which is to me a regression.

mathias-arm (changes requested):

third-party/rust-mbedtls
egrimley-arm: Oh, well spotted. It wouldn't be the first time that I've accidentally included a change to a submodule by changing to another branch, not doing `git submodule update`, and then using `git commit -a`.

mathias-arm:
Closing in favor of #656 which I think is the best solution.

PR #666, opened 2023-10-10
Grant networking capabilities to sandboxee

gbryant-arm:
This is useful for the DPU work.
There are security implications but the sandbox is allowing every syscall anyway. The real solution is to specify sandbox2 policies as an extension of the global policies.

Edit: Don't trust the PR ID, this PR is harmless :innocent:


dreemkiller (approved):

gbryant-arm:
Just signing off the commit

